PR #78, opened 2023-03-12
feat: add semantic-release plugin

RebeccaStevens:
This PR is based of #82 and thus relies on that PR's changes.

https://github.com/semantic-release/semantic-release

### DIFF
--- a/src/plugins/semantic-release/index.ts
+++ b/src/plugins/semantic-release/index.ts
@@ -4,7 +4,7 @@ import { hasDependency } from '../../util/plugin.js';
 import type { PluginConfig } from './types.js';
 import type { IsPluginEnabledCallback, GenericPluginCallback } from '../../types/plugins.js';
 
-// link to docs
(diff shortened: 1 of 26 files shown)


src/plugins/semantic-release/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  'package.json',
+];
+
+export const ENTRY_FILE_PATTERNS = [];
+
+export const PRODUCTION_ENTRY_FILE_PATTERNS = [];
+
+export const PROJECT_FILE_PATTERNS = [];
+
+async function load(filePath: string) {
```
RebeccaStevens, line 31: Other plugin might also need this load behavior (where the file can be either json or yaml when there is no extension) so this might want to be refactor out. Maybe `_load` itself should just handle this.

webpro (approved):
Nice PR, please see the comments in the other plugin you did :)

webpro:
Thanks, @RebeccaStevens! There was a conflict due to all these link ref updates. But this has landed in [v1.16.0]( https://github.com/webpro/knip/releases/tag/1.16.0).

PR #79, opened 2023-03-12
fix: improve support of commitlint configs

RebeccaStevens:


webpro:
I'm happy to merge this, but how did you find out all of this is supported by commitlint? I can't seem to find it in their docs, do you have a link maybe?

RebeccaStevens:
https://github.com/conventional-changelog/commitlint#config

webpro:
> https://github.com/conventional-changelog/commitlint#config

Sorry I missed the comment/link that was already right there 🤦

webpro:
Thanks @RebeccaStevens!

PR #80, opened 2023-03-12
feat: add new plugin cspell

RebeccaStevens:
https://www.npmjs.com/package/cspell

### DIFF
--- a/src/plugins/cspell/index.ts
+++ b/src/plugins/cspell/index.ts
@@ -1,4 +1,5 @@
 import { _load } from '../../util/loader.js';
+import { getPackageNameFromModuleSpecifier } from '../../util/modules.js';
 import { timerify } from '../../util/performance.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { PluginConfig } from './types.js';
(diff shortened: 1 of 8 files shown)


src/plugins/cspell/index.ts
```diff
@@ -0,0 +1,40 @@
...
+// link to docs
+
+export const NAME = 'cspell';
+
+/** @public */
+export const ENABLERS = ['cspell'];
+
+export const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(dependencies, ENABLERS);
+
+// https://github.com/streetsidesoftware/cspell/tree/main/packages/cspell#customization
```
webpro, line 16: Sorry for the nittiest of nitpicks to pick, but all plugins have a link to the docs up there at "link to docs"

src/plugins/cspell/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  'cspell.{json,yaml,yml}',
+  '.c{s,S}pell.json',
+  'cSpell.json',
+];
+
+export const ENTRY_FILE_PATTERNS = [];
+
+export const PRODUCTION_ENTRY_FILE_PATTERNS = [];
+
+export const PROJECT_FILE_PATTERNS = [];
```
webpro, line 28: Arrays that stay empty can be removed

src/plugins/cspell/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const findPluginDependencies: GenericPluginCallback = async configFilePath => {
+  const config: PluginConfig = await _load(configFilePath);
+  const imports = config?.import ?? [];
+  return imports.map(importPath => {
+    const firstSlashIndex = importPath.indexOf('/');
+    const endIndex = importPath.startsWith('@') ? importPath.indexOf('/', firstSlashIndex + 1) : firstSlashIndex;
+    return importPath.substring(0, endIndex);
```
webpro, line 36: If you want, you could use the `getPackageNameFromModuleSpecifier` helper for this (or improve that one if you don't like it, but would be good to be consistent).

webpro (approved):
Thanks! Please also run `npm run docs` to auto-update the list in `README.md`.

webpro:
Thanks, this is great

PR #82, opened 2023-03-12
fix: parse rc configs with no extension as either yaml or json

RebeccaStevens:
rc files can often be written in either yaml or json so knip so try loading them as both.

As yaml is a super type of json. Any valid json can be parsed with the yaml parser - so it's more performant to try this one first. jsonc isn't valid yaml though so we still need to try the json parser if the yaml one fails.

### DIFF
--- a/src/util/loader.ts
+++ b/src/util/loader.ts
@@ -4,19 +4,15 @@ import { pathToFileURL } from 'node:url';
 import { load as esmLoad } from '@esbuild-kit/esm-loader';
 import { require } from '../util/require.js';
 import { LoaderError } from './errors.js';
-import { loadJSON, loadYAML } from './fs.js';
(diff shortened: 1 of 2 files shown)


src/util/loader.ts
```diff
@@ -1,23 +1,30 @@
...
 const load = async (filePath: string) => {
   try {
     const ext = path.extname(filePath);
-    if (ext === '.json' || ext === '.jsonc' || /rc$/.test(filePath)) {
+    if (/rc$/.test(filePath)) {
+      try {
+        // Note: `await` is needed for the try-catch to work.
+        return await loadYAML(filePath);
+      } catch {
+        return loadJSON(filePath);
```
RebeccaStevens, line 18: It would probably be better to load the file only once for both the yaml and json parsers rather than loading it twice.

webpro:
This ended up great, thank you so much! :)

PR #84, opened 2023-03-12
fix: check nested eslint config files

RebeccaStevens:
eslint allows for multiple config files to exists within different directories of a project.

### DIFF
--- a/src/plugins/eslint/README.md
+++ b/src/plugins/eslint/README.md
@@ -12,7 +12,12 @@ or `devDependencies`:
 ```json
 {
   "eslint": {
-    "config": [".eslintrc", ".eslintrc.{js,json,cjs}", ".eslintrc.{yml,yaml}", "package.json"],
+    "config": [
+      "**/.eslintrc",
+      "**/.eslintrc.{js,json,cjs}",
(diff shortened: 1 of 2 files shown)


webpro:
I had this issue myself before. I was in a fairly large Nx project (single repo/package but many libs and apps) and had the same issue. The idea in Knip is provide (only) the default file locations for the sake of performance and clarity. Using all possible variations of eslint config files could lead to unexpected results. However, they can be overridden with custom Knip config. When in a monorepo, there is negated glob patterns to rule out unwanted results from other local packages, but I think this still gets very unwieldy. This becomes more clear when looking at debug output in such projects.

When doing this for the eslint plugin it should be done for all plugins, right? I mean, I think ESLint does not actively search the whole project for these config files.

On the other hand, this setting is very convenient. Especially when not in a monorepo it could help. Just not sure about people's expectations: adding this could lead to surprises, current situations leads to surprises as well. Either way, documentation/awareness should be improved.

webpro:
For now, I'm going to play it safe and (only) improve the docs for plugins (https://github.com/webpro/knip/blob/v2/README.md#multi-project-repositories).

PR #85, opened 2023-03-13
feat: add new plugin ava

RebeccaStevens:
https://github.com/avajs/ava


webpro:
Thanks, @RebeccaStevens. Please see `_getReferencesFromScripts`, the `github-actions` and `husky` plugins are example plugins using this helper to extract binaries and entry files from scripts, also those referenced from the CLI arguments I see in the Ava config. See [the tests](https://github.com/webpro/knip/blob/main/test/tests/getReferencesFromScripts.test.ts) for what it supports. Maybe you need to prepend `node ` to each of `nodeArguments`, or feed it something like `node ${nodeArguments.join(' ')}`.

RebeccaStevens:
Change made :)

webpro:
Wonderful, thank you so much **again**

PR #86, opened 2023-03-13
chore: add launch.json config to allow easily testing a single test file in vscode

RebeccaStevens:


webpro:
Great idea! I have a few myself, but never took the time to actually add them to the repo. Will see if I can add useful ones as well. Thanks!

PR #88, opened 2023-03-15
feat: add support for package self-referencing

### LINKED ISSUE #87: Add support for package self referencing
[package self-referencing](https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name) should be treated the same as local file imports.
re: https://github.com/webpro/knip/issues/77#issuecomment-1465139111

knip `2.0.0-alpha.8` crashes when it encounters a self-referencing import that needs to be looked up in the "exports" field of the "package.json" file.

```sh
> knip --exclude exports

Analyzing workspace (.)...
node:internal/modules/cjs/loader:1075
  const err = new Error(message);
              ^

Error: Cannot find module '/home/rebec/dev/eslint-config-rebeccastevens/script'
Require stack:
- /home/rebec/dev/eslint-config-rebeccastevens
    at Module._resolveFilename (node:internal/modules/cjs/loader:1075:15)
    at Function.resolve (node:internal/modules/cjs/helpers:116:19)
    at resolve (file:///home/rebec/dev/eslint-config-rebeccastevens/node_modules/.pnpm/knip@2.0.0-alpha.8/node_modules/knip/dist/util/require.js:5:48)
    at file:///home/rebec/dev/eslint-config-rebeccastevens/node_modules/.pnpm/knip@2.0.0-alpha.8/node_modules/knip/dist/index.js:144:42
    at Set.forEach (<anonymous>)
    at main (file:///home/rebec/dev/eslint-config-rebeccastevens/node_modules/.pnpm/knip@2.0.0-alpha.8/node_modules/knip/dist/index.js:123:32)
    at async run (file:///home/rebec/dev/eslint-config-rebeccastevens/node_modules/.pnpm/knip@2.0.0-alpha.8/node_modules/knip/dist/cli.js:26:46)
    at async file:///home/rebec/dev/eslint-config-rebeccastevens/node_modules/.pnpm/knip@2.0.0-alpha.8/node_modules/knip/dist/cli.js:57:1 {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/home/rebec/dev/eslint-config-rebeccastevens' ]
}

Node.js v18.15.0
```

RebeccaStevens:
fix #87


src/index.ts
```diff
@@ -172,9 +172,14 @@ export const main = async (unresolvedConfiguration: CommandLineOptions) => {
           const isHandled = deputy.maybeAddReferencedExternalDependency(workspace, packageName);
           if (!isHandled) collector.addIssue({ type: 'unlisted', filePath: containingFilePath, symbol: specifier });
 
-          // Pattern: @local/package/file
+          // Pattern: self-reference/import
           const otherWorkspace = chief.findWorkspaceByPackageName(packageName);
-          if (otherWorkspace && specifier !== packageName && workspace !== otherWorkspace) {
+          if (workspace.pkgName === otherWorkspace?.pkgName) {
```
RebeccaStevens, line 177: `workspace === otherWorkspace`

Doesn't work. The 2 workspaces look identical but aren't the same object.

src/util/require.ts
```diff
@@ -1,8 +1,9 @@
 import { createRequire } from 'node:module';
+import { join } from 'node:path';
 import { toPosix } from './path.js';
 import { timerify } from './performance.js';
 
-const require = createRequire(process.cwd());
+const require = createRequire(join(process.cwd(), 'package.json'));
```
RebeccaStevens, line 6: This doesn't seem safe. We probably need to actually look for the `package.json` file, and each workspace probably needs to refer to its own `package.json`.

src/index.ts
```diff
@@ -172,9 +172,18 @@ export const main = async (unresolvedConfiguration: CommandLineOptions) => {
...
           if (!isHandled) collector.addIssue({ type: 'unlisted', filePath: containingFilePath, symbol: specifier });
 
-          // Pattern: @local/package/file
+          // Pattern: self-reference/import
           const otherWorkspace = chief.findWorkspaceByPackageName(packageName);
-          if (otherWorkspace && specifier !== packageName && workspace !== otherWorkspace) {
+          if (
+            otherWorkspace !== undefined &&
+            workspace.dir === otherWorkspace.dir &&
+            workspace.pkgName === otherWorkspace.pkgName
```
RebeccaStevens, line 180: `workspace === otherWorkspace` Doesn't work. The 2 workspaces look identical but aren't the same object.

src/util/require.ts
RebeccaStevens: Made this change but it might be a bit of a hack.

src/typescript/ast-walker.ts
```diff
@@ -33,7 +34,18 @@ export const getImportsAndExports = (sourceFile: BoundSourceFile, options: Optio
       const filePath = module.resolvedModule.resolvedFileName;
       if (filePath) {
         if (module.resolvedModule.isExternalLibraryImport) {
-          if (module.resolvedModule.extension === '.d.ts') {
+          if (isSelfReferenceImport(workspace, specifier, filePath, module.resolvedModule.extension)) {
```
webpro, line 37: Perhaps too naive, but maybe we could simplify things by checking for `isInNodeModules(filePath)` here?

tests/fixtures/self-reference/node_modules/@sindresorhus/is/index.js
webpro, line 1: It seems like the `node_modules` are not necessary in this fixture?

src/index.ts
webpro: Oh that's embarrassing, great catch!

src/index.ts
```diff
@@ -172,9 +172,18 @@ export const main = async (unresolvedConfiguration: CommandLineOptions) => {
           const isHandled = deputy.maybeAddReferencedExternalDependency(workspace, packageName);
           if (!isHandled) collector.addIssue({ type: 'unlisted', filePath: containingFilePath, symbol: specifier });
 
-          // Pattern: @local/package/file
+          // Pattern: self-reference/import
```
webpro, line 175: Just to be sure I understand: the changes in this file are to accommodate self-references from dependencies coming from plugins, right? I mean this does not seem to affect the fixture added in this PR.

webpro (commented):
TIL about self-referencing a package using its name :) Hats off for picking this up in this oddly organized codebase, including tests & fixtures! Also glad you worked against v2.

src/index.ts
RebeccaStevens: I've added a new test fixture for this.

tests/fixtures/self-reference/node_modules/@sindresorhus/is/index.js
RebeccaStevens: No, they're not. The test just makes sure that the self-references aren't marked as unlisted dependencies.

src/typescript/ast-walker.ts
RebeccaStevens: I've simplified it even more that that. We don't need to check where the actual file is, just how it is being imported.

RebeccaStevens:
Windows test seems to be failing due to upstream issue https://github.com/esbuild-kit/cjs-loader/issues/33

webpro:
> The windows test seems to be failing due to upstream issue. Where exactly that issue is, I'm sure. It might be in node itself.

Yes, facing this too:

```
  error: |-
    Cannot find module './build'
    Require stack:
    - C:/Users/larsk/Projects/knip/tests/fixtures/workspaces-self-reference/packages/eslint-config-self-reference
  code: 'MODULE_NOT_FOUND'
  stack: |-
    Module._resolveFilename (node:internal/modules/cjs/loader:995:15)
    u.default._resolveFilename (C:\Users\larsk\Projects\knip\node_modules\@esbuild-kit\cjs-loader\dist\index.js:1:1519)
    Function.resolve (node:internal/modules/cjs/helpers:109:19)
    <anonymous> (c:\Users\larsk\Projects\knip\src\index.ts:187:40)
```

webpro:
I opened https://github.com/RebeccaStevens/knip/pull/1 to better describe what I meant and added some comments in the code as well. Did you see that? Perhaps a bit odd to open a PR this way, but alas :)

webpro:
Alright, let's go! I'm very grateful for your efforts @RebeccaStevens

PR #90, opened 2023-03-16
fix: add support for .d.cts and .d.mts declaration files

### LINKED ISSUE #89: Add support for `.mts` and `.cts` typescript files
Only `.ts` typescript files seem to work.

RebeccaStevens:
fix #89


webpro:
Thanks Rebecca!

PR #95, opened 2023-03-20
fix: include TEST_FILE_PATTERNS with empty entry

ojeytonwilliams:
Hey @webpro, we've been running into an issue where `knip` would flag our test files as being unused: https://github.com/freeCodeCamp/freeCodeCamp/actions/runs/4466295333/jobs/7844299421#step:6:193

From what I can see, this happens because our workspace config is 
```jsonc
    "client": {
// snipped
      "entry": [],
      "project": ["**/*.{js,ts,tsx}"],
// snipped 
    },
```

It was a relatively small change to get things working again, so I figured a PR was in order.

Hope this helps.

### DIFF
--- a/src/workspace-worker.ts
+++ b/src/workspace-worker.ts
@@ -130,8 +130,7 @@ export default class WorkspaceWorker {
 
   getEntryFilePatterns() {
     const { entry } = this.config;
-    if (entry.length === 0) return [];
-    return [entry, TEST_FILE_PATTERNS, this.negatedWorkspacePatterns].flat();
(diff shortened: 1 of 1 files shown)


webpro:
Thanks for this pull request, @ojeytonwilliams. Using an empty `entry: []` is really what I intend it to be: use no entry files. When omitted, it means to use the default config for `entry` files.

I dug a bit deeper and the issue seems to be something different. I think the issue can be fixed in the FCC config, so I'm going to close this one.

ojeytonwilliams:
Sure, I understand. Thanks for taking a look and for updating fCC's config.

PR #101, opened 2023-03-24
docs: add additional svelte import paths

vojtechsimetka:
SvelteKit suggests to put all non-route code to `src/lib/*` and uses`$lib/*` as import path. It also scans `.env` files and allows importing env variables through `$env/{dynamic,static}/{private,public}` so that they can be type checked and don't come as magic values. There is also `$service-worker` module but I've never used it so I don't feel comfortable suggesting path rewrite.

More info in https://kit.svelte.dev/docs/modules

### DIFF
--- a/docs/compilers.md
+++ b/docs/compilers.md
@@ -77,6 +77,8 @@ export default {
   paths: {
     // This ain't pretty, but Svelte basically does the same
     '$app/*': ['node_modules/@sveltejs/kit/src/runtime/app/*'],
+    '$env/*': ['.svelte-kit/ambient.d.ts'],
+    '$lib/*': ['src/lib/*'],
   },
   compilers: {
     svelte: async (text: string) => {


webpro:
Thanks for taking the time to open a PR for the docs.

In a first attempt I added that `$lib` path manually as well, but then I saw that `tsconfig.json` in the project root extends from `.svelte-kit/tsconfig.json` which contains has that in `paths` already, so I could remove it from the `paths` in the Knip config. In your experience, can you not remove that `$lib` path from the Knip config?

Either way, I'm afraid that `$env` path is necessary indeed.

vojtechsimetka:
You are right that `tsconfig.json` extends `.svelte-kit/tsconfig.json` and there are the `$lib` `paths` already. However, I just tested in fresh SvelteKit project and `$lib` path config is needed in latest `knip`. So it either never worked (the SvelteKit demo app imports just images from `$lib`) or there was some regression.

In the SvelteKit demo app I moved some of the components to `$lib` and here are the results:

**With `$lib` path**


**Without**

webpro:
Oh my, that was a regression in Knip indeed. I'm confident this worked properly before, perhaps changed something during v5 pre-releases. Going to merge this and leave only the `$env` thing. Thanks again!

PR #102, opened 2023-03-27
Update tailwindcss plugin with .cjs support

navin-moorthy:
If a project had `type: module` in `package.json`.

We had to provide `tailwind.config.cjs` to work.


webpro:
Thanks @navin-moorthy! Happy to merge this. Just looked into tailwind source code for the default supported extensions (does not seem to be mentioned in their docs), could you please add `mjs` and `ts` as well?

(source: https://github.com/tailwindlabs/tailwindcss/blob/3c4cfb527252aeece3a72deeec0e963ec26a8328/src/util/resolveConfigPath.js#L5-L8)

navin-moorthy:
Added them, seems the support for .mjs and .ts is coming in v3.3

PR #134, opened 2023-05-22
feat(postcss): enhance dependencies detection

oceandrama:
`next.js` [allows ](https://nextjs.org/docs/pages/building-your-application/configuring/post-css) to declare postcss-config like this
```
{
  "plugins": [
    "postcss-flexbugs-fixes",
    [
      "postcss-preset-env",
      {
        "autoprefixer": {
          "flexbox": "no-2009"
        },
        "stage": 3,
        "features": {
          "custom-properties": false
        }
      }
    ]
  ]
}
```

so I add more checks to detect packages

### DIFF
--- a/src/plugins/postcss/index.ts
+++ b/src/plugins/postcss/index.ts
@@ -18,9 +18,17 @@ const findPostCSSDependencies: GenericPluginCallback = async (configFilePath, { 
     : await load(configFilePath);
 
   return config?.plugins
-    ? (Array.isArray(config.plugins) ? config.plugins : Object.keys(config.plugins)).filter(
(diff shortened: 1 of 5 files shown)


webpro:
Thanks again, Ruslan. This isn't even documented at https://github.com/postcss/postcss, but then e.g. has https://stylelint.io/user-guide/postcss-plugin/ an example: `require("postcss-reporter")({ clearReportedMessages: true })` which is likely the same idea.

webpro:
:rocket: _This pull request is included in v2.12.1. See [Release 2.12.1](https://github.com/webpro/knip/releases/tag/2.12.1) for release notes._

PR #139, opened 2023-05-24
feat: add ignoreExportsUsedInFile option

### LINKED ISSUE #127: Allow ignoring unused exported types that are used by something exported
In TypeScript-land, I like exporting types corresponding to used values, even if those types aren't directly used by anything.

For example, `greet` here would be allowed but `GreetOptions` would be complained about:

```ts
// greet.ts
export interface GreetOptions {
  message: string;
  repeat?: number;
}

export function greet(options: GreetOptions) {
  console.log(options);
}
```

```ts
// index.ts
import { greet } from "./greet";

greet({ message: "Hello, world!" });
```

```plaintext
$ knip

Unused exported types (1)
GreetOptions  interface  greet.ts
```

Could we have an option to allow these used-indirectly-but-not-directly types please? I'd be happy to send a PR!

JoshuaKGoldberg:
Fixes #127.


src/ConfigurationChief.ts
```diff
@@ -51,6 +51,7 @@ const defaultConfig: Configuration = {
   ignore: [],
   ignoreBinaries: [],
   ignoreDependencies: [],
+  ignoreExportsUsedInFile: false,
```
JoshuaKGoldberg, line 54: This feels awfully specific for a high-level option, but I couldn't think of a better way to position it... 😕

webpro:
Thanks, Josh! Looks good at first sight. Will play a bit with it tomorrow. Saw a dangling `console.log` :)

JoshuaKGoldberg:
Ah, dangit :smile: thanks, removed.

tests/exports-in-file.test.ts
```diff
@@ -0,0 +1,33 @@
...
+import baseArguments from './helpers/baseArguments.js';
+import baseCounters from './helpers/baseCounters.js';
+
+const cwd = resolve('tests/fixtures/exports-in-file');
+
+test('Find unused exports respecting ignoreExportsUsedInFile', async () => {
+  const { issues, counters } = await main({
+    ...baseArguments,
+    cwd,
+  });
```
webpro, line 14: Shouldn't this include the `ignoreExportsUsedInFile: true` option?

webpro (commented):
From the initial RFC I was under the impression this flag would only be effective for types. Including exports/values is quite a difference, think tree-shaking (types are always removed).

I see the value of both, but I think it should be possible to allow this feature only for types. Which doesn't help the option/naming challenge 😅

What do you think?

tests/exports-in-file.test.ts
JoshuaKGoldberg: Ah, the tests weren't testing the way I thought they were. Sorry about that. Should be fixed now.

src/ConfigurationValidator.ts
```diff
@@ -34,6 +34,22 @@ const rootConfigurationSchema = z.object({
...
+          z.literal('enum'),
+          z.literal('function'),
+          z.literal('interface'),
+          z.literal('member'),
+          z.literal('type'),
+        ]),
+        z.boolean()
+      ),
+    ])
+    .optional(),
```
JoshuaKGoldberg, line 52: > possible to allow this feature only for types

Makes sense! I based these values on `SymbolType` (`exportedItem.type`), excluding `'unknown'`. WDYT?

webpro:
This is great! I love this config granularity. Thanks Joshua, much appreciated! I'm seeing some really minor things I might want to refactor later, but it's just implementation details. Happy to merge.

webpro:
:rocket: _This pull request is included in v2.13.0. See [Release 2.13.0](https://github.com/webpro/knip/releases/tag/2.13.0) for release notes._

JoshuaKGoldberg:
Yay, thanks! I'm excited to get a first big new change here 😄

Issue #72, opened 2023-03-03
Drop dependency on `bash-parser`


mcmxcdev:
We recently started using https://socket.dev/ as part of our dependencies maintenance and the tool has detected `bash-parser` as problematic due to being unmaintained for 6 years: https://socket.dev/npm/package/bash-parser

`bash-parser` itself also uses old and unmaintained dependencies, e.g. https://socket.dev/npm/package/deep-freeze which is a repo that has a non-existent npm account associated to it.

webpro:
It was difficult to even find such a package, let alone a recent or actively maintained one.

On the other hand, it provides functionality that makes Knip quite powerful in finding dependencies and entry files. [This test](https://github.com/webpro/knip/blob/main/test/tests/getReferencesFromScripts.test.ts) shows quite some cases it can handle, and the scripts can come from `package.json#scripts`, and the GitHub Action, Husky, Lefthook, lint-staged, Nx and release-it plugins. This is exactly the extra mile Knip likes to go.

I would also be happy to replace bash-parser with something more light-weight and well-maintained, but I'm not willing to consider losing this type of functionality. Perhaps it could be some combination of bash-parser becoming an optional dependency and opt-out by configuration or something.

Just for reference, I started out with [a bunch of regexes](https://github.com/webpro/knip/commit/1421206ac147bc86cfbea724cf46da0d6d47e7ad#diff-8c2f19134fb270f9726422ccec7e615bb3d2ce0d6fa1acacd339f9f91ac2b5ab), but that's far from elegant, limited and hard to maintain.

On the other hand, we're talking Bash here, syntax doesn't really change at all. It parses strings in configuration files. I wonder how can this package can actually be a threat or exploited? If it is I should perhaps add a warning to the docs.

Very open to ideas and suggestions here, I want Knip to be a good citizen in anyone's `node_modules`.

mcmxcdev:
Thanks for the detailed explanation!

Since knip is a dev dependency, I am not worried but I simply would like to get rid of unmaintained deps as much as possible. Reporting issues to multiple repositories to move the ecosystem forward.

No need to be actionable on this right away, let's keep it open for reference and maybe a better alternative will come around for the package or someone might have a recommendation!

camsaul:
It's causing errors for me

```
$ knip
Resolving plugin configuration files...
/home/linuxbrew/.linuxbrew/lib/node_modules/knip/node_modules/bash-parser/src/index.js:52
                throw new Error(err.stack || err.message);
[...]

PR #153, opened 2023-07-10
Replace `bash-parser` by `@ericcornelissen/bash-parser`

ericcornelissen:
Relates to #72 - Replaces [`bash-parser`](https://www.npmjs.com/package/bash-parser) by my own fork of the project [`@ericcornelissen/bash-parser`](https://www.npmjs.com/package/@ericcornelissen/bash-parser).

With this fork I aim to provide dependency maintenance. Thus far this only covers the removal of a non-OSI package, namely [`curry`](https://www.npmjs.com/package/curry), from the dependency tree.

Do note that this fork does not promise to provide anything beyond dependency maintenance. A proper replacement of `bash-parser` should be preferred if available.

### DIFF
--- a/@types/bash-parser/index.d.ts
+++ b/@types/bash-parser/index.d.ts
@@ -1,4 +1,4 @@
-declare module 'bash-parser' {
+declare module '@ericcornelissen/bash-parser' {
   type Name = {
     type: string;
     text: string;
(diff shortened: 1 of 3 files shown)


webpro:
Thank you, Eric! Very much appreciated. The unit tests pass, so I think a patch release is appropriate here.

PR #179, opened 2023-08-01
docs: document ci + show-progress

Zamiell:
For reference, it seems to be accomplished by the `process.stdout.isTTY` value, which is set to undefined in CI (i.e. GitHub Actions).


webpro:
:rocket: _This pull request is included in v2.17.3. See [Release 2.17.3](https://github.com/webpro/knip/releases/tag/2.17.3) for release notes._

PR #180, opened 2023-08-01
feat: silent in CI

Zamiell:
a test is failing but I'm not quite sure how to fix it.


webpro:
In the Action everything passes just fine :)

PR #192, opened 2023-08-05
fix(vitest): support default coverage provider

risu729:
Vitest has a default coverage provider, v8, but knip detects it as an unused dependency.
We can safely assume the user uses Vitest's coverage if the config file contains the `coverage` field, so this PR uses `v8` as the default provider if `provider` is not set.

https://vitest.dev/guide/coverage.html#coverage-providers

### DIFF
--- a/fixtures/plugins/vitest/vitest2.config.ts
+++ b/fixtures/plugins/vitest/vitest2.config.ts
@@ -1,0 +1,11 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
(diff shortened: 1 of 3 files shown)


webpro:
Thanks @risu729! Any chance you could add a (something to an existing) test?

risu729:
Sure! I added a test now.

webpro:
:rocket: _This pull request is included in v2.19.0. See [Release 2.19.0](https://github.com/webpro/knip/releases/tag/2.19.0) for release notes._

PR #197, opened 2023-08-11
webpack: handle a couple more cfg.entry formats

thomasjm:
This fixes a crash I was getting in my project based on Razzle.js, because the Webpack plugin code wasn't handling all of the cases that can happen in the configuration's `entry` field. (Specifically, when the entry is an object and its keys are of type `string[]`.)

This is also a little more defensively written and (IMO) easier to understand.

Based on the TS types here:

https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/types.d.ts#L2396-L2400

### DIFF
--- a/src/plugins/webpack/index.ts
+++ b/src/plugins/webpack/index.ts
@@ -70,15 +70,25 @@ const findWebpackDependencies: GenericPluginCallback = async (configFilePath, { 
       const dependencies = (config.module?.rules?.flatMap(resolveRuleSetDependencies) ?? []).map(loader =>
         loader.replace(/\?.*/, '')
       );
-      const entries = (
-        cfg.entry
-          ? typeof cfg.entry === 'string'
-            ? [cfg.entry]
-            : Array.isArray(cfg.entry)
-            ? cfg.entry
-            : Object.values(cfg.entry).map(entry => (typeof entry === 'string' ? entry : entry.filename))
-          : []
-      ).map(entry => (config.context ? join(config.context, entry) : entry));
+      let entries: string[] = [];
+
+      if (typeof cfg.entry === "string") entries.push(cfg.entry);
+      else if (Array.isArray(cfg.entry)) entries.push(...cfg.entry);
+      else if (typeof cfg.entry === "object") {
+        Object.values(cfg.entry).map(entry => {
+          if (typeof entry === "string") entries.push(entry);
+          else if (Array.isArray(entry)) entries.push(...entry);
+          else if (entry && typeof entry === "object" && "filename" in entry) entries.push(entry["filename"] as string);
+          else {
+            // TODO: warn about unrecognized single entry in object
+          }
+        });
+      } else {
+        // TODO: warn about unrecognized entry structure
+      }
+
+      entries = entries.map(entry => (config.context ? join(config.context, entry) : entry));
+
       return [...dependencies, ...entries];
     });
   });
--- a/src/plugins/webpack/index.ts
+++ b/src/plugins/webpack/index.ts
@@ -72,14 +72,14 @@ const findWebpackDependencies: GenericPluginCallback = async (configFilePath, { 
       );
       let entries: string[] = [];
 
-      if (typeof cfg.entry === "string") entries.push(cfg.entry);
+      if (typeof cfg.entry === 'string') entries.push(cfg.entry);
       else if (Array.isArray(cfg.entry)) entries.push(...cfg.entry);
-      else if (typeof cfg.entry === "object") {
+      else if (typeof cfg.entry === 'object') {
         Object.values(cfg.entry).map(entry => {
-          if (typeof entry === "string") entries.push(entry);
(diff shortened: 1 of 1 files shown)


src/plugins/webpack/index.ts
```diff
@@ -70,15 +70,26 @@ const findWebpackDependencies: GenericPluginCallback = async (configFilePath, {
...
+      if (typeof cfg.entry === "string") entries.push(cfg.entry);
+      else if (Array.isArray(cfg.entry)) entries.push(...cfg.entry);
+      else if (typeof cfg.entry === "object") {
+        Object.values(cfg.entry).map(entry => {
+          if (typeof entry === "string") entries.push(entry);
+          else if (Array.isArray(entry)) entries.push(...entry);
+          else if (typeof entry === "function") entries.push((entry as () => string)());
+          else if (entry && typeof entry === "object" && "filename" in entry) entries.push(entry["filename"] as string);
+          else {
+            // TODO: warn about unrecognized single entry in object
```
webpro, line 84: It's not Knip's responsibility to validate/lint the configuration of external tooling. The plugins just try to find dependencies.

src/plugins/webpack/index.ts
```diff
@@ -70,15 +70,26 @@ const findWebpackDependencies: GenericPluginCallback = async (configFilePath, {
...
-      const entries = (
-        cfg.entry
-          ? typeof cfg.entry === 'string'
-            ? [cfg.entry]
-            : Array.isArray(cfg.entry)
-            ? cfg.entry
-            : Object.values(cfg.entry).map(entry => (typeof entry === 'string' ? entry : entry.filename))
-          : []
-      ).map(entry => (config.context ? join(config.context, entry) : entry));
+      let entries: string[] = [];
```
webpro, line 73: Let's stay consistent, I prefer `const` over `let` where possible.

webpro (commented):
Thanks @thomasjm! Happy to merge, but please run `npm run format` first :)

src/plugins/webpack/index.ts
thomasjm: I wouldn't say it's about validation or linting, but instead making it easier to detect when Webpack changes/extends its configuration schema and the plugin doesn't recognize it. Rather than crash as happened in my case :P

I left this note in case `knip` has a preferred way of surfacing such warnings. But I can remove the comments if you like.

src/plugins/webpack/index.ts
thomasjm: I do too but I can't make it `const` and also reassign it further down.

src/plugins/webpack/index.ts
webpro: There's no handling of such config issues (I'd leave that to e.g. Webpack in this case), other than handling valid configuration.

src/plugins/webpack/index.ts
webpro: Then please make the re-assignment a new `const`

thomasjm: I was just going to suggest, how about this? https://github.com/thomasjm/knip/commit/cb190223a338ad51615783d402d8d32f03031f4a

webpro: That's fine as well

src/plugins/webpack/index.ts
thomasjm: Right -- I'm not sure you've understood me. The scenario here is that some configuration is perfectly valid according to Webpack, but hasn't been handled in Knip for some reason. So a warning would be useful to make it clear that Knip needs to handle a new case.

But it sounds like you don't want it, so I'll remove these branches.

src/plugins/webpack/index.ts
thomasjm: Ok done

src/plugins/webpack/index.ts
webpro: Thanks, what you describe is the ongoing process of what is Knip is all about, to keep iterating and improving based on fixes like yours. Unless you already have a specific case in mind that you know should be handled, the comment does not add much for me.

thomasjm: Oh well the suggestion was to turn the comment into a `console.warn`, or some other suitable way of surfacing a warning.

That way the next person who runs into the same scenario won't have to decipher an opaque error. (My error was an `undefined` string deep inside some stacktrace.)

thomasjm: But no worries either way, my use-case works after this fix so I'm happy.

webpro: Exception handling in general is something that should be improved indeed!

webpro:
:rocket: _This pull request is included in v2.19.3. See [Release 2.19.3](https://github.com/webpro/knip/releases/tag/2.19.3) for release notes._

webpro:
Thanks again Tom! 🙏

thomasjm:
No problem, thanks for the fast review!

PR #210, opened 2023-08-21
docs: fix typo

donaldxdonald:


webpro:
Thanks for raising this. This has always been the worst explanation, as the quote that's supposed to be Dutch is not actually Dutch. I'll fix it shortly, like so:

> Knip is Dutch for a "cut". A Dutch expression is "ergens ge**knip**t voor zijn", which means to be perfectly suited for the job. I'm motivated to make Knip perfectly suited for the job of cutting projects to perfection! ✂️

donaldxdonald:
Oh, I see! That's interesting!

PR #214, opened 2023-08-23
fix(readme): devDependencies

wojiangkuanglong:


webpro:
Thanks, but that's intentional and correct. It's also confusing, so I'll try to find a way to make it more clear.

The `devDependencies` key does not exist. By default, `dependencies` means both `dependencies` and `devDependencies`. In `--production` mode it means only `dependencies`.

So I'm going to close your PR.

PR #223, opened 2023-08-30
feat: add `knip.config.ts` and `knip.config.js` config files

ethanneff:
This will match the style of my other configs:

Screen Shot 42


webpro:
Thanks a lot for your PR, @ethanneff. Unfortunately I'm not going to merge it, but know that you can use `-c` or `--config` to use the file name you like. More words about the matter: https://github.com/webpro/knip/issues/182#issuecomment-1664316828

PR #206, opened 2023-08-18
Include declaration files when resolving import specifiers

webpro:
When a project has declaration files (`.d.ts`) added through e.g. the `include: []` configuration in `tsconfig.json`, how to get `ts.resolveModuleName()` take them into account?

This PR includes a fixture that has `assets.d.ts`:

```
declare module '*.html?raw';
```

With a module importing another:

```
import b from './block.html?raw';
```

The result with Knip is an unresolved module, since I don't know how to properly include the declaration files in the TypeScript backend. Full repro:

```
$ git clone -b fix/include-dts git@github.com:webpro/knip.git
$ cd knip
$ npm install
$ cd fixtures/dts/
$ npx knip
Unused files (1)
src/assets.d.ts
Unresolved imports (1)
./block.html?raw  src/index.ts
```

The TypeScript backend is set up mainly in the [constructor of the ProjectPrincipal class](https://github.com/webpro/knip/blob/93fecdabcdcc2324c69256bce3d886cbc3fb8046/src/ProjectPrincipal.ts#L72-L103) using a [language service and compiler host in createHosts.ts](https://github.com/webpro/knip/blob/main/src/typescript/createHosts.ts) with a [custom resolveModuleNames](https://github.com/webpro/knip/blob/main/src/typescript/resolveModuleNames.ts) function.

In this PR I have added some code and comments to indicate where the relevant bits are.

To rephrase my question: how to properly include declaration files not explicitly included through import statements?

(By the way, the current workaround in Knip is to let this slip through and eventually sanitize the import specifiers and ignore specifiers having an [extension from this list](https://github.com/webpro/knip/blob/main/src/constants.ts#L48-L67).)

PR #236, opened 2023-09-12
Fix entrypoints for Next.js plugin when using Next.js app router

### LINKED ISSUE #235: Unused exports are never detected with Next.js plugin in app router projects
https://github.com/webpro/knip/blob/838175cca9b6182cb8500ebf79a134167fd23b4e/src/plugins/next/index.ts#L16

When using the app router (where source code is in `app`), the only entrypoints possible are
- `instrumentation.{js,ts}`
- `middleware.{js,ts}`
- `app/{manifest,sitemap,robots}.{js,ts}`
- `app/**/{opengraph,twitter}-image.{js,ts,tsx}`, except for when any directory in the path starts with a `_`
- `app/**/{icon,apple-icon}-image.{js,ts,tsx}`, except for when any directory in the path starts with a `_`
- `app/**/{page,layout,error,not-found,template,loading}.{jsx,tsx}`, except for when any directory in the path starts with a `_`
- `app/**/route.{js,ts}`, except for when any directory in the path starts with a `_`
- All of those paths, but prefixed with `src/`

Right now the entrypoints in the Next.js plugin are effectively adding every single file in `app/**` as an entrypoint, which prevents Knip from catching any unused exports

jonahsnider:
Closes #235

### DIFF
--- a/fixtures/plugins/next/app/api/auth/route.ts
+++ b/fixtures/plugins/next/app/api/auth/route.ts
@@ -1,0 +1,5 @@
+import { helloWorld } from './util';
+
+export function GET() {
+  helloWorld();
+}
--- a/fixtures/plugins/next/app/api/auth/util.ts
+++ b/fixtures/plugins/next/app/api/auth/util.ts
@@ -1,0 +1,1 @@
+export function helloWorld() {}
--- a/fixtures/plugins/next/app/home/page.tsx
+++ b/fixtures/plugins/next/app/home/page.tsx
@@ -1,0 +1,3 @@
+import React from 'react';
+
+export default () => <main />;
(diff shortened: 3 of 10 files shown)


webpro:
Thanks @jonahsnider! Any chance you could maybe add some test coverage as well?

jonahsnider:
Yes, I can try adding some today/tomorrow most likely.

jonahsnider:
@webpro Added some tests, please let me know if there's anything else I should change

webpro:
Thanks! This is great, will release shortly

webpro:
:rocket: _This pull request is included in v2.24.1. See [Release 2.24.1](https://github.com/webpro/knip/releases/tag/2.24.1) for release notes._

webpro:
- For next time, there's a `npm run docs` to regenerate docs
- I removed the `next-env.d.ts` as that would hide a deeper issue (#206)

PR #245, opened 2023-09-18
Simplify map+flat call

Connormiha:
I did a little optimization. `map` + `flat` can be replaced by single `flatMap`.


webpro:
:rocket: _This pull request is included in v2.25.2. See [Release 2.25.2](https://github.com/webpro/knip/releases/tag/2.25.2) for release notes._

PR #246, opened 2023-09-18
Simplify array.from+map

Connormiha:
`Array.from` has optional second argument. Calling `map` is not required for this case.


webpro:
TIL! Thanks 🙏

PR #247, opened 2023-09-18
Optimize findManifestDependencies

Connormiha:
Spread creates a new array in each iteration, resulting in a complexity of O(n*n). Using `push` has a complexity of O(n) and does not create a new array.

### DIFF
--- a/src/manifest/index.ts
+++ b/src/manifest/index.ts
@@ -18,7 +18,7 @@ const findManifestDependencies = async ({ manifest, isProduction, isStrict, dir,
 
   const scripts = Object.entries(manifest.scripts ?? {}).reduce((scripts, [scriptName, script]) => {
     if (script && (scriptFilter.length === 0 || scriptFilter.includes(scriptName))) {
-      return scripts.push(script);
+      scripts.push(script);
     }
     return scripts;
   }, [] as string[]);
--- a/src/manifest/index.ts
+++ b/src/manifest/index.ts
@@ -18,7 +18,7 @@ const findManifestDependencies = async ({ manifest, isProduction, isStrict, dir,
 
   const scripts = Object.entries(manifest.scripts ?? {}).reduce((scripts, [scriptName, script]) => {
     if (script && (scriptFilter.length === 0 || scriptFilter.includes(scriptName))) {
-      return [...scripts, script];
+      return scripts.push(script);
     }
     return scripts;
   }, [] as string[]);


src/manifest/index.ts
```diff
@@ -18,7 +18,7 @@ const findManifestDependencies = async ({ manifest, isProduction, isStrict, dir,
 
   const scripts = Object.entries(manifest.scripts ?? {}).reduce((scripts, [scriptName, script]) => {
     if (script && (scriptFilter.length === 0 || scriptFilter.includes(scriptName))) {
-      return [...scripts, script];
+      scripts.push(script);
     }
     return scripts;
   }, [] as string[]);
```
homburg, line 24: FWIW, this is more or less what `flatMap`'s made for

```js
  const scripts = Object.entries(manifest.scripts ?? {}).flatMap(([scriptName, script]) => {
    if (script && (scriptFilter.length === 0 || scriptFilter.includes(scriptName))) {
      return [script];
    }

    return [];
  });
```

Connormiha: I think this code should be slower, because we still create new array.

src/manifest/index.ts
```diff
@@ -18,7 +18,7 @@ const findManifestDependencies = async ({ manifest, isProduction, isStrict, dir,
 
   const scripts = Object.entries(manifest.scripts ?? {}).reduce((scripts, [scriptName, script]) => {
     if (script && (scriptFilter.length === 0 || scriptFilter.includes(scriptName))) {
```
me4502, line 20: If you wanted to micro-optimise this for allocations a bit further, you could move `scriptFilter` out of the function and make it a set so that it has faster `.has()` checks, and also is only allocated once. Then rather than conditionally setting it to either `['start', 'posinstall']` or `[]`, move the `isProduction` check down here. So `scriptFilter` becomes `const scriptFilter = new Set(['start', 'postinstall']);` outside of the function, and this check becomes `if (script && (!isProduction || scriptFilter.has(scriptName))) {`

webpro:
Premature optimizations if you'd ask me (most time is spent during with disk IO and creating (and traversing) ASTs), it's still fun to implement and learn from each other here, so feel free to continue. I'm glad Knip receives also this kind of attention for detail! ❤️

I would prefer @me4502's suggestion, but tbh happy to merge any of the ideas here.

src/manifest/index.ts
Connormiha: I suggest making additions a bit later. I try not to make too many changes in the PR to reduce the likelihood of a bug. Moving the "isProduction" check is already a significant change, as it is already checked in another component. @webpro

webpro: That's fine, but don't worry too much! We have test, e.g. the `isProduction` flag with the `start` has coverage.

webpro:
Alright, merging this, thanks @Connormiha et al!

webpro:

PR #252, opened 2023-09-20
feat: add --preprocessor-options flag

me4502:
Adds a `--preprocessor-options` flag, to allow preprocessors to access separate options from reporters. I've opted to pass it in as a separate preprocessorOptions parameter, so that preprocessors still have access to the eventual reporter options.

I can change it to use the options field instead if preferred.

### DIFF
--- a/fixtures/cli-preprocessor/node_modules/with-args/index.ts
+++ b/fixtures/cli-preprocessor/node_modules/with-args/index.ts
@@ -1,0 +1,4 @@
+export default function (options) {
+  console.log(`hi from with-args preprocessor, you gave me: ${JSON.parse(options.preprocessorOptions).food}`);
+  return options;
+}
--- a/src/types/issues.ts
+++ b/src/types/issues.ts
@@ -58,6 +58,7 @@ export type ReporterOptions = {
   isProduction: boolean;
   isShowProgress: boolean;
   options: string;
+  preprocessorOptions: string;
 };
 
 export type Reporter = (options: ReporterOptions) => void;
--- a/tests/cli-preprocessor.test.ts
+++ b/tests/cli-preprocessor.test.ts
@@ -6,13 +6,8 @@ import { resolve } from '../src/util/path.js';
 const cwd = resolve('fixtures/cli-preprocessor');
 
 const exec = (command: string) => {
-  try {
-    const output = execSync(command.replace(/^knip/, 'node ../../dist/cli.js'), { cwd });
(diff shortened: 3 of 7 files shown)


fixtures/cli-preprocessor/node_modules/with-args/index.ts
```diff
@@ -0,0 +1,4 @@
+export default function (options) {
```
me4502, line 1: I put this into a package, because it turns out just adding a `with-args.ts` file was causing a non-zero exit code that took forever to debug, due to Knip finding an unused file :^)

src/types/issues.ts
```diff
@@ -58,6 +58,7 @@ export type ReporterOptions = {
   isProduction: boolean;
   isShowProgress: boolean;
   options: string;
+  preprocessorOptions: string;
```
webpro, line 61: The name is fine, I think I'll deprecate `options` and then fix that up in v3.

tests/cli-preprocessor.test.ts
```diff
@@ -25,3 +25,10 @@ test('knip --preprocessor knip-preprocessor', () => {
 test('knip --preprocessor @org/preprocessor', () => {
   assert.equal(exec('knip --preprocessor @org/preprocessor'), 'hi from scoped preprocessor');
 });
+
+test(`knip --preprocessor with-args --preprocessor-options {"food":"cupcake"}`, () => {
+  assert.equal(
+    exec(`knip --preprocessor with-args --preprocessor-options {\\"food\\":\\"cupcake\\"}`),
```
webpro, line 31: Perhaps more clear to use single quotes over escaping the double quotes? I don't mind it in general for tests, but some people use them as an example.

webpro (approved):
Great, just a minor nitpick, but happy to merge

fixtures/cli-preprocessor/node_modules/with-args/index.ts
webpro: Oh sorry about that, noted :)

tests/cli-preprocessor.test.ts
me4502: Not sure if this is due to a Windows shell oddity, but this was the only way I was able to get the tests actually working

webpro: Alright no problem

webpro:
Thanks, Maddy!

webpro:

PR #257, opened 2023-09-22
Mark .ts files as entrypoints inside Next.js app dir

Dremora:
I have a `page.ts` file that simply performs a redirect without involving JSX. It's also possible that a page simply re-exports a component defined elsewhere.

### DIFF
--- a/src/plugins/next/index.ts
+++ b/src/plugins/next/index.ts
@@ -18,8 +18,8 @@ const productionEntryFilePatternsWithoutSrc = [
   'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}',
   'instrumentation.{js,ts}',
   'app/{manifest,sitemap,robots}.{js,ts}',
-  'app/**/{icon,apple-icon}-image.{js,jsx,ts,tsx}',
-  'app/**/{opengraph,twitter}-image.{js,jsx,ts,tsx}',
+  'app/**/{icon,apple-icon}-image.{js,ts,tsx}',
+  'app/**/{opengraph,twitter}-image.{js,ts,tsx}',
   'pages/**/*.{js,jsx,ts,tsx}',
 ];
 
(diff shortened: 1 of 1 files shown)


src/plugins/next/index.ts
```diff
@@ -15,11 +15,11 @@ export const ENTRY_FILE_PATTERNS = ['next.config.{js,ts,cjs,mjs}'];
...
   'middleware.{js,ts}',
   'app/**/route.{js,ts}',
-  'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,tsx}',
+  'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}',
   'instrumentation.{js,ts}',
   'app/{manifest,sitemap,robots}.{js,ts}',
-  'app/**/{icon,apple-icon}-image.{js,ts,tsx}',
-  'app/**/{opengraph,twitter}-image.{js,ts,tsx}',
+  'app/**/{icon,apple-icon}-image.{js,ts,ts,tsx}',
+  'app/**/{opengraph,twitter}-image.{js,ts,ts,tsx}',
```
webpro, line 22: This `ts` extensions is a duplicate? I read `js,ts,ts,tsx` (twice)

webpro (commented):
Thanks, Kirill. Left one note, but happy to merge then :)

src/plugins/next/index.ts
jgoux: If you're using `minimatch`, this pattern usually gives you all the possible extensions for a js/ts project: `?([cm])[jt]s?(x)`.

webpro: And then some... ;) I like the explicitness and correctness here. This should only represent the extensions that Next.js (not TS) supports.

Dremora: Good catch, I think we should add jsx here.

webpro: No need for jsx, see e.g. https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image#generate-images-using-code-js-ts-tsx

webpro:
Thanks @Dremora! And sorry for the nitpicking.

webpro:

PR #262, opened 2023-09-23
Playwright for components plugin

jgoux:
Playwright now supports component testing which uses a different config + dependencies: https://playwright.dev/docs/test-components

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,7 @@ Knip contains a growing list of plugins:
 - [Nx][plugin-nx]
 - [nyc][plugin-nyc]
 - [Playwright][plugin-playwright]
+- [Playwright for components][plugin-playwright-ct]
 - [PostCSS][plugin-postcss]
 - [Prettier][plugin-prettier]
 - [Release It][plugin-release-it]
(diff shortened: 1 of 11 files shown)


webpro:
Thanks @jgoux, looks like a good addition. Any chance you could add some fixtures and test coverage for this?

jgoux:
Added fixtures and generated the docs. 👍

webpro:
Looked into their docs a bit, what do you think about splitting this up in a separate plugin (e.g. `playwright-ct`)?

I say this, because when I run `npm init playwright@latest -- --ct` I only get `@playwright/experimental-ct-react` so the `ENABLESRS` are also separate dependencies? Didn't actually use CT before, so I'm not sure :)

jgoux:
> Looked into their docs a bit, what do you think about splitting this up in a separate plugin (e.g. `playwright-ct`)?
> 
> I say this, because when I run `npm init playwright@latest -- --ct` I only get `@playwright/experimental-ct-react` so the `ENABLERS` are also separate dependencies? Didn't actually use CT before, so I'm not sure :)

It's pretty new to me as well, I tested removing `@playwright/test` and indeed it works with only `@playwright/experimental-ct-*`. I'll make the changes to split it into a new plugin. 👍

jgoux:
The feature is now isolated into its own `playwright-ct` plugin, and the CI is ✅, I guess it's ready for review @webpro. 😄

webpro:
Awesome! Thanks for this, great addition

webpro:
:rocket: _This pull request is included in v2.27.0. See [Release 2.27.0](https://github.com/webpro/knip/releases/tag/2.27.0) for release notes._

PR #265, opened 2023-09-25
add support for workspace file in vitest plugin

jgoux:
Docs: https://vitest.dev/guide/workspace.html


src/plugins/vitest/index.ts
```diff
@@ -14,13 +14,13 @@ export const ENABLERS = ['vitest'];
...
 
-export const CONFIG_FILE_PATTERNS = ['vitest.config.ts'];
+export const CONFIG_FILE_PATTERNS = ['vitest.config.ts', 'vitest.{workspace,projects}.{ts,js,json}'];
 
 // `TEST_FILE_PATTERNS` in src/constants.ts are already included by default
 export const ENTRY_FILE_PATTERNS = [];
 
 export const findVitestDeps = (config: VitestConfig) => {
-  if (!config || !config.test) return [];
+  if (!config || Array.isArray(config) || !config.test) return [];
```
webpro, line 23: If this means to ignore what's exported from the new files than it would be better to bail out as soon as we know `configFilePath`, perhaps even without loading the file at all?

src/plugins/next/README.md
```diff
@@ -19,15 +19,15 @@ or `devDependencies`:
       "app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}",
       "instrumentation.{js,ts}",
       "app/{manifest,sitemap,robots}.{js,ts}",
-      "app/**/{icon,apple-icon}-image.{js,ts,tsx}",
+      "app/**/{icon,apple-icon}.{js,ts,tsx}",
```
jgoux, line 22: This change is due to me regenerating the docs.

src/plugins/vitest/index.ts
```diff
@@ -30,8 +30,20 @@ export const findVitestDeps = (config: VitestConfig) => {
...
 
+const findVitestWorkspaceDeps = (config: VitestWorkspaceConfig) => {
+  let deps: string[] = [];
+  for (const cfg of config) {
+    if (typeof cfg === 'string') continue;
+    deps = [...deps, ...findVitestDeps(cfg)];
+  }
+  return compact(deps);
+};
+
```
jgoux, line 41: I realized that you can define vitest configs within the workspace's array. 😄

src/plugins/vitest/index.ts
jgoux: In the end I changed my implementation to stick to what's possible with the workspace config.

src/plugins/next/README.md
webpro: Yeah I just merged something else, forgot to add that. Thanks!

webpro (approved):

webpro:
Great, the plugins are having a good week here

PR #266, opened 2023-09-25
drizzle plugin

jgoux:
Docs: https://orm.drizzle.team/kit-docs/overview

I also added `playwright-ct` to the `schema.json` file.

### DIFF
--- a/fixtures/plugins/drizzle/drizzle.config.ts
+++ b/fixtures/plugins/drizzle/drizzle.config.ts
@@ -1,0 +1,5 @@
+import type { Config } from "drizzle-kit";
+
+export default {
+  schema: "./src/schema.ts",
+} satisfies Config;
--- a/fixtures/plugins/drizzle/drizzle.config.ts
+++ b/fixtures/plugins/drizzle/drizzle.config.ts
@@ -1,5 +1,5 @@
 import type { Config } from "drizzle-kit";
 
 export default {
-  schema: "./src/schema.ts",
+  schema: ["./src/schema.ts", "./src/schema/*.ts"],
 } satisfies Config;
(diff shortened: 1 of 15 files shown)


fixtures/plugins/drizzle/drizzle.config.ts
```diff
@@ -0,0 +1,5 @@
+import type { Config } from "drizzle-kit";
+
+export default {
+  schema: "./src/schema.ts",
```
jgoux, line 4: I think ideally we should mark this file as entrypoint as well, but not sure how to do that.

webpro: Inside `findDependencies` you can `load` the `configFilePath` (first argument) and then return the value `config.schema` in the array of dependencies.

jgoux: Thanks, on it. Which lib should I use to expand/process globs?

webpro: That should be mostly handled, the plugin can literally return e.g. `['.src/schema.ts']` and Knip will figure it out (e.g. abs/rel paths are entries, the rest is package names).

The first argument for `findDependencies` is `configFilePath` and that should be an absolute path already.

webpro: Here's a similar example: https://github.com/webpro/knip/blob/main/src/plugins/commitizen/index.ts#L25-L26

jgoux: `drizzle` accepts globs as well, I tried returning them but it didn't work, so I now process the paths with globby before returning the array, it did the trick. 👍

jgoux:
It should be good for a review. ✅

webpro:
Thanks, Julien! 🙏

webpro:
:rocket: _This pull request is included in v2.28.0. See [Release 2.28.0](https://github.com/webpro/knip/releases/tag/2.28.0) for release notes._

PR #273, opened 2023-09-29
feat: allow gatsby plugin to see local plugins as entrypoints

me4502:
Gatsby allows "local plugins" in the plugins folder that don't exist as full proper workspaces, but instead are run by Gatsby itself. This PR modifies the Gatsby plugin to include the "local plugins" folders by default, so these are included.

Plugins don't have a gatsby-config like the main site does, so I made the config entry only take the gatsby-node file.


webpro:
Thanks! TIL :)

PR #277, opened 2023-10-03
fix: readme dev dep flag

giulianok:
The correct key for `Unused devDependencies` is `devDependencies` (not `dependencies`)


webpro:
Thanks, but duplicate of #214. Sorry, still need to clarify that.

PR #285, opened 2023-10-07
feat: expose used in module 

utvara:
ts-prune supported detection for functions that are exported but used in module. This is especially useful when trying to detect functions/classes that are only exported for testing purposes

Example:
```bash
knip --production

....
Unused exports in namespaces (13)
getCommunicationRecipients                 src/api/email/db.ts                                             (used in module)
selectOneEmployeeByEmail                   src/api/employees/db.ts                                                         
bitePurposes                               src/api/journeys/bites/biteService.ts                           (used in module)
getBiteDetailsMetaInfo                     src/api/journeys/bites/biteService.ts                           (used in module)
....
```

Here selectOneEmployeeByEmail is not used at all (only used in tests) while getCommunicationRecipients is used within the module and tests.


webpro:
Thanks @utvara! I really appreciate you tried to make sense of the code and went in to improve it.

The idea I had was to use `--production --ignore-internal` to filter out things that are marked with the `/** @internal */` JSDoc tag and exported only for tests (https://github.com/webpro/knip#ignore-internal-exports). This way, there should be no false positives in `--production` mode.

What do you think?

webpro:
Also see https://github.com/webpro/knip#ignore-exports-used-in-file to ignore exports that are used in the same (non-entry) file.

utvara:
> The idea I had was to use `--production --ignore-internal` to filter out things that are marked with the `/** @internal */` JSDoc tag and exported only for tests (https://github.com/webpro/knip#ignore-internal-exports). This way, there should be no false positives in `--production` mode.
> 
> What do you think?

Yeah, ignore-internal-exports is what I was looking for, thanks for the clarification. Sorry for wasting your time.

webpro:
No worries! It's honestly another good indication Knip needs better docs. Again, thanks a lot for your interest in Knip and diving in.

PR #287, opened 2023-10-10
Add binaries to JSON reporter

Codex-:
I'm currently working on a simple github action that will comment on a PR with the report outputs from the JSON report and noticed that the `binaries` field was missing in the json blob.

Also noticed a small mistake in the documentation around the report keys

### DIFF
--- a/src/reporters/json.ts
+++ b/src/reporters/json.ts
@@ -52,6 +52,7 @@ export default async ({ report, issues, options }: ReporterOptions) => {
       ...(report.devDependencies && { devDependencies: [] }),
       ...(report.optionalPeerDependencies && { optionalPeerDependencies: [] }),
       ...(report.unlisted && { unlisted: [] }),
+      ...(report.binaries && { binaries: [] }),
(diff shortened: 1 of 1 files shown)


Codex-:
No problem, done :)

webpro:
:rocket: _This pull request is included in v2.33.1. See [Release 2.33.1](https://github.com/webpro/knip/releases/tag/2.33.1) for release notes._

Codex-:
Oh one thing that I'd love to get into the output would be the line number where relevant, my phase 2 for the action would be to use the output to annotate the sources on the pr, guessing we'd have to adapt the node position from the ts ast steps to relevant lines though

webpro:
Yes. The file position is easy to expose, the AST visitors already have pos. Only need to calc line + line pos from that. Will pick this up soon. Or maybe you are up to it?

Codex-:
I don't think I'd get to it for quite a while if you'd be happy with having a look? 🙏

webpro:
Of course, no worries!

Issue #201, opened 2023-08-16
Astro support


DaniFoldi:
Hi :wave:

I recently discovered knip, and love its simplicity! I was in the process of converting some projects to use knip, however I discovered that .astro is not currently supported.

Unfortunately, I don't have enough knowledge in the world of js compilers to integrate the compiler myself, however it seems that other than supporting the frontmatter syntax, creating a plugin should be rather straightforward, as astro is easy to find in a project (dev/dependency of `astro`), and entry files are simple to match (4 possible config files and src/pages/**).

webpro:
Extensions that are not JS/TS/JSX are not supported. Same for MDX, Vue, Svelte, etc.

Not an Astro user myself, However, adding the Astro as a custom compiler should be possible. I tried briefly before a few months back, but got stuck on the Astro compiler not playing nice with `esm-loader` (which was later replaced by jiti).

Perhaps even just some regexes to extract the import statements (and configuring the entry files you mention) might get you pretty far.

Did you see the examples in https://github.com/webpro/knip/blob/main/docs/compilers.md?

webpro:
We can close this now that @DaniFoldi implemented the plugin and a configuration example is documented.

PR #298, opened 2023-10-15
Add astro plugin

DaniFoldi:
Hi :wave:,

Following up my issue #201 I implemented a plugin that I believe matches the behaviour of Astro quite closely, following the docs for exact patterns.

The only remaining issue is ignoring the `astro:content` module, as it is resolved at build/runtime by astro, and not a separate dependency. Happy to fix if you could just point me in the right direction.


fixtures/plugins/astro/.vscode/extensions.json
webpro, line 1: File can be removed

fixtures/plugins/astro/.vscode/launch.json
webpro, line 1: Directory can be removed

fixtures/plugins/astro/README.md
webpro, line 1: File can be removed

fixtures/plugins/astro/knip.ts
```diff
@@ -0,0 +1,7 @@
+export default {
+  ignore: ['**/*.d.ts'],
```
webpro, line 2: This `ignore` option can be removed (I need to update the docs!)

fixtures/plugins/astro/knip.ts
```diff
@@ -0,0 +1,7 @@
...
+  compilers: {
+    astro: (text: string) => [...text.matchAll(/import[^;]+/g)].join('\n'),
+    css: (text: string) => [...text.matchAll(/(?<=@)import[^;]+/g)].join('\n'),
```
webpro, line 5: I think we also need a `mdx` compiler here (identical to `astro` to extract the `import` statements?)

fixtures/plugins/astro/package-lock.json
webpro, line 1: File can be removed

fixtures/plugins/astro/public/favicon.svg
webpro, line 1: Please minimize the number of images/fonts/binary files. Maybe we could even do with only the SVG?

There's [a number of extensions that are ignored](https://github.com/webpro/knip/blob/main/src/constants.ts#L46-L66) and tested elsewhere already.

fixtures/plugins/astro/src/content/blog/third-post.md
webpro, line 1: I think in general the `.md` files can be removed (or maybe just leave one small one), I think they're not relevant as they don't import anything and don't need to be compiled.

src/plugins/astro/index.ts
```diff
@@ -0,0 +1,17 @@
...
+export const NAME = 'astro';
+
+/** @public */
+export const ENABLERS = ['astro'];
+
+export const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(dependencies, ENABLERS);
+
+export const ENTRY_FILE_PATTERNS = ['astro.config.{js,cjs,mjs,ts}', 'src/content/config.ts'];
+
+export const PRODUCTION_ENTRY_FILE_PATTERNS = ['src/pages/**/*.{astro,md,mdx,html,js,ts}'];
```
webpro, line 15: I think the `.md` and `.html` files can be ignored, as they don't `import` anything? Maybe `.html` could be "compiled", but could be tricky to find `import` statements only within `<script>` tags? I'd be fine to leave that out for now.

There can also be `.mdx` files in `src/content` dir, so this would be my suggestion:

```suggestion
export const PRODUCTION_ENTRY_FILE_PATTERNS = ['src/pages/**/*.{astro,mdx,js,ts}', 'src/content/**/*.mdx'];
```

webpro (commented):
Thanks for putting this together!

I happen to have built an Astro site recently, so this was the perfect case to try it out.

Just a bit of clean up would be great. I think more irrelevant HTML and CSS can be removed. If you could cut some of that away, that would be great. Either way, I'll be happy to merge this.

The issue with `astro:content` is a little bit tricky, but I'll have a fix ready after the merge. If you like, you can also include it in this PR and add a colon (`:`) to the second group of the regex on this line: https://github.com/webpro/knip/blob/6cd5a1d4e9575ff2dee60e1ca6cc47cbf7788acd/src/util/modules.ts#L63

src/plugins/astro/index.ts
DaniFoldi: Good catch! I believe for imports to be "resolved" by astro, the extension has to be .astro anyways, so .html is just copied: https://docs.astro.build/en/core-concepts/astro-components/#html-components - and I believe same is true for .md, just translated into .html

fixtures/plugins/astro/.vscode/extensions.json
DaniFoldi: [`05e9895` (#298)](https://github.com/webpro/knip/pull/298/commits/05e9895c49f7d373caa4122b6d846f6235188860)

fixtures/plugins/astro/.vscode/launch.json
DaniFoldi: [`05e9895` (#298)](https://github.com/webpro/knip/pull/298/commits/05e9895c49f7d373caa4122b6d846f6235188860)

fixtures/plugins/astro/knip.ts
DaniFoldi: [`05e9895` (#298)](https://github.com/webpro/knip/pull/298/commits/05e9895c49f7d373caa4122b6d846f6235188860)

I saw a changelog entry recently, just copied the config from a project of mine, sorry about that.

fixtures/plugins/astro/package-lock.json
DaniFoldi: [`05e9895` (#298)](https://github.com/webpro/knip/pull/298/commits/05e9895c49f7d373caa4122b6d846f6235188860)

fixtures/plugins/astro/README.md
DaniFoldi: [`05e9895` (#298)](https://github.com/webpro/knip/pull/298/commits/05e9895c49f7d373caa4122b6d846f6235188860)

fixtures/plugins/astro/knip.ts
DaniFoldi: [`05e9895` (#298)](https://github.com/webpro/knip/pull/298/commits/05e9895c49f7d373caa4122b6d846f6235188860)

fixtures/plugins/astro/public/favicon.svg
DaniFoldi: I left only favicon.svg in, so in theory the fixture now has no binary files.

fixtures/plugins/astro/src/content/blog/third-post.md
DaniFoldi: Removed all but one .md file, and I trimmed that one down to a single paragraph.

DaniFoldi:
Thanks for your prompt response! I've implemented all suggestions from your review, and was about to also change the test (as astro:content is no longer reported as unlisted - but test are somehow worse now 😅, both locally and on a fresh clone in Codespaces, so I think it's best if I wait for your input on the current state before I mess up even more - my understanding of the codebase is still fairly limited.

webpro:
I've added two commits for the remaining bits :)

webpro:
Next step (for me) is to rebase from the `main` branch, because I did a major refactor just the other day (you couldn't know).

DaniFoldi:
Many thanks for your kind help! ✂️ 🚀

webpro:
:rocket: _This pull request is included in v2.34.0. See [Release 2.34.0](https://github.com/webpro/knip/releases/tag/2.34.0) for release notes._

PR #301, opened 2023-10-16
Add test suite for all current config loading systems

beaussan:
This PR introduce two things : 

A test suite to validate all config provider system, in a e2e manner :
- `knip` with a `knip.js` file that expose an object
- `knip` with a `knip.js` file that expose an async function
- `knip` with a `knip.json` file
- `knip -c knip.mjs` with a `knip.mjs` file that expose an object
- `knip -c knip.mjs` with a `knip.mjs` file that expose an async function
- `knip` with a config into a `package.json#knip`
- `knip` with a `knip.ts` file that expose an async function
- `knip` with a `knip.ts` file that expose a function
- `knip` with a `knip.ts` file that expose an object
- `knip -c knip.yaml` with a `knip.yaml` file

And enhance the current config loading mechanism to support js function & async function as a config export. I've added : 
- [x] Documentation
- [x] e2e test
- [x] Error handling
- [x] Debug of the error message if the function exposed failed 

If you need anything modified, feel free to reach out ! And thanks for your work on Knip :heart:

### DIFF
--- a/test/helpers/execKnip.ts
+++ b/test/helpers/execKnip.ts
@@ -1,15 +1,8 @@
 import { execSync } from 'node:child_process';
-import { fileURLToPath } from 'node:url';
-import { resolve } from '../../src/util/path.js';
+// eslint-disable-next-line n/no-restricted-import
+import { resolve } from 'node:path';
 
-const thisFileUrl = fileURLToPath(import.meta.url);
-
-const cliPath = resolve(
-  thisFileUrl,
-  // Needed if run in the /tmp (like in the npm test script) to find the correct dist folder
-  thisFileUrl.endsWith('tmp/test/helpers/execKnip.js') ? '..' : '.',
-  '../../../dist/cli.js'
-);
+const cliPath = resolve('dist/cli.js');
 
 export const execFactory = (cwd: string) => {
   return (command: string) => {


README.md
```diff
@@ -123,6 +123,19 @@ const config: KnipConfig = {
 export default config;
 ```
 
+And if you need, you can also expose a function or an async function if you need, like so :
```
webpro, line 126:
```suggestion
And if you need, you can also expose an (async) function if you need, like so:
```

src/ConfigurationChief.ts
```diff
@@ -159,6 +162,15 @@ export class ConfigurationChief {
     await this.setWorkspaces();
   }
 
+  private async loadResolvedConfigurationFile(configPath: string) {
+    const loadedValue = await _load(configPath);
+    try {
+      return await unwrapFunction(loadedValue);
+    } catch (e) {
+      throw new ConfigurationError(`Unable to run the function from the knip config file`);
```
webpro, line 170:
```suggestion
      throw new ConfigurationError(`Error running the function from ${configPath}`);
```

src/util/unwrapFunction.ts
```diff
@@ -0,0 +1,13 @@
+import { debugLogObject } from './debug.js';
+
+export const unwrapFunction = async (possibleFunction: unknown) => {
+  if (typeof possibleFunction === 'function') {
+    try {
+      return await possibleFunction();
+    } catch (error) {
+      debugLogObject('Error executing function:', error);
```
webpro, line 8: Could you please rebase from the `main` branch. Just before your PR, I've added a `context` argument so it would become like so:

```suggestion
      debugLogObject('*', 'Error executing function:', error);
```

The `'*'` might look a bit odd (had to come up with something representing "global" context), but now we can "assign" debug output to workspaces.

test/config/yaml.test.ts
```diff
@@ -0,0 +1,23 @@
...
+const exec = (command: string) => {
+  try {
+    const output = execSync(command.replace(/^knip/, 'node ../../../dist/cli.js'), { cwd });
+    return output.toString().trim();
+  } catch (error) {
+    if (error instanceof Error && 'stdout' in error && Buffer.isBuffer(error['stdout'])) {
+      return error['stdout'].toString();
+    }
+    // Unknown error
+    throw error;
```
webpro, line 17: This helper is great! Looks like it resolves https://github.com/webpro/knip/issues/253 :)

Will be very useful going forward and testing CLI arguments, etc.

This looks like using the same `exec` function in the new test files, do you think it could be a single reusable test helper? There's a `test/helpers` folder for this.

webpro (commented):
Thanks for putting this together, Nicolas!

I few comments, but I'm sure we can merge this swiftly.

test/config/yaml.test.ts
beaussan: Sure thing! Will do that ASAP 😁

beaussan: Done ! I've used a "dumb" way where you need to pass the cli path, this could be enhanced further where the execFactory crawls back up to find the cli itself, but for a first iteration, I opted to keeping it KISS. Let me know what you think !

test/cli-preprocessor.test.ts
```diff
@@ -1,18 +1,11 @@
...
 
-const exec = (command: string) => {
-  try {
-    const output = execSync(command.replace(/^knip/, 'node ../../dist/cli.js'), { cwd });
-    return output.toString().trim();
-  } catch {
-    console.error(`Error during execution of command: ${command}`);
-  }
-};
+const exec = execFactory(cwd, '../../dist/cli.js');
```
webpro, line 8: Would it be an idea to resolve `../dist/cli.js` in the helper module using `import.meta.url` from there? That would mean an argument less here :)

test/config/yaml.test.ts
webpro: Resolving the path to `cli.js` from the helper seems more ergonomic and clean, so would be great if you could still add that. Happy to merge either way!

test/cli-preprocessor.test.ts
beaussan: It can be done yes, but then if the files are in a different location (like the `npm run test` script), then there need to be extra logic

```ts
const thisFileUrl = fileURLToPath(import.meta.url);

const cliPath = resolve(
  thisFileUrl,
  // Needed if run in the /tmp (like in the npm test script) to find the correct dist folder
  thisFileUrl.endsWith('tmp/test/helpers/execKnip.js') ? '..' : '.',
  '../../../dist/cli.js'
);
```

The safest way is to go back up to find the package.json with the name `knip`, this way we are sure we are finding the right directory and is less prone to error if the test are moved around for example.

I've added the simple way, and I'll let you decide if it's worth the effort to look for the package.json

test/config/yaml.test.ts
beaussan: See the other comment, I've added it in the least intrusive way, since doing it the proper way could have an impact on test speed (fs apis to check for file and read files, for every test that uses the exec util)

test/cli-preprocessor.test.ts
webpro: Ah yes, that `/tmp` dir I forgot about.

Maybe something like `require.resolve('dist/cli.js)` to make an absolute path that should be OK from anywhere?

beaussan: It should be indeed ! Thank you for pushing the change directly to the branch :)

webpro: Hah, I was waiting for CI to pass 💚 

Wanted to get it done 🏓 Let's merge it?

Many many thanks for the great improvements!

beaussan: Let's merge it !

My pleasure ! I will experiment further with knip and the new dynamic capabilities ! I'll post my experiment on discord ;)

webpro:
:rocket: _This pull request is included in v2.34.0. See [Release 2.34.0](https://github.com/webpro/knip/releases/tag/2.34.0) for release notes._

PR #320, opened 2023-10-27
fix: bugged schema.json for JSONC

Zamiell:
this should be kind of a self-explanatory pull request.
JSONC allows trailing commas, so the schema is bugged.


webpro:
Thanks @Zamiell.

Where is this `allowTrailingCommas` specified? And wouldn't this be an issue for regular JSON files?

Zamiell:
> Where is this allowTrailingCommas specified?

[Sort of here.](https://github.com/microsoft/vscode/issues/102061#issuecomment-657080963)

> And wouldn't this be an issue for regular JSON files?

Does Knip parse files with a JSONC extension and a JSON extension differently? I would assume they get parsed with the same JSONC parser, which would mean that it is irrelevant if users put trailing commas into their "knip.json" files.

webpro:
It doesn't sit right with me. It means trailing commas are allowed in `.json` files.

I know TypeScript allows this in `tsconfig.json` and there are probably many more examples, but we are talking JSON Schema here. Personally I don't like this practice (just like rc files that can be either contain YAML or JSON).

Perhaps there should be a separate schema, like literally `schema.jsonc` that extends from `schema.json`?

Zamiell:
> It doesn't sit right with me. It means trailing commas are allowed in .json files.

That's true. Personally, I don't think it is an issue though, since as I said above, you are using the JSONC parser to parse JSON, so it is fine to encode that implementation detail into the schema itself. In some cases, it is bad for end-users to rely on implantation details, but for this case, this is not something I ever see changing. Even if you did change the implementation in the future, you could always update the schema at the same time to instantly invalidate trailing commas again.

> Perhaps there should be a separate schema, like literally schema.jsonc that extends from schema.json?

If that sits bitter with you, it seems reasonable to me. The downsides are discoverability though. i.e. I think a user story will be something like:

1) Alice starts using Knip in her TypeScript project.
2) She makes a "knip.json" file to hold her project-specific settings.
3) Over time, her project grows bigger and bigger, and so does her "knip.json" file.
4) Eventually, her "knip.json" file is so big that she wants to start putting comments in it in order to document all the exceptions she is making. Handily, she realizes that Knip supports JSONC out of the box, so she renames her "knip.json" to "knip.jsonc" and everything just works in the way it did before, and now she can add some comments without getting red squiggly lines in VSCode. (Side note: The red squiggly lines don't actually matter in this case, since the comments in the JSON file would be parsed perfectly fine by Knip.)
5) However, there is nothing to signal to Alice that merely the act of renaming ".json" to ".jsonc" also requires a new schema, [since similar tools like CSpell to not have this behavior](https://github.com/streetsidesoftware/cspell/pull/4930). Alice will be very confused when trailing commas do not work - especially if her auto formatter (i.e. Prettier) automatically inserts trailing commas and then she gets "unfixable" red squiggly lines in VSCode!

So I think you would have to signal this to end-users beyond a mere mention in the manual. I would actually go so far as to update the README.md to recommend using a "knip.jsonc" over a "knip.json" file to start with for brand new users who are onboarding into the tool. Specifically, [in this section](https://github.com/webpro/knip?tab=readme-ov-file#default-configuration) we could add the JSONC schema that brand-new users will be copy pasting, and [this section](https://github.com/webpro/knip?tab=readme-ov-file#configuration) with the JSONC over JSON recommendation.

Zamiell:
@webpro Have you had time to think about this? What is the current status of this PR?

webpro:
Like I said:

> Perhaps there should be a separate schema, like literally `schema.jsonc` that extends from `schema.json`?

I don't want to lure people into thinking their JSON is valid when it's not.

Zamiell:
Ok, I duplicated the schema as per your request.

Note that in addition to Alice's unresolvable Prettier errors that I outlined above, this decision has the maintenance burden of having to maintain two separate config files. Since this violates DRY, it's in danger of getting out of sync, so in the future we might want to add a lint rule to check for that, or alternatively programmatically create the schema from the actual knip source code. (That's what I do actually in my own CLI application that has a config file, I use "ts-json-schema-generator".)

webpro:
No, I don't fancy maintaining two separate files, that's why I said "extends from". This `schema.json` is also auto-updated when adding a plugin (I don't want to do that twice either).

Zamiell:
Ok, sorry about that, extending makes much more sense indeed. I tested this briefly and the "allOf" extension does seem to work properly

Zamiell:
We just have to remember to update the "v3" part when we bump a major version though. You would probably catch it though when doing a Ctrl+Shift+F to find the v3 string when updating the docs.

Anyways this should be ready for review now.

packages/docs/src/content/docs/overview/configuration.md
```diff
@@ -46,9 +46,9 @@ If your project structure does not match the default `entry` and `project`
 files, you can customize them. Here's an example configuration to include `.js`
 files in the `scripts` folder:
 
-```json title="knip.json"
+```json title="knip.jsonc"
 {
-  "$schema": "https://unpkg.com/knip@3/schema.json",
+  "$schema": "https://unpkg.com/knip@3/schema-jsonc.json",
```
webpro, line 51: Let's keep it simple for people new here. Other options JSONC and TypeScript are offered later on. So I think it's better not to introduce JSONC here.

packages/docs/src/content/docs/reference/configuration.md
```diff
@@ -9,14 +9,26 @@ configure Knip.
...
 
+In JSONC, you can use the provided JSONC schema:
+
+```json title="knip.json"
+{
+  "$schema": "https://unpkg.com/knip@3/schema-jsonc.json"
+}
+```
+
+(It is recommended to use JSONC over JSON so that you can use comments and trailing commas.)
```
webpro, line 30:
```suggestion
Use JSONC if you want to use comments and/or trailing commas.
```

packages/docs/src/content/docs/reference/configuration.md
```diff
@@ -9,14 +9,26 @@ configure Knip.
...
 
 ```json title="knip.json"
 {
   "$schema": "https://unpkg.com/knip@3/schema.json"
 }
 ```
 
+In JSONC, you can use the provided JSONC schema:
+
+```json title="knip.json"
```
webpro, line 24:
```suggestion
```json title="knip.jsonc"
```

packages/docs/src/content/docs/reference/configuration.md
```diff
@@ -9,14 +9,26 @@ configure Knip.
 
 ## JSON Schema
 
-In JSON, you can use the provided JSON Schema:
+A "@schema" field is a URL that you put at the top of your JSON file. This allows you to get red squiggly lines inside of your IDE when you make a typo or provide an otherwise invalid configuration option.
```
webpro, line 12:
```suggestion
A "$schema" field is a URL that you put at the top of your JSON file. This allows you to get red squiggly lines inside of your IDE when you make a typo or provide an otherwise invalid configuration option.
```

webpro (approved):
Alright, almost there 😃

If you accept my suggestions, could you also run `npm run remark`? Need to automate that somehow, I'll just run it otherwise, no problem.

PR #332, opened 2023-11-03
Suppress report for `@astrojs/check`

### LINKED ISSUE #329: `@astrojs/check` is reported as unused dependency
[`@astrojs/check`](https://www.npmjs.com/package/@astrojs/check?activeTab=readme), the library for [`astro check`](https://docs.astro.build/en/reference/cli-reference/#astro-check) command, is reported as unused dependency.

I would be happy to create a PR for this.
However, I would like to ask whether it is acceptable to add it without any conditions. Some users might add it and not use it, and they will not be able to realize that if it is ignored in the default of the Vitest plugin.

risu729:
fix #329

### DIFF
--- a/test/plugins/astro.test.ts
+++ b/test/plugins/astro.test.ts
@@ -13,7 +13,7 @@ test('Find dependencies in Astro configuration', async () => {
     cwd,
   });
 
-  assert(!issues.devDependencies['package.json']['@astrojs/check']);
+  assert(!issues.devDependencies['package.json']?.['@astrojs/check']);
 
   assert.deepEqual(counters, {
     ...baseCounters,
--- a/test/plugins/astro.test.ts
+++ b/test/plugins/astro.test.ts
@@ -8,10 +8,12 @@ import baseCounters from '../helpers/baseCounters.js';
 const cwd = resolve('fixtures/plugins/astro');
 
 test('Find dependencies in Astro configuration', async () => {
-  const { counters } = await main({
+  const { issues, counters } = await main({
     ...baseArguments,
     cwd,
   });
+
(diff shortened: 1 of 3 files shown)


test/plugins/astro.test.ts
```diff
@@ -8,11 +8,13 @@ import baseCounters from '../helpers/baseCounters.js';
 const cwd = resolve('fixtures/plugins/astro');
 
 test('Find dependencies in Astro configuration', async () => {
-  const { counters } = await main({
+  const { issues, counters } = await main({
     ...baseArguments,
     cwd,
   });
 
+  assert(!issues.devDependencies['package.json']?.['@astrojs/check']);
```
webpro, line 16: Many tests use `counters` to assert that there are no issues. So we don't need to assert the issues that are not there (you could verify this by removing either `@astrojs/check` from the dependencies or remove your implementation).

The conditional `?.` here is also a little bit confusing, does the test succeed or fail with or without the `['package.json']` key?

webpro (commented):
This really is a great first PR, thanks! One minor thing, but happy to merge then!

test/plugins/astro.test.ts
risu729: Thanks! Then, I will remove the assertion.

For the condition, I intended the test to succeed if the `package.json` key does not exist or exists but `@astrojs/check` does not exist.

webpro:
:rocket: _This pull request is included in v2.39.0. See [Release 2.39.0](https://github.com/webpro/knip/releases/tag/2.39.0) for release notes._

webpro:
Thanks again @risu729, let's go 🚀

PR #361, opened 2023-11-22
Use @latest for npx commands in doc

AbhiPrasad:
When developing https://github.com/getsentry/sentry-wizard we found that users ran into a lot of bugs because they weren't using the latest version of the package via the npx command.

This patch updates the docs to explicitly add `knip@latest` so that users will get the most up to date version of knip when copying the command from the docs.

### DIFF
--- a/packages/docs/src/content/docs/overview/getting-started.mdx
+++ b/packages/docs/src/content/docs/overview/getting-started.mdx
@@ -102,21 +102,21 @@ Alternatively, try Knip without adding it to your project:
   <TabItem label="npm">
 
     ```shell
-    npx knip
+    npx knip@latest
     ```
 
   </TabItem>
   <TabItem label="pnpm">
 
     ```shell
-    pnpm dlx knip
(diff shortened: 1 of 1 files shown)


webpro:
Thanks, @AbhiPrasad. I understand your concern, but I also think one of the goals of the `npx` command is to use the package/version installed in the project.

If the project does not yet contain the `knip` package and someone uses it for the first time it will install the latest version.

When there's already more existing API surface (a configuration file or command-line arguments), things might suddenly break after major version bumps of Knip. This won't happen with just `npx knip`.

Users should update their dependencies regularly, but that's another concern I guess.

AbhiPrasad:
My perspective here is that users who look at this from the docs often are trying to test stuff out for the first time, or coming back to the project after a while, so the docs recommending `@latest` will reduce the most amount of friction.

Usually when using npx internally (like with node scripts or in small helpers) just using whatever is installed makes sense.

Closing - thanks for the review!

PR #374, opened 2023-11-29
Pass ignoredWorkspacePatterns to mapWorkspaces ignore param (resolves…

### LINKED ISSUE #373: Package.json files having the same name property results in error even when added to ignoreWorkspaces
I [opened a PR](https://github.com/TanStack/query/pull/6195) to add an Angular adapter to TanStack query. Tried running Knip on [the PR branch](https://github.com/inshared/query/tree/feature/angular-query) but get this error:

```
Error: must not have multiple workspaces with the same name
package '@tanstack/angular-query-devtools-experimental' has conflicts in the following paths:
    /Users/arnoud/code/opensource/query/packages/angular-query-devtools-experimental
    /Users/arnoud/code/opensource/query/packages/angular-query-devtools-experimental/build
package '@tanstack/angular-query-experimental' has conflicts in the following paths:
    /Users/arnoud/code/opensource/query/packages/angular-query-experimental
    /Users/arnoud/code/opensource/query/packages/angular-query-experimental/build
```

Reason being that an Angular package that is built with ng-packagr has a 'source' package.json and in the build directory the one that will be published. So you end up with two package.json files having the same name. This is easily fixable by:
1. Adding a glob for the build directories to ignoreWorkspaces in the knip config
2. Knip should pass the ignoreWorkspaces patterns to the ignore param of mapWorkspaces

I tested this on my machine and it fixes the error, will submit a PR.

arnoud-dv:
Pass the ignoreWorkspaces patterns to the ignore param of mapWorkspaces to resolve #373 

Note that I did test that this prevents the error from occurring but did not perform other testing to make sure nothing breaks.

### DIFF
--- a/packages/knip/src/ConfigurationChief.ts
+++ b/packages/knip/src/ConfigurationChief.ts
@@ -270,6 +270,7 @@ export class ConfigurationChief {
     const workspaces = await mapWorkspaces({
       pkg: this.manifest ?? {},
       cwd: this.cwd,
+      ignore: this.ignoredWorkspacePatterns,
     });
 
     const manifestWorkspaces = new Map();


webpro:
Thanks, @arnoud-dv. Sorry you have to go through all this :) I'm not against this PR per se, but that error message is not coming from Knip.

That would be "Duplicate package name" (https://github.com/webpro/knip/blob/main/packages/knip/src/ConfigurationChief.ts#L323)

So I'm a bit surprised how modifying Knip's behavior would get you past that error?

(Apparently there's also something else saying it's not great to have duplicate workspace names...)

arnoud-dv:
Hi, did you see the issue I added for this? I describe it more in depth there.

arnoud-dv:
While looking at the Knip code and considering that it shouldn't break the configuration hints I couldn't think of a solution either.

I think I solved it for my PR though: the `pnpm-workspace.yaml` of TanStack Query contains double asterisk globs

```yaml
packages:
  - 'packages/**'
  ```
  
  Changing these to single ones fixes the problem and I think is actually an improvement because no real packages are nested inside other directories.

**Comment by webpro** [comment]
> While looking at the Knip code and considering that it shouldn't break the configuration hints I couldn't think of a solution either.
> 
> I think I solved it for my PR though: the `pnpm-workspace.yaml` of TanStack Query contains double asterisk globs
> 
> ```yaml
> packages:
>   - 'packages/**'
> ```
> 
> Changing these to single ones fixes the problem and I think is actually an improvement because no real packages are nested inside other directories.

You can use negations there too, btw:

```
packages:
  - 'packages/**'
  - '!packages/not-this-one'
  - '!packages/**/build/**'
```


> Knip's pretty cool btw! 😊

Thank you!

PR #378, opened 2023-12-04
vitest: Detect whether coverage is enabled in npm scripts

### LINKED ISSUE #372: vitest plugin: check npm scripts
Hi there,
since #347 the vitest plugin checks whether the `coverage.enabled` is set in the vitest config to optionally include the respective coverage reporter.
I prefer to run tests while implementing without coverage enabled and define a separate script in the `package.json` with coverage enabled, e.g.
```json
{
	"scripts" {
		"test": "vitest run",
		"coverage": "vitest run --bail=1 --coverage"
	}
}
```
As such it would be great if this method is also supported to optionally include the coverage reporter.
I am willing to submit a PR implementing this, if it is approved.

DMartens:
This fixes #372 by doing base searches for either `--coverage` or `coverage.enabled=true` for commands starting with `vitest`. This should be good enough but does not account for e.g. npx or for overriding the [coverage instrumenter](https://github.com/search?q=--coverage.provider&type=code) in the command line arguments.

<details>
	<summary>Unfortunately the "test" script in the root did not work for me</summary>

```js
node --no-warnings --test tmp

node:internal/modules/cjs/loader:1147
  throw err;
  ^

Error: Cannot find module '/home/dany/dev/forks/knip/packages/knip/tmp'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
    at node:internal/main/run_main_module:28:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```
</details>

But I tested by using the advice from the [guide](https://knip.dev/guides/writing-a-plugin/#tests) with `npx tsx test/plugins/vitest-npm-script.test.ts` run from `packages/knip` folder.

### DIFF
--- a/packages/knip/fixtures/plugins/vitest-npm-script/package.json
+++ b/packages/knip/fixtures/plugins/vitest-npm-script/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "@fixtures/vitest-npm-script",
+  "scripts": {
+    "test:coverage": "vitest --coverage"
+  },
+  "devDependencies": {
+    "vitest": "^0.34.6"
+  }
+}
--- a/packages/knip/fixtures/plugins/vitest-npm-script/vitest.config.ts
+++ b/packages/knip/fixtures/plugins/vitest-npm-script/vitest.config.ts
@@ -1,0 +1,8 @@
+export default {
+  test: {
+    coverage: {
+      enabled: false,
+      provider: 'v8',
+    },
+  },
+};
--- a/packages/knip/src/plugins/vitest/index.ts
+++ b/packages/knip/src/plugins/vitest/index.ts
@@ -5,6 +5,7 @@ import { hasDependency, load, tryResolve } from '../../util/plugin.js';
 import { toEntryPattern } from '../../util/protocols.js';
 import { getEnvPackageName, getExternalReporters } from './helpers.js';
 import type { ViteConfigOrFn, VitestWorkspaceConfig, ViteConfig, MODE, COMMAND } from './types.js';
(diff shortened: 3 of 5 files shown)


DMartens:
Hi,
as I said the `test` script does not seem to work even with manually running pretest ran from either the root or "packages/knip". The folder `tmp` is created with the transpiled files with a `src` and `test` subfolder. Maybe it has to do something with the node version (21.2.0) as running the file directly works: `node --test tmp/test/plugins/vitest-npm-script.test.ts`.
I changed the check to use a regular expression and formatted the code.

webpro:
> Hi, as I said the `test` script does not seem to work even with manually running pretest ran from either the root or "packages/knip". The folder `tmp` is created with the transpiled files with a `src` and `test` subfolder. Maybe it has to do something with the node version (21.2.0) as running the file directly works: `node --test tmp/test/plugins/vitest-npm-script.test.ts`.

I thought maybe you were running e.g. pnpm which does not execute `pre*` and `post*` scripts. Now I understand that Node.js v21 made a breaking change so we can do something like `node --test 'tmp/test/**/*.test.js'` to also work in Windows, which is great (but breaks current script).

> I changed the check to use a regular expression and formatted the code.

Thank you!

webpro:
:rocket: _This pull request is included in v3.5.0. See [Release 3.5.0](https://github.com/webpro/knip/releases/tag/3.5.0) for release notes._

PR #385, opened 2023-12-08
fix(schema): Added "binaries" to `issueTypes`

fnannestad:
The current schema identifies `binaries` as invalid when adding it the the knip config for `include` or `exclude`, however it is a valid option.


webpro:
Thanks @fnannestad! Now I see "unresolved" was also missing, which I'll add. Have a great day 🌞

PR #391, opened 2023-12-12
Deduplicate workspace names

### LINKED ISSUE #390: “Duplicate package name” error when using package at workspace root with pnpm
In a legacy project I’m using have a package at the repository/workspace root. When I run `knip` I get the error “Duplicate package name: \<root package name\>”.

**Reproduction**

```bash
mkdir knip-root-workspace
cd knip-root-workspace
pnpm init
cat >pnpm-workspace.yaml <<<"packages:
- .
- packages/*"
pnpm add -w knip
pnpm exec knip
```

knip version 3.7.0

I’ll try to provide a PR that fixes this

geigerzaehler:
It is possible that the manifest workspaces include the root workspace. In this case `ConfigurationChie#getAvailableWorkspaceNames()` would return the root name twice. This in turn would result in the root package being included twice and throw a `ConfigurationError` because of a duplicate package name.

To address this we deduplicate the workspace names.

I’d be happy to add some tests for this but I’d need some guidance where the best spot for these would be.

Fixes #390

### DIFF
--- a/packages/knip/src/ConfigurationChief.ts
+++ b/packages/knip/src/ConfigurationChief.ts
@@ -311,9 +311,9 @@ export class ConfigurationChief {
   }
 
   private getAvailableWorkspaceNames() {
-    return [ROOT_WORKSPACE_NAME, ...this.manifestWorkspaces.keys(), ...this.additionalWorkspaceNames].filter(
-      name => !micromatch.isMatch(name, this.ignoredWorkspacePatterns)
-    );
+    return [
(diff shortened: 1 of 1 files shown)


webpro:
Thanks, @geigerzaehler!

webpro:
:rocket: _This pull request is included in v3.7.1. See [Release 3.7.1](https://github.com/webpro/knip/releases/tag/3.7.1) for release notes._

PR #401, opened 2023-12-18
fix: use svelte plugin options from user config

### LINKED ISSUE #400: schema mismatch (missing svelte key)
svelte is included in the json [schema](https://github.com/webpro/knip/blob/ae91dac79e4d0d05ac3d10aefd4b49ecab61ed6d/packages/knip/schema.json?plain=1#L389-L392), but not in the zod [schema](https://github.com/webpro/knip/blob/ae91dac79e4d0d05ac3d10aefd4b49ecab61ed6d/packages/knip/src/ConfigurationValidator.ts?plain=1#L115-L116). Although it's not clear to me which one is correct; is the svelte plugin supposed to be editable? Even after adding svelte to the zod schema, adding patterns to `svelte.entry` in my config had no effect (but adding the same pattern in the plugin source code worked fine).

thenbe:
closes #400

### DIFF
--- a/packages/knip/src/ConfigurationValidator.ts
+++ b/packages/knip/src/ConfigurationValidator.ts
@@ -113,6 +113,7 @@ const pluginsSchema = z.object({
   storybook: pluginSchema,
   stryker: pluginSchema,
   stylelint: pluginSchema,
+  svelte: pluginSchema,
   tailwind: pluginSchema,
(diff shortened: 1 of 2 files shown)


SParth033 (approved):

webpro:
Thanks @thenbe, much appreciated!

PR #406, opened 2023-12-19
Add Next.js global-error entry file

nielsdB97:
The [global-error](https://nextjs.org/docs/app/api-reference/file-conventions/error#global-errorjs) entry file is missing in the Next.js plugin.


webpro:
Thanks, Niels! 🙏

nielsdB97:
Thank you for your efforts in building and maintaining this package! 🚀 
When can we expect the fix to be released? 🙏

webpro:
Not sure why it isn't auto-commented here, but it has been released in v3.8.4.

PR #411, opened 2023-12-21
docs: show an alternate method to compile svelte files

### LINKED ISSUE #410: knip does not detect type imports that happen in svelte files
In the [current](https://github.com/webpro/knip/blob/45ff78c8b5a2283a82e84967255093203c2cf3a5/packages/docs/src/content/docs/features/compilers.md?plain=1#L130-L136) svelte compiler recipe, we compile a svelte file straight to javascript, then pass the javascript to knip. As a result, knip never sees type imports that happen in svelte files. For example:

```typescript
// myfile.ts
export type Foo = any
export type Bar = any
export const baz = ''
```

```svelte
<script lang='ts'>
  import type { Foo } from './myfile'
  import { type Bar } from './myfile'
  import { baz } from './myfile'
</script>
```

Here, knip will report both `Foo` and `Bar` as "unused type exports" (assuming no other _typescript_ file also imports them).

thenbe:
closes #410

This PR adds a second, alternative recipe to the svelte compiler docs which enables Knip to accurately detect type imports that happen in svelte files.

### Implementation Details

- In svelte, you use an `export let myprop` to declare a [prop](https://svelte.dev/docs/svelte-components#script-1-export-creates-a-component-prop). So we replace all instances of `export let` with `let` to prevent knip from thinking those are actual exports.

- The implementation is admittedly hacky in that knip may see part of a file as duplicated. From testing, this does not adversely effect knip's ability to report issues. Regardless, this is why I'm suggesting it as an second alternative in the docs, rather than updating the current recipe. This way, svelte projects have a workaround at their disposal that does not require them to completely disable knip's unused type exports detection.

### Related

- The upcoming svelte v5 release will contain enhancements which could be relevant in the future: https://github.com/sveltejs/svelte/pull/9482
- Old thread in the svelte repo discussing type imports: https://github.com/sveltejs/svelte-preprocess/issues/318

### DIFF
--- a/packages/docs/src/content/docs/features/compilers.md
+++ b/packages/docs/src/content/docs/features/compilers.md
@@ -118,8 +118,6 @@ Use a configuration like this to compile non-standard files in Svelte projects:
 import sveltePreprocess from 'svelte-preprocess';
 import { preprocess, compile } from 'svelte/compiler';
 
-const sveltePreprocessor = sveltePreprocess();
-
 export default {
   paths: {
     // This ain't pretty, but Svelte basically does the same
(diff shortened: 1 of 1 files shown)


webpro:
Thanks @thenbe, this is great. Given how Svelte and Knip work, it's very useful to show both methods with pros and cons. Two things I'd like your opinion on:

1. The example having two methods might not be super clear to everyone, I can imagine separating them out in distinct code blocks with pros and cons explained above or below (I guess duplication is not a big deal here). If this all adds up being too lengthy we could move this to its own page.
2. Now that things are getting used more, perhaps it would be good to add fixtures with tests for both methods (even if it's example code)? Might also make iterations easier later on if changes in Svelte break our examples/hacks.

Just let me know if you agree, and if so, if you'd like to work on it. Otherwise I'm happy to merge this and iterate on it myself.

Thanks again! Really appreciate the thoughts and efforts here.

thenbe:
> Just let me know if you agree

I agree with both points. Being more explicit with the docs in the case feels like a good thing in this scenario as we're still exploring the best approach ourselves. That way, users who are interested can clearly see what approaches are available to them and can then provide feedback on each.

> Otherwise I'm happy to merge this and iterate on it myself.

I think that would be best as I'm not sure how exactly how to proceed on this one tbh, especially now that I might've introduced a [third](https://github.com/webpro/knip/issues/410#issuecomment-1868537396) method (I'm sorry).

PR #439, opened 2024-01-10
fix: plugins eslint-config-next is always reported as unused

minami-minami:
eslint-config-next is required when using `next`, or `next/core-web-vitals` in extends field.
However, eslint-config-next is reported as unused.


webpro:
Thanks for the PR, @minami-minami. From the Next.js docs:

> Next.js provides an ESLint plugin, [eslint-plugin-next](https://www.npmjs.com/package/@next/eslint-plugin-next), already bundled within the base configuration

This is why the exception is there in the code: `next` has `eslint-plugin-next` already as a dependency (and `eslint-plugin-next` should actually be a peer dependency of `next`).

So I think you can remove it from your `package.json`, unless there's another change or condition I've missed?

webpro:
This issue is haunting me. But your fix looks totally valid today.

I've cherry-picked your commit onto the `v4` branch, as I feel it might be a breaking change to in certain cases (but it's still your commit/contribution). It's now available in [4.0.0-canary.9](https://github.com/webpro/knip/releases/tag/4.0.0-canary.9), you can install using e.g. `npm install -D knip@canary`

Thanks again! Going to close this issue.

minami-minami:
@webpro 
Thank you!
I am looking forward to the release of v4 :)

And thanks for the great tools.❤️

PR #441, opened 2024-01-10
Update allowed files in Next.js

phibr0:
This allows nextjs route handlers to include JSX Syntax, I noticed that knip marks these files as unused, as I use JSX route handlers to send emails with react-email. 
I also noticed that there is no allowed pattern for the `default.{js,ts,tsx,jsx}` files yet.

## Commits

- fix: allow router handlers to be jsx files, add default.{js,jsx,ts,tsx}

### DIFF
--- a/packages/knip/src/plugins/next/index.ts
+++ b/packages/knip/src/plugins/next/index.ts
@@ -15,14 +15,13 @@ export const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDepen
 export const ENTRY_FILE_PATTERNS = ['next.config.{js,ts,cjs,mjs}'];
 
 const productionEntryFilePatternsWithoutSrc = [
-  'middleware.{js,ts}',
-  'app/**/route.{js,ts}',
+  'middleware.{js,jsx,ts,tsx}',
   'app/global-error.{js,jsx,ts,tsx}',
-  'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}',
-  'instrumentation.{js,ts}',
-  'app/{manifest,sitemap,robots}.{js,ts}',
-  'app/**/{icon,apple-icon}.{js,ts,tsx}',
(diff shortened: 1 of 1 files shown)


webpro:
Thanks @phibr0! So I'm seeing a few changes:

- Does Next.js handle `middleware.jsx`?
- Does it have docs for `app/**/route.{js,jsx,ts,tsx}`?
- Same for `app/**/default.{js,jsx,ts,tsx}`?

webpro:
I think I'd prefer no `jsx` or `tsx` for any of them, it's just weird? That it works doesn't mean it's a good idea, and the globs are already pretty extensive (maintenance, performance).

phibr0:
Yeah okay I understand, especially for the middleware as I cant see a reason you would need jsx in there.
I think the `default.js` should be included though.

I haven't looked into the docs that much yet, is there an easy way to extend this locally for my specific needs, I think i just define a plugin and add the `route.tsx` files there, right?

webpro:
Oh, sorry for not being clear, I'd love to add all the additions, except for the extensions ending with the `..x` :)

You can always customize it, see the `knip.json` with Playwright example: https://knip.dev/explanations/plugins#entry-files

Or perhaps even better, copy and modify the Next.js defaults into your own config: https://knip.dev/reference/plugins/next

webpro:
I can also just merge this one and fix it up? I'm cutting a release atm.

phibr0:
Sure!

webpro:
:rocket: _This pull request is included in v3.13.1. See [Release 3.13.1](https://github.com/webpro/knip/releases/tag/3.13.1) for release notes._

webpro:
Thanks, Phillip!

webpro:
:rocket: _This pull request is included in v4.0.0. See [Release 4.0.0](https://github.com/webpro/knip/releases/tag/4.0.0) for release notes._

PR #442, opened 2024-01-10
fix: document links

baseballyama:
A fews links are 404. So I fixed it out.


webpro:
Thanks @baseballyama! They seem to work fine, mainly looking at https://knip.dev/reference/configuration. Also in the editor the Markdown links are looking fine to me. So I wonder what you're seeing and why they are 404-ing?

baseballyama:
Please go to https://knip.dev/reference/configuration/#include and click `Rules & Filters.`.

baseballyama:
I found the way.
From searching feature, I can go to URL which has `/` before hash. 

image

webpro:
> Please go to https://knip.dev/reference/configuration/#include and click `Rules & Filters.`.

Ah that one is wrong indeed!

Thanks a lot for your work, I'm happy to merge your fixes when they're ready.

PR #444, opened 2024-01-10
docs: Remove slash before URL fragment

baseballyama:
I wanted to fix pagefind side, but I couldn't find the way, so I fixed the 404 issue at least for now.

### DIFF
--- a/packages/docs/src/content/docs/reference/plugins.md
+++ b/packages/docs/src/content/docs/reference/plugins.md
@@ -5,110 +5,57 @@ tableOfContents: false
 
 :::section{.columns.min200}
 
-- [Angular][1]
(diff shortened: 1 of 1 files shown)


webpro:
Thank you @baseballyama, but this file does not seem to contain 404s and is auto-generated and auto-formatted (here: https://github.com/webpro/knip/blob/v4/packages/docs/scripts/generate-plugin-docs.ts). That's why I'm going to close this issue :)

baseballyama:
@webpro 

Oh sorry for that. I'm not sure reason, but a completely unrelated file was pushed.
So I created one more PR so I would appreciate it if you could review this.
https://github.com/webpro/knip/pull/447

PR #449, opened 2024-01-12
using "forEach" instead of "map" as its return value is not being used

marco-carvalho:
# Description

## QA Checklist

- [ ] Try to author code and/or docs similar to the rest of the repository
- [ ] Run `npm run format` (from root)
- [ ] Run `npm test` (from root)


webpro:
Nice catch! Appreciated 🙏

webpro:
:rocket: _This pull request is included in v4.0.0. See [Release 4.0.0](https://github.com/webpro/knip/releases/tag/4.0.0) for release notes._

PR #451, opened 2024-01-15
use the 'String#endsWith' method instead

marco-carvalho:
In the case of checking if a file path ends with `rc`, `filePath.endsWith("rc")` is likely to be faster. This is because it's a straightforward string comparison without the need for the more complex pattern matching capabilities of regular expressions.


webpro:
Absolutely, thanks for the PR! Please follow the checklist and format the code (`npm run format`) and I'm happy to merge this.

webpro:
Thanks, Marco!

webpro:
:rocket: _This pull request is included in v4.0.2. See [Release 4.0.2](https://github.com/webpro/knip/releases/tag/4.0.2) for release notes._

PR #454, opened 2024-01-17
Add Eleventy data directory as entry pattern

### LINKED ISSUE #453: Eleventy plugin: ignore JavaScript "data" files
The Eleventy plugin exists, but doesn't resolve anything from the configuration. I'm hoping we can extract the `dirs.data` property returned from the configuration file function and count files in that directory as entry files, but unfortunately Eleventy configuration files export functions that take in a very specific class with many methods. Is there a way to lie to JavaScript about what methods exist on a class and just pass a dummy class?

uncenter:
Closes #453.

### DIFF
--- a/packages/knip/src/plugins/eleventy/helpers.ts
+++ b/packages/knip/src/plugins/eleventy/helpers.ts
@@ -1,5 +1,6 @@
 import type { EleventyConfig } from './types.js';
 
+// https://github.com/11ty/eleventy/blob/main/src/UserConfig.js
 export class DummyEleventyConfig {
   constructor() {}
   _getUniqueId() {}
@@ -139,6 +140,7 @@ export class DummyEleventyConfig {
   templateFormatsAdded = {};
 }
 
+// https://www.11ty.dev/docs/config/#configuration-options
 export const defaultEleventyConfig: EleventyConfig = {
   dir: {
     input: '.',
--- a/packages/knip/src/plugins/eleventy/helpers.ts
+++ b/packages/knip/src/plugins/eleventy/helpers.ts
@@ -149,5 +149,5 @@ export const defaultEleventyConfig: EleventyConfig = {
     layouts: '_includes',
     data: '_data',
   },
-  templateFormats: 'html,liquid,ejs,md,hbs,mustache,haml,pug,njk,11ty.js',
+  templateFormats: '11ty.js',
 };
--- a/packages/knip/src/plugins/eleventy/helpers.ts
+++ b/packages/knip/src/plugins/eleventy/helpers.ts
@@ -149,4 +149,5 @@ export const defaultEleventyConfig: EleventyConfig = {
     layouts: '_includes',
     data: '_data',
   },
+  templateFormats: 'html,liquid,ejs,md,hbs,mustache,haml,pug,njk,11ty.js',
 };
--- a/packages/knip/src/plugins/eleventy/helpers.ts
+++ b/packages/knip/src/plugins/eleventy/helpers.ts
@@ -1,150 +1,150 @@
-import type { EleventyConfig } from "./types.js";
+import type { EleventyConfig } from './types.js';
 
 export class DummyEleventyConfig {
-    constructor() {};
-	_getUniqueId() {};
-	reset() {};
-	versionCheck() {};
-	on() {};
-	emit() {};
-	_enablePluginExecution() {};
-	addMarkdownHighlighter() {};
-	addLiquidTag() {};
-	addLiquidFilter() {};
-	addNunjucksAsyncFilter() {};
-	addNunjucksFilter() {};
-	addHandlebarsHelper() {};
-	addFilter() {};
-	addAsyncFilter() {};
-	getFilter() {};
-	addNunjucksTag() {};
-	addGlobalData() {};
-	addNunjucksGlobal() {};
-	addTransform() {};
-	addLinter() {};
-	addLayoutAlias() {};
-	setLayoutResolution() {};
-	enableLayoutResolution() {};
-	getCollections() {};
-	addCollection() {};
-	addPlugin() {};
-	_getPluginName() {};
-	_executePlugin() {};
-	getNamespacedName() {};
-	namespace() {};
-	addPassthroughCopy() {};
-	_normalizeTemplateFormats() {};
-	setTemplateFormats() {};
-	addTemplateFormats() {};
-	setLibrary() {};
-	amendLibrary() {};
-	setPugOptions() {};
-	setLiquidOptions() {};
-	setNunjucksEnvironmentOptions() {};
-	setNunjucksPrecompiledTemplates() {};
-	setEjsOptions() {};
-	setDynamicPermalinks() {};
-	setUseGitIgnore() {};
-	addShortcode() {};
-	addAsyncShortcode() {};
-	addNunjucksAsyncShortcode() {};
-	addNunjucksShortcode() {};
-	addLiquidShortcode() {};
-	addHandlebarsShortcode() {};
-	addPairedShortcode() {};
-	addPairedAsyncShortcode() {};
-	addPairedNunjucksAsyncShortcode() {};
-	addPairedNunjucksShortcode() {};
-	addPairedLiquidShortcode() {};
-	addPairedHandlebarsShortcode() {};
-	addJavaScriptFunction() {};
-	setDataDeepMerge() {};
-	isDataDeepMergeModified() {};
-	addWatchTarget() {};
-	setWatchJavaScriptDependencies() {};
-	setServerOptions() {};
-	setBrowserSyncConfig() {};
-	setChokidarConfig() {};
-	setWatchThrottleWaitTime() {};
-	setFrontMatterParsingOptions() {};
-	setQuietMode() {};
-	addExtension() {};
-	addDataExtension() {};
-	setUseTemplateCache() {};
-	setPrecompiledCollections() {};
-	setServerPassthroughCopyBehavior() {};
-	addUrlTransform() {};
-	setDataFileSuffixes() {};
-	setDataFileBaseName() {};
-	getMergingConfigObject() {};
+  constructor() {}
+  _getUniqueId() {}
+  reset() {}
+  versionCheck() {}
+  on() {}
+  emit() {}
+  _enablePluginExecution() {}
+  addMarkdownHighlighter() {}
+  addLiquidTag() {}
+  addLiquidFilter() {}
+  addNunjucksAsyncFilter() {}
+  addNunjucksFilter() {}
+  addHandlebarsHelper() {}
+  addFilter() {}
+  addAsyncFilter() {}
+  getFilter() {}
+  addNunjucksTag() {}
+  addGlobalData() {}
+  addNunjucksGlobal() {}
+  addTransform() {}
+  addLinter() {}
+  addLayoutAlias() {}
+  setLayoutResolution() {}
+  enableLayoutResolution() {}
+  getCollections() {}
+  addCollection() {}
+  addPlugin() {}
+  _getPluginName() {}
+  _executePlugin() {}
+  getNamespacedName() {}
+  namespace() {}
+  addPassthroughCopy() {}
+  _normalizeTemplateFormats() {}
+  setTemplateFormats() {}
+  addTemplateFormats() {}
+  setLibrary() {}
+  amendLibrary() {}
+  setPugOptions() {}
+  setLiquidOptions() {}
+  setNunjucksEnvironmentOptions() {}
+  setNunjucksPrecompiledTemplates() {}
+  setEjsOptions() {}
+  setDynamicPermalinks() {}
+  setUseGitIgnore() {}
+  addShortcode() {}
+  addAsyncShortcode() {}
+  addNunjucksAsyncShortcode() {}
+  addNunjucksShortcode() {}
+  addLiquidShortcode() {}
+  addHandlebarsShortcode() {}
+  addPairedShortcode() {}
+  addPairedAsyncShortcode() {}
+  addPairedNunjucksAsyncShortcode() {}
+  addPairedNunjucksShortcode() {}
+  addPairedLiquidShortcode() {}
+  addPairedHandlebarsShortcode() {}
+  addJavaScriptFunction() {}
+  setDataDeepMerge() {}
+  isDataDeepMergeModified() {}
+  addWatchTarget() {}
+  setWatchJavaScriptDependencies() {}
+  setServerOptions() {}
+  setBrowserSyncConfig() {}
+  setChokidarConfig() {}
+  setWatchThrottleWaitTime() {}
+  setFrontMatterParsingOptions() {}
+  setQuietMode() {}
+  addExtension() {}
+  addDataExtension() {}
+  setUseTemplateCache() {}
+  setPrecompiledCollections() {}
+  setServerPassthroughCopyBehavior() {}
+  addUrlTransform() {}
+  setDataFileSuffixes() {}
+  setDataFileBaseName() {}
+  getMergingConfigObject() {}
 
-    _uniqueId = {};
-	events = {};
-	benchmarkManager = {};
-	benchmarks = {};
-	collections = {};
-	precompiledCollections = {};
-	templateFormats = {};
-	liquidOptions = {};
-	liquidTags = {};
-	liquidFilters = {};
-	liquidShortcodes = {};
-	liquidPairedShortcodes = {};
-	nunjucksEnvironmentOptions = {};
-	nunjucksPrecompiledTemplates = {};
-	nunjucksFilters = {};
-	nunjucksAsyncFilters = {};
-	nunjucksTags = {};
-	nunjucksGlobals = {};
-	nunjucksShortcodes = {};
-	nunjucksAsyncShortcodes = {};
-	nunjucksPairedShortcodes = {};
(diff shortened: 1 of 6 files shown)


packages/knip/src/plugins/eleventy/index.ts
```diff
@@ -18,6 +21,14 @@ const PRODUCTION_ENTRY_FILE_PATTERNS = ['posts/**/*.11tydata.js', '_data/**/*.{j
 const findEleventyDependencies: GenericPluginCallback = async (configFilePath, options) => {
   const { config } = options;
 
+  let localConfig = await load(configFilePath);
+  if (typeof localConfig === 'function') localConfig = await localConfig(new DummyEleventyConfig());
+
+  const result = { ...localConfig, ...defaultEleventyConfig };
+
+  // Data directory pattern:
+  toEntryPattern(`${join(result.dir.input, result.dir.data)}/**/*.js`);
```
uncenter, line 30: Here is the new pattern that we want to be considering. How should I factor this into the existing return logic below?

webpro: I've written some comments in the plugin template:

https://github.com/webpro/knip/blob/d8babee3fbb54d2e3fdf2fdb8f7dd55f52970d4d/packages/knip/src/plugins/_template/index.ts

The Playwright plugin seems similar to me:
 https://github.com/webpro/knip/blob/d8babee3fbb54d2e3fdf2fdb8f7dd55f52970d4d/packages/knip/src/plugins/playwright/index.ts

Two differences:

- The playwright plugin has an extra exported function to be reused by the playwright-ct plugin (you don't necessarily need to split it like that).
- The 11ty plugin should use `toProductionEntryPattern` (instead of `toEntryPattern`, sorry forgot about that before), as there's a difference between "dev dependencies" (like test frameworks target e.g. `*.test.ts` entry files) and "production dependencies" (plugins for Next.js, Remix, Gatsby use that too).

Eager to know whether the template comments together with https://knip.dev/guides/writing-a-plugin makes sense, and where it should be improved :)

uncenter: Thank you. How does it look now?

packages/knip/src/plugins/eleventy/helpers.ts
```diff
@@ -0,0 +1,153 @@
...
+// https://www.11ty.dev/docs/config/#configuration-options
+export const defaultEleventyConfig: EleventyConfig = {
+  dir: {
+    input: '.',
+    output: '_site',
+    includes: '_includes',
+    layouts: '_includes',
+    data: '_data',
+  },
+  templateFormats: 'html,liquid,ejs,md,hbs,mustache,haml,pug,njk,11ty.js',
```
uncenter, line 152: This is just the default value for this, I probably should remove all the other non-JS ones?

webpro: Yes, would be good to remove indeed. They would not actually get added eventually (there's a filter for unknown extensions), but it's just a waste.

packages/knip/src/plugins/eleventy/index.ts
```diff
@@ -13,14 +15,24 @@ const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(d
...
+  localConfig = { ...localConfig, ...defaultEleventyConfig };
+
+  return (
+    config?.entry ?? [
+      join(localConfig.dir.input, localConfig.dir.data, '**/*.js'),
+      join(
+        localConfig.dir.input,
+        `**/*.{${typeof localConfig.templateFormats === 'string' ? localConfig.templateFormats : localConfig.templateFormats.join(',')}}`
+      ),
+      join(localConfig.dir.input, '**/*.11tydata.js'),
```
webpro, line 32: Please keep using and exporting the `PRODUCTION_ENTRY_FILE_PATTERNS`, since the documentation is generated from that as well (it ends up here: https://knip.dev/reference/plugins/eleventy)

packages/knip/src/plugins/eleventy/index.ts
```diff
@@ -13,14 +15,24 @@ const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(d
...
-
 const findEleventyDependencies: GenericPluginCallback = async (configFilePath, options) => {
   const { config } = options;
 
-  return config.entry
-    ? config.entry.map(toProductionEntryPattern)
-    : [...ENTRY_FILE_PATTERNS.map(toEntryPattern), ...PRODUCTION_ENTRY_FILE_PATTERNS.map(toProductionEntryPattern)];
+  let localConfig = await load(configFilePath);
+  if (typeof localConfig === 'function') localConfig = await localConfig(new DummyEleventyConfig());
+  localConfig = { ...localConfig, ...defaultEleventyConfig };
```
webpro, line 23: Not sure whether this will work as intended, `defaultEleventyConfig` will always overwrite `localConfig` props?

packages/knip/test/plugins/eleventy.test.ts
```diff
@@ -1,21 +1,18 @@
...
-    processed: 2,
-    total: 2,
-  });
+  const configFilePath = join(cwd, 'eleventy.config.cjs');
+  const dependencies = await eleventy.findDependencies(configFilePath, { manifest, config });
+  assert.deepEqual(dependencies, [
+    'entry:_data/**/*.js',
+    'entry:**/*.{html,liquid,ejs,md,hbs,mustache,haml,pug,njk,11ty.js}',
+    'entry:**/*.11tydata.js',
+  ]);
```
webpro, line 17: I think it might be good to keep the other test in there, it doesn't hurt to keep it?

Also, would be great to have a fixture, used by a test, that includes a dynamic `eleventy.config.cjs` as the one you have yourself? Then we'll cover both the default scenario and the new more dynamic one.

webpro (commented):
Great, looks good so far but I do have a few comments.

packages/knip/src/plugins/eleventy/index.ts
uncenter: Oops yeah. Hmm wait I should merge this better anyway ahah, should I just do a lot of `localConfig.dir.input || defaultEleventyConfig.dir.input` or add something like `just-extend`?

packages/knip/src/plugins/eleventy/index.ts
webpro: The exported variables (`CONFIG_FILE_PATTERNS`, `ENTRY_FILE_PATTERNS`, `PRODUCTION_ENTRY_FILE_PATTERNS`) should contain the default values if there are any. I think Eleventy has default values, such as the `PRODUCTION_ENTRY_FILE_PATTERNS` currently in the `main` branch.

packages/knip/src/plugins/eleventy/index.ts
webpro: I'd prefer not to bring in a dependency for this. Maybe a mapper and/or a little helper function can do?

packages/knip/test/plugins/eleventy.test.ts
uncenter: Or does the name actually matter since I'm passing `configFilePath`? I could just make it like `eleventy.config.dynamic.cjs` right?

packages/knip/src/plugins/eleventy/index.ts
uncenter: https://github.com/webpro/knip/blob/main/packages/knip/src/plugins/eleventy/index.ts#L16
```js
const PRODUCTION_ENTRY_FILE_PATTERNS = ['posts/**/*.11tydata.js', '_data/**/*.{js,cjs,mjs}'];
```

This is incorrect and hard coded. Not sure why these would only be "production" unless I am misunderstanding this, and those don't take into account the `dir.input`. These paths are in fact still entry paths but are more correct now. For example, Knip was not applying the `_data/**/*.{js,cjs,mjs}` glob as my input directory is `src`.

packages/knip/test/plugins/eleventy.test.ts
webpro: Yes, that way you can even put multiple config files in a single fixture folder. Or create multiple fixture folders. Whatever makes sense :)

packages/knip/src/plugins/eleventy/index.ts
webpro: They're hardcoded, and serve as fallback/default values. If the user has `eleventy.entry` patterns and/or a custom `.eleventy.js` file then the default values won't be actually used.

They're "production" for `--production` mode in Knip, but will also be used in default mode. In contrast to regular `ENTRY_FILE_PATTERNS` which are not taken into account in production mode.

uncenter: Hmmm what? So is `findDependencies` not always used?

webpro: It's always used. Within `findDependencies` it's our job to figure it all out (also see the comments in the plugin template).

uncenter: Yes. Neither of those two paths are guaranteed or likely to work though... if `findDependencies` is always used, why do we need those?

webpro: For two reasons:

- A configuration file is often not required, and indeed seems not required for Eleventy
- Documentation is generated from it, so users can see the default values without having to look at source code.

uncenter: Ah, right. Well then I'll at least update those.

uncenter:
Will `PRODUCTION_ENTRY_FILE_PATTERNS` merge with the results of `findDependencies`?

webpro:
> Will `PRODUCTION_ENTRY_FILE_PATTERNS` merge with the results of `findDependencies`?

Not automatically, we'll have to do that inside `findDependencies`

uncenter:
I actually think we don't want to that, since it might accidentally count some files as entry files when they really aren't. More correct versions of all of the `PRODUCTION_ENTRY_FILE_PATTERNS` are detectable by `findDependencies`.

uncenter:
Right. So maybe instead of `if (!localConfig) return []`, I return `PRODUCTION_ENTRY_FILE_PATTERNS`?

webpro:
> Ah right. So maybe `if (!localConfig) return PRODUCTION_ENTRY_FILE_PATTERNS` instead of `if (!localConfig) return []`?

Exactly, in that case it should do something like it did without your PR :)

uncenter:
Done. Anything else?

packages/knip/src/plugins/eleventy/index.ts
```diff
@@ -13,14 +16,31 @@ const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(d
 
 const ENTRY_FILE_PATTERNS = ['.eleventy.js', 'eleventy.config.{js,cjs}'];
 
-const PRODUCTION_ENTRY_FILE_PATTERNS = ['posts/**/*.11tydata.js', '_data/**/*.{js,cjs,mjs}'];
+const PRODUCTION_ENTRY_FILE_PATTERNS = ['**/*.11tydata.js', '_data/**/*.{js,cjs,mjs}']
+  .map(pattern => [`src/${pattern}`, `content/${pattern}`])
```
webpro, line 20: Why are `src/` and `content/` added here? They don't seem to be defaults when I look in at https://github.com/11ty/eleventy/blob/main/src/defaultConfig.js but I might miss something?

packages/knip/src/plugins/eleventy/index.ts
```diff
@@ -13,14 +16,31 @@ const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(d
...
 const findEleventyDependencies: GenericPluginCallback = async (configFilePath, options) => {
   const { config } = options;
 
-  return config.entry
-    ? config.entry.map(toProductionEntryPattern)
-    : [...ENTRY_FILE_PATTERNS.map(toEntryPattern), ...PRODUCTION_ENTRY_FILE_PATTERNS.map(toProductionEntryPattern)];
+  let localConfig = (await load(configFilePath)) as
+    | Partial<EleventyConfig>
+    | ((arg: DummyEleventyConfig) => Promise<Partial<EleventyConfig>>);
+  if (!localConfig) return PRODUCTION_ENTRY_FILE_PATTERNS.map(toProductionEntryPattern);
```
webpro, line 29: This early return `if (!localConfig)` does not take into account `config.entry` anymore. I think we can return the same as before:

```
  return config.entry
    ? config.entry.map(toProductionEntryPattern)
    : [...ENTRY_FILE_PATTERNS.map(toEntryPattern), ...PRODUCTION_ENTRY_FILE_PATTERNS.map(toProductionEntryPattern)];
```

packages/knip/src/plugins/eleventy/index.ts
```diff
@@ -13,14 +16,31 @@ const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(d
...
+  const dataDir = localConfig?.dir?.data || defaultEleventyConfig.dir.data;
+  const templateFormats = localConfig.templateFormats || defaultEleventyConfig.templateFormats;
+
+  return (
+    config?.entry ?? [
+      join(inputDir, dataDir, '**/*.js'),
+      join(inputDir, `**/*.{${typeof templateFormats === 'string' ? templateFormats : templateFormats.join(',')}}`),
+      join(inputDir, '**/*.11tydata.js'),
+    ] ??
+    ENTRY_FILE_PATTERNS
```
webpro, line 42: I think this will never resolve to `ENTRY_FILE_PATTERNS` as the last `??` is behind an array. Looks like we can remove it?

packages/knip/src/plugins/eleventy/index.ts
```diff
@@ -13,14 +16,31 @@ const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(d
 
 const ENTRY_FILE_PATTERNS = ['.eleventy.js', 'eleventy.config.{js,cjs}'];
```
webpro, line 17: The `ENTRY_FILE_PATTERNS` should become `CONFIG_FILE_PATTERNS`, otherwise the results don't end up in `findDependencies` - now I realize maybe this is why the tests using `main()` didn't work properly?

webpro (commented):
This is looking great, almost there :)

packages/knip/src/plugins/eleventy/index.ts
uncenter: They aren't but the majority of sites I've encountered use that as the source directory (e.g. the official Eleventy blog starter uses `content/`, I and most of the sites I help with in the forum are using `src/`).

packages/knip/src/plugins/eleventy/index.ts
uncenter: That makes sooo much more sense!

packages/knip/src/plugins/eleventy/index.ts
uncenter: I guess we can just do the root directory if you want though. I imagine most people will have a configuration file though, so I don't foresee this being used much anyway.

webpro: Anything that's not default should come from the custom config, i.e. the return value of the `localConfig` function.

By the way, I often run `npm run watch` in the knip repo and then run `knip` from anywhere to test changes. Often that's the fixture folder for the thing I'm workin on. And it's often very insightful to run knip in some project where the code is relevant, in this case such as the ones you mention.

uncenter:
Wow that `npm run watch` tip was very helpful! I've reverted the change to `PRODUCTION_ENTRY_FILE_PATTERNS` now, and it works on my website! :D

webpro:
Thanks @uncenter!

By the way, I'm introducing Knip in the ESLint repo (https://github.com/eslint/eslint/pull/18005) and they're using Elventy for their docs (in the `/docs` folder). I think it might be interesting for users of Knip + 11ty to look if it's possible to extract out [that dependency](https://github.com/webpro/eslint/blob/introduce-knip/docs/.eleventy.js#L408):

```
./node_modules/algoliasearch/dist/algoliasearch-lite.esm.browser.js
```

Knip has function(s) to extract the `algoliasearch` from this so it does not need to be [ignoreDependencies](https://github.com/eslint/eslint/pull/18005/files#diff-48d5ba4681726b42e98dae10c08bd94f7f9836644c8f183c6475d10dcf67ebf1R40)

My question is, do you think even trying this could be interesting for more Eleventy users? We don't need to solve it for a single case of course, but if there's more interest we might be able to give it a shot.

But I guess that should not be part of this PR. This PR is ready, right? Happy to merge it :)

uncenter:
Yep, good to merge!

uncenter:
And yep, I've actually been working on that! That kind of pattern of passthrough copying files from node_modules is also in the 11ty-website repo. I'll have a go at it in another PR later today.

uncenter:
Entry files are prefixed with`entry:` and are dependencies just the name of the package with no prefix? Is that right?

webpro:
> Entry files are prefixed with`entry:` and are dependencies just the name of the package with no prefix? Is that right?

Yes, exactly. When we know for sure in the plugin we're dealing with entry files we can mark them as such. When we don't do that, Knip might still figure it out, but that's more error-prone and slower.

webpro:
:rocket: _This pull request is included in v4.0.4. See [Release 4.0.4](https://github.com/webpro/knip/releases/tag/4.0.4) for release notes._

PR #456, opened 2024-01-17
use the "RegExp.exec()" method instead

marco-carvalho:
`String.match()` behaves the same way as `RegExp.exec()` when the regular expression does not include the global flag `g`. While they work the same, `RegExp.exec()` can be slightly faster than `String.match()`. Therefore, it should be preferred for better performance.

### DIFF
--- a/packages/knip/src/util/tag.ts
+++ b/packages/knip/src/util/tag.ts
@@ -6,7 +6,7 @@ export const splitTags = (tags: string[]) =>
     .flatMap(tag => tag.split(','))
     .reduce<Tags>(
       ([incl, excl], tag) => {
-        (tag.startsWith('-') ? excl : incl).push(tag.match(/[a-zA-Z]+/)![0]);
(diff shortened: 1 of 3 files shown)


packages/knip/src/util/tag.ts
```diff
@@ -6,7 +6,7 @@ export const splitTags = (tags: string[]) =>
     .flatMap(tag => tag.split(','))
     .reduce<Tags>(
       ([incl, excl], tag) => {
-        (tag.startsWith('-') ? excl : incl).push(tag.match(/[a-zA-Z]+/)![0]);
+        (tag.startsWith('-') ? excl : incl).push(/[a-zA-Z]+/.exec(tag)![0]);
```
uncenter, line 9: If we are really itching for performance, RegEx is very slow: https://jsperf.app/tezuge. That might be premature optimization though ahah.

marco-carvalho: I don't think `String.includes()` works with Regex.

webpro:
I'm going to close this PR due to red CI lights and inactivity.

PR #462, opened 2024-01-19
Check for node_modules presence instead of isInternal for Eleventy plugin

uncenter:
Resolves https://github.com/webpro/knip/pull/460#issuecomment-1900344718. Would love to hear your thoughts on this.

### DIFF
--- a/packages/knip/fixtures/plugins/eleventy2/.eleventy.js
+++ b/packages/knip/fixtures/plugins/eleventy2/.eleventy.js
@@ -20,6 +20,9 @@ module.exports = function (eleventyConfig) {
   eleventyConfig.addShortcode('related_rules', () => {});
   eleventyConfig.addWatchTarget('./src/assets/');
   eleventyConfig.addPassthroughCopy({
+    'src/_includes/abc.js': '/assets/abc.js',
+  });
+  eleventyConfig.addPassthroughCopy({
     './src/static': '/',
   });
   eleventyConfig.addPassthroughCopy('./src/assets/');
--- a/packages/knip/src/plugins/eleventy/index.ts
+++ b/packages/knip/src/plugins/eleventy/index.ts
@@ -1,6 +1,6 @@
 import { DEFAULT_EXTENSIONS } from '../../constants.js';
 import { isDirectory } from '../../util/fs.js';
-import { dirname, isInternal, join } from '../../util/path.js';
+import { dirname, isInNodeModules, isInternal, join } from '../../util/path.js';
 import { timerify } from '../../util/Performance.js';
 import { hasDependency, load } from '../../util/plugin.js';
 import { toProductionEntryPattern } from '../../util/protocols.js';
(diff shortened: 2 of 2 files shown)


webpro:
Great! This is the fiddling that we sometimes need to get right. Ideally we don't need to actually glob or resolve modules here, because it could degrade performance and/or duplicate unnecessary work. But I've done it before, sometimes there's no way around it:

https://github.com/webpro/knip/blob/main/packages/knip/src/plugins/vitest/index.ts#L27-L33

webpro:
Happy to merge this! Any chance you could add something to an existing test/fixture to prevent regression later on?

uncenter:
> But I've done it before, sometimes there's no way around it:

Are you suggesting for me to do this? Not sure what this means.

> Any chance you could add something to an existing test/fixture to prevent regression later on?

Yep, I'm trying to think of how to do it...

webpro:
> Are you suggesting for me to do this? Not sure what this means.

No, sorry, just saying it's sometimes necessary and this is an example of when/how.
 
> Yep, I'm trying to think of how to do it...

What I often try to do is add something to a fixture that makes the related test fail. Then change the implementation to make the test pass. In this case you could add a `addPassthroughCopy()` call with that file (`src/_includes/components/throbber.js` or simpler in an existing folder).

uncenter:
Still not sure I get it. I'm not even positive I know why that was happening. How did I even realize this was an issue? 😆

webpro:
In the `main` branch, you can add something like this to e.g. `fixtures/plugins/eleventy2/.eleventy.js`:

```
eleventyConfig.addPassthroughCopy({
  'src/_includes/components/throbber.js': '/assets/throbber.js',
});
```

Run the test:

```
❯ npx tsx test/plugins/eleventy2.test.ts 
✖ Find dependencies in Eleventy configuration (static) (298.531542ms)
  AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
  + actual - expected ... Lines skipped
  
    {
      binaries: 0,
  ...
      total: 4,
      types: 0,
  +   unlisted: 1,
  -   unlisted: 0,
      unresolved: 0
    }
```

Or run `knip` in that fixtures directory:

```
lars@Neo ../knip/fixtures/plugins/eleventy2 fix/eleventy ❯ knip
Unlisted dependencies (1)
src/_includes/components/throbber.js  .eleventy.js
```

And so on :)

uncenter:
Ah, right, I forgot about that kind of test. Thank you, I'll do that.

uncenter:
Should be good now, I added that line and ran tests with the changes in this PR (fine) and without (error).

webpro:
:rocket: _This pull request is included in v4.2.0. See [Release 4.2.0](https://github.com/webpro/knip/releases/tag/4.2.0) for release notes._

webpro:
Thanks! I've also added that file to reflect it being "used".

PR #466, opened 2024-01-20
Add Netlify plugin

uncenter:
Adds a Netlify plugin!


packages/knip/src/plugins/netlify/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const ENABLERS = [/^@netlify\//, 'netlify-cli'];
+
+const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(dependencies, ENABLERS);
+
+const CONFIG_FILE_PATTERNS: string[] = ['netlify.toml'];
+
+const ENTRY_FILE_PATTERNS: string[] = [];
+
+const PRODUCTION_ENTRY_FILE_PATTERNS: string[] = [`netlify/functions/**/*.{${validFunctionExtensions()}}`];
```
webpro, line 21: Maybe we could list them explicitly here? Easier to read/update, Netlifiy does not necessarily follow Knip (or vice-versa). Also, looking at https://docs.netlify.com/functions/get-started/?fn-language=ts#module-format I only see `.mts`, `.cts` and `.ts`?

packages/knip/src/plugins/netlify/index.ts
```diff
@@ -0,0 +1,57 @@
...
+import { timerify } from '../../util/Performance.js';
+import { hasDependency, load } from '../../util/plugin.js';
+import { toProductionEntryPattern } from '../../util/protocols.js';
+import { extractFunctionsConfigProperty, validFunctionExtensions } from './helpers.js';
+import type { NetlifyConfig } from './types.js';
+import type { IsPluginEnabledCallback, GenericPluginCallback } from '../../types/plugins.js';
+
+// https://docs.netlify.com/ or https://docs.netlify.com/functions/get-started/?fn-language=ts
+
+const NAME = 'Netlify';
```
webpro, line 11: Maybe it's more appropriate and future-proof to use name `Netlify Functions` and dir `netlify-functions`?

packages/knip/src/plugins/netlify/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const ENABLERS = [/^@netlify\//, 'netlify-cli'];
```
webpro, line 13: Looking at the docs, `@netlify/functions` should be the only enabler?

packages/knip/package.json
```diff
@@ -71,6 +71,7 @@
     "picocolors": "1.0.0",
     "picomatch": "3.0.1",
     "pretty-ms": "8.0.0",
+    "smol-toml": "1.1.3",
```
webpro, line 74: Great, smol and dependency free :)

packages/knip/src/util/fs.ts
```diff
@@ -1,6 +1,7 @@
 import { readFileSync, statSync } from 'node:fs';
 import { readFile } from 'node:fs/promises';
 import yaml from 'js-yaml';
+import * as toml from 'smol-toml';
```
webpro, line 4: Please use named imports whenever available

webpro (commented):
Very cool! Never thought of a plugin like this, I love it. A few comments, but nothing spectacular I guess :)

packages/knip/src/plugins/netlify/index.ts
uncenter: Go to https://docs.netlify.com/functions/get-started/?fn-language=js#module-format-2. Notice how the URL now has `fn-language=js` instead of `fn-language=ts`!

packages/knip/src/plugins/netlify/index.ts
uncenter: I've actually changed this to not be specific to Netlify Functions as I read into the docs more, just Netlify in general.

packages/knip/src/plugins/netlify/index.ts
uncenter: Again, same as my previous reply, I've decided to go for a general plugin since I'm guessing many people use Netlify without using Netlify Functions.

packages/knip/package.json
uncenter: Glad you approve!

packages/knip/src/util/fs.ts
uncenter: I wasn't sure if importing `parse` would clutter the namespace.

uncenter:
With the addition of this PR we can also now support the `.prettierrc.toml` extension: https://github.com/webpro/knip/blob/main/packages/knip/src/plugins/prettier/index.ts#L18.

packages/knip/src/plugins/netlify/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const localConfig: NetlifyConfig | undefined = await load(configFilePath);
+  if (!localConfig)
+    return config.entry
+      ? config.entry.map(toProductionEntryPattern)
+      : PRODUCTION_ENTRY_FILE_PATTERNS.map(toProductionEntryPattern);
+
+  const dependencies = [
+    ...(localConfig?.plugins?.map(plugin => plugin.package) ?? []),
+    ...extractFunctionsConfigProperty(localConfig.functions || {}, 'external_node_modules'),
```
webpro, line 36: Could we use the existing `getValuesByKeyDeep` helper for this or is it different?

packages/knip/src/plugins/netlify/helpers.ts
```diff
@@ -0,0 +1,11 @@
+import type { FunctionsConfig } from './types.js';
+
+export const validFunctionExtensions = () =>
+  ['.js', '.mjs', '.cjs', '.ts', '.mts', '.cts'].map(ext => ext.slice(1)).join(',');
```
webpro, line 4: Sorry for the nitpicking, but this looks like it could just as well be literal strings/globs in `PRODUCTION_ENTRY_FILE_PATTERNS` directly?

packages/knip/src/plugins/netlify/index.ts
webpro: What is a more general plugin? Shouldn't we make a different plugin as the need arises? If it makes sense (e.g. there's different plugins for playwright/playwright-ct and vite/vitest as well).

packages/knip/src/plugins/netlify/index.ts
webpro: Ohh I missed that, thanks for clarifying.

webpro:
> With the addition of this PR we can also now support the `.prettierrc.toml` extension: https://github.com/webpro/knip/blob/main/packages/knip/src/plugins/prettier/index.ts#L18.

Yes, this is great!

packages/knip/src/plugins/netlify/index.ts
uncenter: I don't see why there should be - just extra code, parsing the same `netlify.toml` config file twice! Vite and Vitest are very different in terms of literally having separate config files and just in scope.

webpro: If we're dealing with the same config file, then I guess it makes sense. I just have no idea what "a more general plugin" could refer to.

packages/knip/src/plugins/netlify/index.ts
uncenter: It looks maybe similar? I can't get it to work though, just gives me an empty array.

packages/knip/src/plugins/netlify/index.ts
uncenter: Sorry, I meant "a plugin dealing with the `netlify.toml` configuration file" or something.

webpro: There seem to be more packages (like `@netlify/blobs` - there are  more), in which we shouldn't activate the plugin, right?

uncenter: Fair, I'll look more careful at which ones we should include.

uncenter:
Thanks for merging!

webpro:
:rocket: _This pull request is included in v4.2.0. See [Release 4.2.0](https://github.com/webpro/knip/releases/tag/4.2.0) for release notes._

webpro:
Thank you! Very good idea, great execution 🙏 

I did minor changes in your work after a major refactoring today, but nothing essential.

PR #484, opened 2024-01-30
Include `-r` param detection in docs

Faithfinder:


packages/docs/src/content/docs/features/script-parser.md
```diff
@@ -50,6 +50,10 @@ From this example, Knip automatically adds the following files as entry files:
 By the way, Knip would not add the `exports` if the `dist` folder is matching a
 pattern in a relevant `.gitignore` file or `ignore` option.
 
+### Scripts parsing
+
+When parsing `scripts` entry of `package.json`, `knip` also detects as dependencies parameters that follow `-r`, `--require` or `--loader`. This is done to correctly detect usage of, say, `dotenv` or `tsconfig-paths` packages.
```
voxpelli, line 55: Will this possibly be hard to keep in sync? If more parameters are auto-detected in the future?

webpro: Knip can handle this on a specific and generic/fallback basis. It's not very maintenance-friendly, but so far so good :)

PR #489, opened 2024-02-01
Add `jsx` and `tsx` in nextjs route entry

SandroMaglione:
With [`vercel/og`](https://www.npmjs.com/package/@vercel/og) you often create a `route.tsx` that export a `ImageResponse` with a react component. This PR adds `tsx` and `jsx` in the entry configuration for nextjs.

### DIFF
--- a/packages/knip/src/plugins/next/index.ts
+++ b/packages/knip/src/plugins/next/index.ts
@@ -1,6 +1,6 @@
+import type { GenericPluginCallback, IsPluginEnabledCallback } from '../../types/plugins.js';
 import { hasDependency } from '../../util/plugin.js';
 import { toEntryPattern, toProductionEntryPattern } from '../../util/protocols.js';
-import type { GenericPluginCallback, IsPluginEnabledCallback } from '../../types/plugins.js';
 
 // https://nextjs.org/docs/getting-started/project-structure
 
@@ -16,7 +16,7 @@ const productionEntryFilePatternsWithoutSrc = [
   '{instrumentation,middleware}.{js,ts}',
   'app/global-error.{js,jsx,ts,tsx}',
   'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}',
-  'app/**/{route,default}.{js,ts}',
+  'app/**/{route,default}.{js,jsx,ts,tsx}',
   'app/{manifest,sitemap,robots}.{js,ts}',
   'app/**/{icon,apple-icon}.{js,jsx,ts,tsx}',
   'app/**/{opengraph,twitter}-image.{js,jsx,ts,tsx}',


SandroMaglione:
@webpro are docs auto-generated from the configuration file or do we need to update another file somewhere?

webpro:
Thanks @SandroMaglione!

Looking at your tweet with the example config, and the `@vercel/og` docs with the `/pages/api/og.jsx` example, I would think those files would be matching one of the patterns that's already there:

```
'pages/**/*.{js,jsx,ts,tsx}'
```

So I wonder if we really need this change?

> @webpro are docs auto-generated from the configuration file or do we need to update another file somewhere?

The plugin docs are auto-generated!

SandroMaglione:
`'pages/**/*.{js,jsx,ts,tsx}'` only includes the `pages` directory.

For new projects that uses the `app` directory the current configuration doesn’t cover it:

スクリーンショット 2024-02-01 17 26 07

uncenter:
How does one configure the app vs pages directory? Couldn't we detect that setting and adjust our globs that way?

webpro:
> `'pages/**/*.{js,jsx,ts,tsx}'` only includes the `pages` directory.
> 
> For new projects that uses the `app` directory the current configuration doesn’t cover it:
> 
> スクリーンショット 2024-02-01 17 26 07

Sorry for delaying this. I wanted to ask what it is that we should cover:

- Specifically routes like `app/api/**/route.tsx`, OR
- More generic `app/**/*.tsx'

I'm not up to date on all the new Next.js features. https://nextjs.org/docs/getting-started/project-structure doesn't really cover this it seems. Any chance you could clear this up?

> How does one configure the app vs pages directory? Couldn't we detect that setting and adjust our globs that way?

Iirc, there's support (or a migration path that supports) both, so probably safer to just glob for both at the same time?

SandroMaglione:
Both `pages` and `app` can be used in the same app, so they should be both covered.

Routes can be defined also outside of `api`, so `app/api/**/route.tsx` won’t work ([documentation](https://nextjs.org/docs/app/building-your-application/routing/route-handlers#convention)). 

Inside `app` you can also define components, so probably `app/**/*.{js,jsx,ts,tsx}` is more safe. In special cases it’s possible to exclude files anyway.

webpro:
Another thing to consider is that we could add a new plugin specifically for this purpose: if the `@vercel/og` dependency is included, we'll add `app/api/og/route.tsx` as an entry file.

https://vercel.com/docs/functions/edge-functions/og-image-generation#getting-started

webpro:
:rocket: _This issue has been resolved in v4.5.0. See [Release 4.5.0](https://github.com/webpro/knip/releases/tag/4.5.0) for release notes._

webpro:
I've added a new plugin for this. Unfortunately still depends on `next` in case of app router where `next/og` is imported. Please let me know if there are any issues!

PR #500, opened 2024-02-07
Support husky v9

### LINKED ISSUE #495: lint-staged reported as unused with husky 9 after install
Forwarding from https://github.com/JoshuaKGoldberg/create-typescript-app/issues/1291:

```shell
gh repo clone JoshuaKGoldberg/create-typescript-app
cd create-typescript-app

pnpm i
pnpm lint:knip # No issues detected ✅

pnpx husky install
pnpm lint:knip # lint-staged is unused? ❌
```

The full output from that second Knip run is:

```plaintext
Analyzing workspace ....
Unused devDependencies (1)
lint-staged  package.json
 ELIFECYCLE  Command failed with exit code 1.
```

I'm a little stumped here 😕. I haven't had time to debug deep into Knip, sorry.

risu729:
This PR fixes #495. Fixture changes are based on the migration guide in the [release note](https://github.com/typicode/husky/releases/tag/v9.0.1).

This drops support for husky v8 or lower.
I'm unsure about the support policy for plugins, but I tried to keep it.
However, it requires a change in `CONFIG_FILE_PATTERNS` depending on the husky version, which seems impossible.


risu729:
Not much related to this PR, but I found an npm package [git-hooks-list](https://github.com/fisker/git-hooks-list) that might resolve the TODO here.

https://github.com/webpro/knip/blob/b49b0904fbb767287dfd55d697ed9b95feefc895/packages/knip/src/util/git.ts#L4-L10

webpro:
> This PR fixes #495. Fixture changes are based on the migration guide in the [release note](https://github.com/typicode/husky/releases/tag/v9.0.1).

Thanks @risu729, much appreciated!

> This drops support for husky v8 or lower.

These npm download stats are not in support of this idea, so I'm really hoping we can find a way to keep support for v8 as well: https://majors.nullvoxpopuli.com/q?packages=husky

> I'm unsure about the support policy for plugins, but I tried to keep it. However, it requires a change in `CONFIG_FILE_PATTERNS` depending on the husky version, which seems impossible.

Good point, I think Knip plugins should have a way of knowing the version number of the tool and adjust logic based on that. Maybe for now we can use `options.manifest` (`options` is the second argument each plugin receives) which contains the whole `package.json` object?

risu729:
Thank you for your comment!

I supported v8 or lower, but in a kind of hacky way.

It includes config files for both versions in `CONFIG_FILE_PATTERNS` because we can't get the version in the plugin definition step.
Then, it filters the config files in `findHuskyDependencies`.

It does not filter if the version is unknown or if the version range includes both versions.
At least now, in `husky@9.0.10`, knip does not report any error even if we check the `.husky/_/` generated in v9.

Also, I added `semver` to the dependencies to handle with ranges in `package.json`.

packages/knip/src/plugins/husky/index.ts
```diff
@@ -12,15 +13,35 @@ const ENABLERS = ['husky'];
 
 const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(dependencies, ENABLERS);
 
-const gitHookPaths = getGitHookPaths('.husky');
+const gitHooksPathInV8 = getGitHookPaths('.husky', true);
+// husky v9 registers hooks in .husky/_/ to git and calls user defined hooks in .husky/ from there
+const gitHookPathsInV9 = getGitHookPaths('.husky', false);
 
-const CONFIG_FILE_PATTERNS = [...gitHookPaths];
+// Add patterns for both v8 and v9 because we can't know which version is installed at this point
```
webpro, line 20: I understand the issue. This is OK for now.

Another solution we could think about is to split the plugin in two. Because the `isEnabled` function receives the `manifest` as well. So the "husky-v8" can be enabled separately from the "husky-v9" plugin.

But let's merge this plugin first, if you like this idea you could work on it later if you want.

webpro (approved):
Thanks for the fix, @risu729!

I'm happy to merge this, but is there any chance you please split the test and fixtures in two? One for v8 and one for v9?

risu729:
Sure! I will work on it soon.

packages/knip/src/plugins/husky/index.ts
risu729: Thanks for the advice! I didn't notice that.

I worked on it but faced an issue with the plugin names.
If I change them, the configuration of the husky plugin won't work the same way as it does now.

I came up with the idea to make an alias `husky` for `husky-v9` here, but I'm not sure it would work without any problems.
https://github.com/webpro/knip/blob/957e02a1f6ef8d92f7240a5a3c9d6e6e360c9ef0/packages/knip/src/plugins/index.ts#L17

Shall I open another PR for this?

webpro: Thanks for working on this!

> If I change them, the configuration of the husky plugin won't work the same way as it does now.

Good point!

It could be confusing for the user: how to name the config in `knip.json`, is it "husky8" or "husky-v8" or "husky"? Maybe it's not too bad that we handle the version differences in a single plugin (so we leave it as it is in `main` now). WDYT?

webpro:
Thanks a bunch @risu729, much appreciated! ❤️

packages/knip/src/plugins/husky/index.ts
risu729: I agree with the idea. Leave the single plugin as is might be the best way to avoid confusion.

webpro:
:rocket: _This pull request is included in v4.6.0. See [Release 4.6.0](https://github.com/webpro/knip/releases/tag/4.6.0) for release notes._

voxpelli:
Made a follow up to improve compatibility: https://github.com/webpro/knip/pull/537

PR #501, opened 2024-02-07
Ignore @types/bun by default

risu729:
As the same as `@types/node`, ignore `@types/bun`.

Since `bun-types` is not supposed to install directly as stated in the [npm package docs](https://www.npmjs.com/package/bun-types), we do not need to ignore `bun-types`.


webpro:
Thanks @risu729, agreed!

webpro:
:rocket: _This pull request is included in v4.5.0. See [Release 4.5.0](https://github.com/webpro/knip/releases/tag/4.5.0) for release notes._

PR #502, opened 2024-02-08
feat: improve markdown reporter

drodil:
- make the output valid for prettier
- add line+col to the location of issue
- add severity as it's own column

### DIFF
--- a/packages/knip/test/cli-reporter-markdown.test.ts
+++ b/packages/knip/test/cli-reporter-markdown.test.ts
@@ -16,16 +16,16 @@ test('knip --reporter markdown', () => {
 
 ## Unlisted dependencies (2)
 
-| Name            | Location     |
-|:----------------|:-------------|
-| unresolved      | src/index.ts |
-| @org/unresolved | src/index.ts |
+| Name            | Location     | Severity |
+| :-------------- | :----------- | :------- |
+| @org/unresolved | src/index.ts | error    |
+| unresolved      | src/index.ts | error    |
 
 ## Unresolved imports (1)
 
-| Name         | Location     |
-|:-------------|:-------------|
-| ./unresolved | src/index.ts |
+| Name         | Location          | Severity |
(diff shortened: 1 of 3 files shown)


drodil:
@webpro somehow when I run `npm run format` from the root, it messes up the prettier for lots of files. Is this something in my environment or are there really these files that need formatting?

packages/knip/test/cli-reporter-markdown.test.ts
```diff
@@ -16,16 +16,16 @@ test('knip --reporter markdown', () => {
...
+| unresolved      | src/index.ts | error    |
 
 ## Unresolved imports (1)
 
-| Name         | Location     |
-|:-------------|:-------------|
-| ./unresolved | src/index.ts |
+| Name         | Location            | Severity |
+| :----------- | :------------------ | :------- |
+| ./unresolved | src/index.ts (8:23) | error    |
```
webpro, line 28: TIL this format `src/index.ts (8:23)` can be copy-pasted and VS Code goes to that pos (I was used to only `src/index.ts:8:23`).

webpro (approved):
Thanks, I like it.

The formatting is odd indeed, not sure what causes this. Could you please undo only the formatting of `packages/docs/src/content/docs/features/reporters.md` (as I guess that results in a tall example visually in docs).

packages/knip/test/cli-reporter-markdown.test.ts
drodil: Oh yeah, I will change it

webpro: Well I'll leave it up to you, but I think your initial version is more readable (which is basically the purpose of Markdown reporter), and it's still easily copy-pastable.

webpro:
:rocket: _This pull request is included in v4.5.0. See [Release 4.5.0](https://github.com/webpro/knip/releases/tag/4.5.0) for release notes._

PR #506, opened 2024-02-11
Update Next.js plugin for needed exception - for mocking with next-router-mock

jmcpeak:
missing a needed exception - needed for mocking with next-router-mock


webpro:
Is this specific to Next.js or next-router-mock? Is it documented somewhere?

PR #507, opened 2024-02-11
Add missing jsx/tsx extensions to route/default files for Next.js

jmcpeak:
Add missing jsx/tsx extensions to route/default files for Next.js


webpro:
Thanks @jmcpeak! Did you see #489? Does v4.5.0 not solve it for you?

jmcpeak:
I'm running 5.0.0 and it still outputs these as unused

webpro:
The PR I linked to resulted in understanding that `@vercel/og` is the package that uses certain file paths so there's a new plugin for that.

Do you think it's Next.js itself using these paths you've added here or might another package use them?

jmcpeak:
> The PR I linked to resulted in understanding that `@vercel/og` is the package that uses certain file paths so there's a new plugin for that.
> 
> Do you think it's Next.js itself using these paths you've added here or might another package use them?

It is Next.js itself - I don't have [Open Graph (OG) Image Generation](https://vercel.com/docs/functions/og-image-generation) installed.

webpro:
Maybe another plugins/packages that might need it? I'm asking, last week ago I didn't find it in the Next.js docs and discovered it's another package that uses this file.

jmcpeak:
> Maybe another plugins/packages that might need it? I'm asking, last week ago I didn't find it in the Next.js docs and discovered it's another package that uses this file.

These files are used for parallel routes - https://nextjs.org/docs/app/api-reference/file-conventions/default

They need to support jsx/tsx

webpro:
Thanks for bearing with me, I see. Any chance you could split up `default` and `route` globs as the latter don't need to support JSX, right?

jmcpeak:
> Thanks for bearing with me, I see. Any chance you could split up `default` and `route` globs as the latter don't need to support JSX, right?

I will make the changes....

packages/knip/src/plugins/next/index.ts
```diff
@@ -16,7 +16,8 @@ const productionEntryFilePatternsWithoutSrc = [
   '{instrumentation,middleware}.{js,ts}',
   'app/global-error.{js,jsx,ts,tsx}',
   'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}',
-  'app/**/{route,default}.{js,ts}',
+  'app/**/{route}.{js,jsx,ts,tsx}',
+  'app/**/{default}.{js,ts,tsx}',
```
webpro, line 20:
```suggestion
  'app/**/default.{js,jsx,ts,tsx}',
```

packages/knip/src/plugins/next/index.ts
```diff
@@ -16,7 +16,8 @@ const productionEntryFilePatternsWithoutSrc = [
   '{instrumentation,middleware}.{js,ts}',
   'app/global-error.{js,jsx,ts,tsx}',
   'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}',
-  'app/**/{route,default}.{js,ts}',
+  'app/**/{route}.{js,jsx,ts,tsx}',
```
webpro, line 19:
```suggestion
  'app/**/route.{js,ts}',
```

webpro:
Thanks Jason! 😄

webpro:

PR #508, opened 2024-02-11
Add support for  cypress/support/commands.js file

jmcpeak:
https://docs.cypress.io/api/cypress-api/custom-commands

"We recommend defining queries is in your cypress/support/commands.js file, since it is loaded before any test files are evaluated via an import statement in the [supportFile](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Support-file)."


webpro:
Thanks, Jason. I don't think JSX is allowed in those files, i.e. we don't need to add those?

jmcpeak:
> Thanks, Jason. I don't think JSX is allowed in those files, i.e. we don't need to add those?

Good point, I think they can only be .js/ts files, I'll tidy it up

webpro:
Thanks again!

webpro:

PR #509, opened 2024-02-11
fix: make nx crystal work with standalone plugins

beaussan:
Nx plugins syntax allows for using standalone `string` identifiers and not only objects. This pr addresses that to ensure knip works with string and not only the object syntax.

[Here is the json schema for the plugin key in the `nx.json` config : 
](https://github.com/nrwl/nx/blob/master/packages/nx/schemas/nx-schema.json#L499)

```json
    "plugins": {
      "oneOf": [
        {
          "type": "string",
          "description": "A plugin module to load with default options"
        },
        {
          "type": "object",
          "properties": {
            "plugin": {
              "type": "string",
              "description": "The plugin module to load"
            },
            "options": {
              "type": "object",
              "description": "The options passed to the plugin when creating nodes and dependencies"
            }
          }
        }
      ]
    },
```

### DIFF
--- a/packages/knip/src/plugins/nx/index.ts
+++ b/packages/knip/src/plugins/nx/index.ts
@@ -28,9 +28,11 @@ const findNxDependenciesInNxJson: GenericPluginCallback = async configFilePath =
 
   const plugins =
     localConfig.plugins && Array.isArray(localConfig.plugins)
-      ? localConfig.plugins.map(it => getPackageNameFromModuleSpecifier(it.plugin)).filter(value => value !== undefined)
+      ? localConfig.plugins
+          .map(item => (typeof item === 'string' ? item : item.plugin))
+          .map(it => getPackageNameFromModuleSpecifier(it))
+          .filter(value => value !== undefined)
       : [];
-
   const generators = localConfig.generators
     ? Object.keys(localConfig.generators)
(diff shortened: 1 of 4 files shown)


packages/knip/src/plugins/nx/index.ts
```diff
@@ -28,9 +28,11 @@ const findNxDependenciesInNxJson: GenericPluginCallback = async configFilePath =
 
   const plugins =
     localConfig.plugins && Array.isArray(localConfig.plugins)
-      ? localConfig.plugins.map(it => getPackageNameFromModuleSpecifier(it.plugin)).filter(value => value !== undefined)
+      ? localConfig.plugins
+          .map(item => (typeof item === 'string' ? item : item.plugin))
+          .map(it => getPackageNameFromModuleSpecifier(it))
+          .filter(value => value !== undefined)
```
webpro, line 34: Nitpick: I usually skip this filter if there's a `compact()` on the final result (filtering out falsy values and duplicates).

webpro (approved):
Thank you! 👌

packages/knip/src/plugins/nx/index.ts
beaussan: Noted for next time! Thanks

webpro:

PR #512, opened 2024-02-13
fix truncated JSON output with JSON reporter

refractalize:
Hello - I noticed that when using `--reporter=json` on a project with a lot of unused code the JSON output was truncated. This seems to be due to this known issue, a combination of large output (> 65k) followed by `process.exit()`. See this issue: https://github.com/nodejs/node/issues/6379.

I've applied the workaround documented in that (and related) issues.

Let me know if you think I should write a test for it.

### DIFF
--- a/packages/knip/src/reporters/json.ts
+++ b/packages/knip/src/reporters/json.ts
@@ -98,6 +98,7 @@ export default async ({ report, issues, options }: ReporterOptions) => {
     issues: Object.values(json),
   });
 
+  // See: https://github.com/nodejs/node/issues/6379
   // @ts-expect-error _handle is private
   process.stdout._handle?.setBlocking?.(true);
   process.stdout.write(output);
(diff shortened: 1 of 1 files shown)


webpro:
Thanks @refractalize! I'm willing to just merge it, unless you have a simple/performant way to test this?

refractalize:
> unless you have a simple/performant way to test this?

Nothing that comes to mind. We could get some tests around it perhaps if we extract a function that writes output, and test that in isolation, but yeah, it'd be launching a new executable that writes a lot of stdout then exits immediately afterwards, not that elegant. What I might do is just add a comment with a link referring to the known issue in node.js.

webpro:
Thanks @refractalize!

webpro:
:rocket: _This pull request is included in v5.0.2. See [Release 5.0.2](https://github.com/webpro/knip/releases/tag/5.0.2) for release notes._

PR #513, opened 2024-02-13
Add Mock Service Worker - msw

jmcpeak:
https://mswjs.io/docs/integrations/browser


packages/knip/src/plugins/msw/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const ENABLERS = ['msw'];
+
+const isEnabled: IsPluginEnabledCallback = ({ dependencies }) => hasDependency(dependencies, ENABLERS);
+
+const ENTRY_FILE_PATTERNS = [  '**/mockServiceWorker.{js,ts}',
+  'mocks/browser.{js,ts}',
+  'mocks/handlers.{js,ts}"',
+  'mocks/index.{js,ts}"',
+  'mocks/server.{js,ts}"',
```
webpro, line 17: Are these the default locations?

Feels like files like those listed and `src/mocks/node.js` could be located anywhere without clear defaults, right? If most projects would follow a certain convention that could work too, but I'm not aware of any. Do you have any insights or links that could help here?

packages/knip/src/plugins/next/index.ts
```diff
@@ -16,7 +16,7 @@ const productionEntryFilePatternsWithoutSrc = [
   '{instrumentation,middleware}.{js,ts}',
   'app/global-error.{js,jsx,ts,tsx}',
   'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}',
-  'app/**/{route,default}.{js,ts}',
+  'app/**/{route,default}.{js,jsx,ts,tsx}',
```
webpro, line 19: This is not part of this PR, please remove.

packages/knip/fixtures/plugins/msw/msw.toml
```diff
@@ -0,0 +1 @@
+[[plugins]]
```
webpro, line 1: Does MSW allow for configuration files? If so, and if it could contain references to dependencies, the file patterns could added to `CONFIG_FILE_PATTERNS`. Otherwise this file does not serve any purpose and can be removed.

webpro (commented):
Thanks @jmcpeak! Thanks a lot for the PR. I'd love to add a plugin for MSW, but also wonder what would be the best default configuration for entry files (I've added some comments).

packages/knip/src/plugins/msw/index.ts
jmcpeak: The default everyone does is in a root directory of /mocks see all the examples - https://github.com/mswjs/examples/tree/main/examples/with-vitest - this is vitest, but its true for all examples.
And here is another example of the /mocks dir - https://mswjs.io/docs/integrations/browser#conditionally-enable-mocking

jmcpeak: I'm not 100% sure if I'm using the correct knip functions to check for these files. The tests pass and it works - just not sure.

jmcpeak: Also, if we could peek inside their package.json we could get the correct directory:

```
"msw": {
  "workerDirectory": "public"
}
```

packages/knip/fixtures/plugins/msw/msw.toml
jmcpeak: > Does MSW allow for configuration files?

Not that I know of, I'll remove it.

packages/knip/src/plugins/next/index.ts
jmcpeak: Will do

packages/knip/src/plugins/msw/index.ts
webpro: Thanks!

The `findReferences` function receives a second `options` arguments which contains `options.manifest`.

webpro: > I'm not 100% sure if I'm using the correct knip functions to check for these files. The tests pass and it works - just not sure.

Many tests are more "end to end" and use the `main()` function, then you can assert things on the returned `issues` and `counters` objects. See e.g. https://github.com/webpro/knip/blob/main/packages/knip/test/plugins/ava2.test.ts

webpro (commented):
Looks like it comes down to adding (only) `${package.json#msw.workerDirectory}/mockServiceWorker.js` into the process :)

packages/knip/test/plugins/msw.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  assert.deepEqual(
+    issues.files,
+    new Set([
+      join(cwd, 'public/mockServiceWorker.js'),
+      join(cwd, 'src/mocks/browser.ts'),
+      join(cwd, 'src/mocks/handlers.ts'),
+      join(cwd, 'src/mocks/index.ts'),
+      join(cwd, 'src/mocks/server.ts'),
+    ])
```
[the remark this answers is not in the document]
jmcpeak: > Seems like the msw.workerDirectory value is not taken into account yet?

So I've been looking at this deeper - seems like it is really only used for the npx init script...

"Save the given PUBLIC_DIR in package.json for future automatic updates of the worker script."

https://mswjs.io/docs/cli/init/#flags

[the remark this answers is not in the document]
jmcpeak: > Maybe you had some other plugin missing that led to the mock/* files not being included?

Nope, tried with vanilla knip install, these files are still listed as not used. I *did not* have vitest installed or configured in this vanilla install.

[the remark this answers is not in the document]
jmcpeak: > The patterns here don't have src/ included while the fixture does have that.

I didn't want to assume they create a src directory - i used /src in my example, guess I should have icluded them in my root dir as well.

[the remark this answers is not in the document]
webpro: > I can still put the check in, my guess is most people installing it don't use the `--save` flag.

This is the default, my guess is that MSW has a `postinstall` script that does this by default (also without `--save`).

> Nope, tried with vanilla knip install, these files are still listed as not used. 

I wonder what imports those files. Either there are source file(s) importing them, but somehow not part of the process, or they're imported by some external tool and we're missing a plugin. By now I'm inclined to think it's not MSW importing them automatically (except for the service worker).

jmcpeak:
Anything else you want me to modify? I have changed the code/tests to work properly now.

webpro:
The conversations/questions are closed/marked "resolved" while in fact they aren't. In general I think it's better to have the reviewer mark them as resolved.

jmcpeak:
> The conversations/questions are closed/marked "resolved" while in fact they aren't. In general I think it's better to have the reviewer mark them as resolved.

My bad, I thought they we an indicator to you that I finished the work and you should take a look.

packages/knip/src/plugins/msw/index.ts
jmcpeak:
```
const localConfig = basename(configFilePath) === 'package.json' ? manifest.msw : undefined;
const workerDirectory = localConfig?.workerDirectory;

if (workerDirectory) {
  ENTRY_FILE_PATTERNS.push(`${workerDirectory}/mockServiceWorker.{js,ts}`)
}
```

This bit of code looks for the entry somewhere else other than the default of public

packages/knip/test/plugins/msw.test.ts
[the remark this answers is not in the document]
jmcpeak: Yeah, I don't know what to do about this one.

jmcpeak:
> Let's wrap up this one. Here's how I think it should work:
> 
> * Default `entry` is (only) `['mockServiceWorker.js']`
> * In `findDependencies` we can look up `localConfig?.workerDirectory` and only if found, return the `toEntryPattern(join())` of that dir and `mockServiceWorker.js`
> * No other entry patterns should be there, I think also `mockServiceWorker.ts` is not picked up by MSW?
> * I think that's it, really :)
> 
> Let me know you'd like to finish this (happy to have you be the contributor!), otherwise I'm happy to pick it up.

I'd l;ove to take it over the finish line - I will implement the changes you requested today

packages/docs/scripts/generate-plugin-docs.ts
```diff
@@ -10,7 +10,7 @@ import remarkStringify from 'remark-stringify';
 import { unified } from 'unified';
 import { u } from 'unist-builder';
 import { base } from '../config.js';
-import type { Plugin } from '../../knip/src/types/plugins.js';
+import type { Plugin } from 'knip/dist/types/plugins.ts';
```
webpro, line 13: I think this is an unrelated change?

packages/knip/src/plugins/msw/index.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line n/no-restricted-import
+import path from 'node:path';
```
webpro, line 2: The restriction is there for a reason, please use the internal `join`.

packages/knip/test/plugins/msw.test.ts
```diff
@@ -14,15 +14,15 @@ test('Should not see the msw files in issues', async () => {
...
+  assert.equal(issues.files.size, 13);
   assert(issues.files.has(join(cwd, 'junk.js')));
   assert(issues.files.has(join(cwd, 'mocks/junk.js')));
   assert(issues.files.has(join(cwd, 'src/mocks/junk.js')));
 
   assert.deepEqual(counters, {
     ...baseCounters,
     devDependencies: 1,
-    files: 3,
+    files: 13,
```
webpro, line 25: Please remove unused files, we don't need to assert that functionality in this test.

webpro (approved):

package-lock.json
webpro, line 1: Please undo changes in the lockfile, I think they're unrelated?

jmcpeak: They are, just forgot about it

webpro: QA didn't pass because of this change: https://github.com/webpro/knip/pull/513#discussion_r1522804880

webpro:
:rocket: _This pull request is included in v5.1.0. See [Release 5.1.0](https://github.com/webpro/knip/releases/tag/5.1.0) for release notes._

webpro:
Thanks for bearing with me, Jason. Hope you don't mind I wrapped up the PR. It's out now! 🚀

jmcpeak:
> Thanks for bearing with me, Jason. Hope you don't mind I wrapped up the PR. It's out now! 🚀

Don't mind at all, glad that I could contribute!

PR #519, opened 2024-02-22
Fix knip --version command

sloanesturz:
Hi! This PR bumps the constant string in `version.ts` and updates the test implementation to ensure that the string doesn't get out of sync with `package.json` again.

I may or may not have gone on a goose chase through my NPM setup when `knip --version` was return 4.4.0 earlier today.

### DIFF
--- a/packages/knip/src/version.ts
+++ b/packages/knip/src/version.ts
@@ -1,1 +1,1 @@
-export const version = '4.4.0';
+export const version = '5.0.1';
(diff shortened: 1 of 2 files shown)


webpro:
Ooohhh noes 🤦‍♂️

There was a hiccup with that release indeed. Thanks!

webpro:
:rocket: _This pull request is included in v5.0.2. See [Release 5.0.2](https://github.com/webpro/knip/releases/tag/5.0.2) for release notes._

PR #532, opened 2024-03-02
Improve configuration of `include/exclude` types for better discoverability

romansp:
Currently `include` and `exclude` options in config are of type `string[]`. But these are already validated by `zod` to conform to `issueTypeSchema`, e.g. `files`, `dependencies` etc.

This PR improves types of `include` and `exclude` so that allowed values could be type-checked and also improves config discoverability without the need to refer to the docs.


webpro:
:rocket: _This pull request is included in v5.0.3. See [Release 5.0.3](https://github.com/webpro/knip/releases/tag/5.0.3) for release notes._

webpro:
Thanks Raman!

PR #537, opened 2024-03-10
fix: husky plugin compatibility

voxpelli:
`husky` v9 only required `knip` to stop follow the git hook path and always check in `.husky` – something that happens to work great for `husky` v8 as well.

The other logic in #500 causes issues rather than helps, as `gitHookPathsInV9` will be equal to `gitHooksPathInV8` in a traditional setup.

I simply reverted the plugin to the implementation prior to #500 and then added the `false` to `getGitHookPaths('.husky', false)` and that keeps all tests passing – both Husky v8 and Husky v9 – and made it find the dependencies used in my husky projects, both for v9 and v8 husky projects.

(Side note: Husky v9 says its backwards compatible with Husky v8 and that one shouldn't have to do anything when upgrading for it to work, this small change aligns with that)

Does this look okay to you @risu729?


risu729:
First of all,

risu729:
I think this PR should work at all and is a better solution for supporting v9.

I made it to follow the git config because the original implementation of the husky plugin followed it.

I just followed it since I was not sure about older versions of husky, so I had a concern about what if `core.hooksPath` was configurable in the past.

However, husky started using `core.hooksPath` in v5 ([source](https://blog.typicode.com/posts/husky-git-hooks-javascript-config/)), and seems it didn't have any configuration of the hooks directory name (which is `.husky/`).

So, I believe this PR doesn't cause any issues.

voxpelli:
@risu729 Makes sense 🙂

webpro:
Thanks @voxpelli! Sometimes less is literally more. Truly appreciate the efforts + team work here.

PR #548, opened 2024-03-14
Add documented type for second string arg to compilers

ryanwilsonperkin:
Fixes https://github.com/webpro/knip/issues/547

Compilers receive a filename string as their second argument, I believe this is the correct place in the zod schema to provide the definition so that they'll be available in typing.


webpro:
Yeah that should be enough, thanks a bunch @ryanwilsonperkin!

webpro:
:rocket: _This pull request is included in v5.1.1. See [Release 5.1.1](https://github.com/webpro/knip/releases/tag/5.1.1) for release notes._

PR #572, opened 2024-03-24
Add wrangler plugin

DaniFoldi:
Hi :wave:,

This PR aims to implement a plugin for Cloudflare Workers, or more specifically their `wrangler` cli. The config file is `wrangler.toml` with json config in experimental status (has been for quite a while), and contains one relevant entry, the main entrypoint of the worker.

I think I managed to add a correctly minimal fixture this time, the new config api made it even simpler compared to the Astro one :tada:, great work!

### DIFF
--- a/packages/knip/src/ConfigurationValidator.ts
+++ b/packages/knip/src/ConfigurationValidator.ts
@@ -77,6 +77,7 @@ export const pluginSchema = z.union([
 ]);
 
 const pluginsSchema = z.object({
+  wrangler: pluginSchema,
   astro: pluginSchema,
   angular: pluginSchema,
   ava: pluginSchema,
--- a/packages/knip/src/ConfigurationValidator.ts
+++ b/packages/knip/src/ConfigurationValidator.ts
@@ -77,7 +77,6 @@ export const pluginSchema = z.union([
 ]);
 
 const pluginsSchema = z.object({
-  wrangler: pluginSchema,
   astro: pluginSchema,
   angular: pluginSchema,
   ava: pluginSchema,
@@ -132,6 +131,7 @@ const pluginsSchema = z.object({
   vitest: pluginSchema,
   webpack: pluginSchema,
   wireit: pluginSchema,
+  wrangler: pluginSchema,
   yorkie: pluginSchema,
 });
 
--- a/packages/knip/src/plugins/index.ts
+++ b/packages/knip/src/plugins/index.ts
@@ -54,3 +54,4 @@ export { default as vue } from './vue/index.js';
 export { default as webpack } from './webpack/index.js';
 export { default as wireit } from './wireit/index.js';
 export { default as yorkie } from './yorkie/index.js';
+export { default as wrangler } from './wrangler/index.js';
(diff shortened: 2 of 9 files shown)


packages/knip/test/plugins/wrangler.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const cwd = resolve('fixtures/plugins/wrangler');
+
+test('Find dependencies with the wrangler plugin', async () => {
+  const { /* issues, */ counters } = await main({
+    ...baseArguments,
+    cwd,
+  });
+
+  // console.log(issues);
```
webpro, line 16: Please remove commented out code

packages/knip/src/plugins/index.ts
```diff
@@ -54,3 +54,4 @@ export { default as vue } from './vue/index.js';
 export { default as webpack } from './webpack/index.js';
 export { default as wireit } from './wireit/index.js';
 export { default as yorkie } from './yorkie/index.js';
+export { default as wrangler } from './wrangler/index.js';
```
webpro, line 57: Please move it up so it's ordered alphabetically

packages/knip/src/ConfigurationValidator.ts
```diff
@@ -77,6 +77,7 @@ export const pluginSchema = z.union([
 ]);
 
 const pluginsSchema = z.object({
+  wrangler: pluginSchema,
```
webpro, line 80: Please move it down so it's ordered alphabetically

webpro (approved):
Thanks! Seems like some smooth sailing, just a few minor notes and I'll be happy to merge it

webpro:
Please also make sure the tests pass

packages/knip/test/plugins/wrangler.test.ts
DaniFoldi: Fixed in the latest commit

packages/knip/src/plugins/index.ts
DaniFoldi: Fixed in the latest commit

packages/knip/src/ConfigurationValidator.ts
DaniFoldi: Fixed in the latest commit

DaniFoldi:
Hi,

Thanks for the quick review! I'm actually not sure why the test is failing, at first it didn't fail locally, but now it is so I'm not sure why that was. I'm not sure what the right way to fix the test is, actually, shouldn't `wrangler` being the enabler mark it as used? Or should I add a script to package.json that references it?

webpro:
> shouldn't `wrangler` being the enabler mark it as used? Or should I add a script to package.json that references it?

Usually it's best to have it close to a real-world scenario, so both sounds good :)

DaniFoldi:
Well I gave it a go 😅, now it's complaining about wrangler being unused and unlisted at the same time, I'm quite tired so I'll probably fix it tomorrow:
```
    +   binaries: 1,
    -   binaries: 0,
        classMembers: 0,
    +   dependencies: 1,
    -   dependencies: 0,
    ```

webpro:
Knip checks the `package.json#bin` field of the package in `node_modules`, so I just added that :)

webpro:
:rocket: _This pull request is included in v5.6.0. See [Release 5.6.0](https://github.com/webpro/knip/releases/tag/5.6.0) for release notes._

webpro:
Thanks again, Dániel!

DaniFoldi:
Makes sense, thank you!

PR #578, opened 2024-04-01
fix: add jsx extensions for next default file convention

Jaxenormus:
This is currently flagged as a false positive since any `default.tsx` or `default.jsx` file is not included in the entry point of the project but are also not imported since this is a special next.js file which marks it as unused by knip.

### DIFF
--- a/packages/knip/src/plugins/next/index.ts
+++ b/packages/knip/src/plugins/next/index.ts
@@ -14,8 +14,8 @@ const entry = ['next.config.{js,ts,cjs,mjs}'];
 const productionEntryFilePatternsWithoutSrc = [
   '{instrumentation,middleware}.{js,ts}',
   'app/global-error.{js,jsx,ts,tsx}',
-  'app/**/{error,layout,loading,not-found,page,template}.{js,jsx,ts,tsx}',
-  'app/**/{route,default}.{js,ts}',
+  'app/**/{error,layout,loading,not-found,page,template,default}.{js,jsx,ts,tsx}',
+  'app/**/{route}.{js,ts}',
   'app/{manifest,sitemap,robots}.{js,ts}',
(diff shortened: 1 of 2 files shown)


webpro:
Thanks for the PR! I'll look into it tomorrow, but iirc there was a similar discussion recently where the file turned out to be used by some plugin (not necessarily next/Next.js itself). Any chance you could confirm and/or point me to docs that do?

Jaxenormus:
> Thanks for the PR! I'll look into it tomorrow, but iirc there was a similar discussion recently where the file turned out to be used by some plugin (not necessarily next/Next.js itself). Any chance you could confirm and/or point me to docs that do?

I assume your referring to [this](https://github.com/webpro/knip/pull/441#issuecomment-1885186427) discussion only one i could find that was related to this on the repo, next.js doesn't explicitly state in the docs that the extension could also be `jsx` or `tsx` but in one of their [usage examples](https://nextjs.org/docs/app/building-your-application/routing/parallel-routes#modals) for modals they have you write a `default.tsx` file

CleanShot 2024-04-02 at 15 19 33@2x

webpro:
Thanks again for the PR and all the details, let's merge this.

webpro:
:rocket: _This pull request is included in v5.7.1. See [Release 5.7.1](https://github.com/webpro/knip/releases/tag/5.7.1) for release notes._

VladSez:
Sorry, not sure 100% if this is related but after updating to `knip@5.7.2` next route handler is flagged now as unused export.
I am using `knip --include-entry-exports`

On `knip@5.7.0` this is not reported and works as expected

image

webpro:
How about v5.7.3? 😜

VladSez:
It works! Thank you :)

PR #586, opened 2024-04-07
Add JUnit reporter

### LINKED ISSUE #568: Junit reporter
Could it be possible to include a JUnit reporter as part of Knip? For example, Bitbucket (and Jenkins, I believe, along with potentially other platforms) is capable of reading JUnit XML reports and displaying errors in a separate pipeline tab for end users. The reporter itself would generate an XML file containing all caught unused dependencies, unused exports, etc.

If that's acceptable to you, I can prepare a pull request for this feature/reporter. 😊

mrdannael:
Closes #568

PR aims to implement a reporter for generating JUnit xml. Without `--reporter-options` it will output the results to the console. By providing `path` value, the reporter will try to put the output into the file. 

Added small "fast-xml-parser" package to the knip package.

Also updated the docs with info about Junit reporter.


packages/knip/test/cli-reporter-junit.test.ts
```diff
@@ -0,0 +1,36 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites name="Knip report" tests="4" failures="4">
+  <testsuite name="Unused files" tests="1" failures="1">
+    <testcase tests="1" failures="1" name="Unused files" classname="src/unused.ts">
+      <failure message="Unused files" type="Unused files">Unused files: src/unused.ts</failure>
+    </testcase>
+  </testsuite>
+  <testsuite name="Unlisted dependencies" tests="2" failures="2">
+    <testcase tests="1" failures="1" name="Unlisted dependencies" classname="src/index.ts">
+      <failure message="Unlisted dependencies - unresolved" type="Unlisted dependencies">Unlisted dependencies: "unresolved" inside src/index.ts</failure>
```
webpro, line 21: I don't know anything about this format, but will each issue be in its own node, or should it be? This example gives me the impression that all unlisted dependencies will end up in this single node? Or will each one have the "Unlisted dependencies: " prefix?

Maybe you could show an example of how it looks like in the CI environment (eg. Bitbucket)?

webpro:
Is there still interest in wrapping up this pull request?

mrdannael:
Yeah, absolutely.
Thanks for the reminder.

That's how bitbucket diplays the test results file.
Screenshot from 2024-06-15 16-23-00

And partial source of that file.
[knip-test-example.txt](https://github.com/user-attachments/files/15848226/knip-test-example.txt)

Had to convert it from xml to txt to be able to attach it here.

mrdannael:
This is the actual output from junit reporter on Bitbucket pipelines run on the live project:

Screenshot from 2024-06-16 19-41-14

And this is part of the output xml file that was reported after running knip (in txt because Github won't allow to attach xml file). I've cut some sensitive data from the file (that's why issues count don't match with the image from BB).

[test-knip-report.txt](https://github.com/user-attachments/files/15859271/test-knip-report.txt)

webpro:
I feel like we could improve the output a bit.

For instance, I see this in one section:

> json.Unused devDependencies  package     ⋁
>
> 1 Unused devDependencies - postcss-scss
> 2
> 3 Unused devDependencies: "postcss-scss" inside package.json
> 4

Maybe all unused devDependencies of a file can go into a single foldout, without the duplication? Eg:

> Unused devDependencies in package.json     ⋁
>
> 1 postcss-scss
> 2 postcss-syntax

Doesn't need to be perfect, but I wonder how close we can get to something more like this?

mrdannael:
I need to test if BitBucket will accept something like that
```xml
<testsuite name="Unused dependencies" tests="1" failures="4">
    <testcase tests="1" failures="4" name="Unused dependencies" classname="package.json">
      <failure message="Unused dependencies - @types/google.accounts" type="Unused dependencies">Unused dependencies: "@types/google.accounts" inside package.json</failure>
      <failure message="Unused dependencies - fast-xml-parser" type="Unused dependencies">Unused dependencies: "fast-xml-parser" inside package.json</failure>
      <failure message="Unused dependencies - type-fest" type="Unused dependencies">Unused dependencies: "type-fest" inside package.json</failure>
      <failure message="Unused dependencies - web-vitals" type="Unused dependencies">Unused dependencies: "web-vitals" inside package.json</failure>
    </testcase>
  </testsuite>
```

or other similar output. If yes, then more compact output without duplications should be possible.

webpro:
Closing due to inactivity.

PR #590, opened 2024-04-09
fix(plugin/babel): bypass directly required plugins or presets in babel config

mridgway:
Babel supports importing or directly declaring plugins in the JS configuration. Currently `knip` will throw an error trying to parse the plugin name if a plugin/preset is directly imported.

Example config directly from babel monorepo: https://github.com/babel/babel/blob/main/babel.config.js#L199

I'm not sure if `knip` will actually handle the `require`/`import` in JS babel configs yet, but this at least makes it not throw.

### DIFF
--- a/packages/knip/fixtures/plugins/babel/.babelrc.js
+++ b/packages/knip/fixtures/plugins/babel/.babelrc.js
@@ -37,6 +37,8 @@ const config = {
         },
       },
     ],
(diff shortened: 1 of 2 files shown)


webpro:
:rocket: _This pull request is included in v5.9.2. See [Release 5.9.2](https://github.com/webpro/knip/releases/tag/5.9.2) for release notes._

PR #596, opened 2024-04-19
Escape all regex special characters in identifiers

### LINKED ISSUE #595: Arguably terrible enum identifiers causes breakage.
So I was trying to run `knip` on a larger project and was runing into errors that was not very clear to me at first. 

```
file:///Z:/My_Large_Repo/node_modules/knip/dist/typescript/getImportsAndExports.js:298
        for (const match of sourceFile.text.matchAll(new RegExp(item.identifier.replace(/\$/g, '\\$'), 'g'))) {
                                                     ^

SyntaxError: Invalid regular expression: /+/g: Nothing to repeat
    at new RegExp (<anonymous>)
    at setRefs (file:///Z:/My_Large_Repo/node_modules/knip/dist/typescript/getImportsAndExports.js:298:54)     
    at getImportsAndExports (file:///Z:/My_Large_Repo/node_modules/knip/dist/typescript/getImportsAndExports.js:314:13)
    at ProjectPrincipal.analyzeSourceFile (file:///Z:/My_Large_Repo/node_modules/knip/dist/ProjectPrincipal.js:140:47)
    at analyzeSourceFile (file:///Z:/My_Large_Repo/node_modules/knip/dist/index.js:187:66)
    at main (file:///Z:/My_Large_Repo/node_modules/knip/dist/index.js:268:17)
    at async run (file:///Z:/My_Large_Repo/node_modules/knip/dist/cli.js:25:73)
    at async file:///Z:/My_Large_Repo/node_modules/knip/dist/cli.js:85:1
```

After adding some logging into the `getImportsAndExports.js` I managed to find that it was choking parsing an enum we have to handle potential values of `KeyboardEvent.key`, including keys like `*` and `+`, which is arguably a terrible idea, but is allowed in Typescript, so I guess knip shouldn't fail on it either.

Here's a repro: https://stackblitz.com/edit/github-4tmeki?file=enum.ts%3AL21

I think it should be pretty easy to amend, I might send a PR too

oBusk:
Rather than just escaping `$`, escape all regex special characters in any identifiers to make sure the RegExp that is created is valid. 

Used the list of characters to escape from the regex-escaping proposal 

https://github.com/tc39/proposal-regex-escaping/blob/66d654ea6561ea064045c7ecf2da6870c892c9be/polyfill.js#L4

Closes #595

### DIFF
--- a/packages/knip/src/typescript/getImportsAndExports.ts
+++ b/packages/knip/src/typescript/getImportsAndExports.ts
@@ -13,6 +13,7 @@ import type {
   UnresolvedImport,
 } from '../types/serializable-map.js';
 import { timerify } from '../util/Performance.js';
+import { escapeRegex } from '../util/regex.js';
 import { isStartsLikePackageName, sanitizeSpecifier } from '../util/modules.js';
 import { extname, isInNodeModules } from '../util/path.js';
 import { shouldIgnore } from '../util/tag.js';
@@ -400,7 +401,7 @@ const getImportsAndExports = (
   const setRefs = (item: SerializableExport | SerializableExportMember) => {
     if (!item.symbol) return;
     const symbols = new Set<ts.Symbol>();
-    for (const match of sourceFile.text.matchAll(new RegExp(item.identifier.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&'), 'g'))) {
+    for (const match of sourceFile.text.matchAll(new RegExp(escapeRegex(item.identifier), 'g'))) {
       const isDeclaration = match.index === item.pos || match.index === item.pos + 1; // off-by-one from `stripQuotes`
       if (!isDeclaration) {
         // @ts-expect-error ts.getTokenAtPosition is internal fn
--- a/packages/knip/src/typescript/getImportsAndExports.ts
+++ b/packages/knip/src/typescript/getImportsAndExports.ts
@@ -400,7 +400,7 @@ const getImportsAndExports = (
   const setRefs = (item: SerializableExport | SerializableExportMember) => {
     if (!item.symbol) return;
     const symbols = new Set<ts.Symbol>();
-    for (const match of sourceFile.text.matchAll(new RegExp(item.identifier.replace(/\$/g, '\\$'), 'g'))) {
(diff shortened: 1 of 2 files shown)


packages/knip/src/typescript/getImportsAndExports.ts
```diff
@@ -387,7 +387,7 @@ const getImportsAndExports = (
 
   const setRefs = (item: SerializableExport | SerializableExportMember) => {
     if (!item.symbol) return;
-    for (const match of sourceFile.text.matchAll(new RegExp(item.identifier.replace(/\$/g, '\\$'), 'g'))) {
+    for (const match of sourceFile.text.matchAll(new RegExp(item.identifier.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&'), 'g'))) {
```
webpro, line 390: This targets (enum) members, perhaps it makes sense to only apply this to members at https://github.com/webpro/knip/blob/main/packages/knip/src/typescript/getImportsAndExports.ts#L43? Ideally we'd create the regex once and reuse it (I should've done that myself here too).

webpro (commented):
Thanks! Happy to merge, but I do have a question.

packages/knip/src/typescript/getImportsAndExports.ts
oBusk: @webpro do you mean to make the escaping of special characters already when creating the serializableMember? I've just had a cursory look, and it looks like modifying the `identifier` would have lot's of other effects, but perhaps adding another `regex` property to the serializable members which does identifier->escape->regex and then we can use that pre-created regex here for the `matchAll()`?

oBusk: Yeah ok I had another look, and I can't figure out what exactly you mean. I've moved the Regex into it's own file to be reused, also tried to search for anywhere in the code here a similiar regex was used, but couldn't find anywhere. 
Let me know what you think!

webpro:
@oBusk Thanks for pushing this, @oBusk! When looking into it myself I figured it could be faster and simpler by doing a literal text-based search without regexes at all. So I implemented that including fixtures and test to prevent regressions.

oBusk:
> @oBusk Thanks for pushing this, @oBusk! When looking into it myself I figured it could be faster and simpler by doing a literal text-based search without regexes at all. So I implemented that including fixtures and test to prevent regressions.

Oh crap, of course! That's a lot better! Nice!

PR #606, opened 2024-04-27
fix: next route handler extensions

Jaxenormus:
This is unfortunately undocumented like most next.js quirks but this does work if you are making an og image route handler kinda like the `opengraph-image` convention but with less magic

### DIFF
--- a/packages/knip/src/plugins/next/index.ts
+++ b/packages/knip/src/plugins/next/index.ts
@@ -15,7 +15,7 @@ const productionEntryFilePatternsWithoutSrc = [
   '{instrumentation,middleware}.{js,ts}',
   'app/global-error.{js,jsx,ts,tsx}',
   'app/**/{error,layout,loading,not-found,page,template,default}.{js,jsx,ts,tsx}',
-  'app/**/route.{js,ts}',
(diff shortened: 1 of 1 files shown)


webpro:
Thanks for the PR, @Jaxenormus. Any chance you could share what routes exactly aren't matching?

There's https://knip.dev/reference/plugins/vercel-og but that's not enough I guess. What related dependencies are listed in `package.json`?

Jaxenormus:
> Thanks for the PR, @Jaxenormus. Any chance you could share what routes exactly aren't matching?
> 
> There's [knip.dev/reference/plugins/vercel-og](https://knip.dev/reference/plugins/vercel-og) but that's not enough I guess. What related dependencies are listed in `package.json`?

in our use case the route that did not match was `apps/hub/app/api/images/ [companyRoute]/og-png/route.tsx` where the route was sending an `ImageResponse` which lets you return jsx but to do so the files has to be a tsx file

webpro:
Alright, thanks for the clarification!

webpro:
:rocket: _This pull request is included in v5.12.0. See [Release 5.12.0](https://github.com/webpro/knip/releases/tag/5.12.0) for release notes._

PR #621, opened 2024-05-03
feat(plugins): add plugin for xo config

bbeesley:
Adds a plugin for [xo](https://github.com/xojs/xo#config). Essentially its an opinionated wrapper around eslint with a bunch of bundled rules and plugins, and the config object is just the eslint config object with some extra options. So to keep it consistent with the eslint plugin and avoid duplication I've just used the eslint helper to parse out any additional eslint plugins that have been added in the xo config.

### DIFF
--- a/packages/knip/src/ConfigurationValidator.ts
+++ b/packages/knip/src/ConfigurationValidator.ts
@@ -135,6 +135,7 @@ const pluginsSchema = z.object({
   webpack: pluginSchema,
   wireit: pluginSchema,
   wrangler: pluginSchema,
+  xo: pluginSchema,
   yorkie: pluginSchema,
 });
 
--- a/packages/knip/src/plugins/xo/types.ts
+++ b/packages/knip/src/plugins/xo/types.ts
@@ -1,0 +1,13 @@
+import type { ESLintConfig } from '../eslint/types.js';
+
+export type XOConfig = ESLintConfig & {
+  envs?: string[] | undefined;
+  globals?: string[] | undefined;
+  ignores?: string[] | undefined;
+  nodeVersion?: string | boolean | undefined;
+  prettier?: boolean | undefined;
+  printConfig?: string | undefined;
+  semicolon?: boolean | undefined;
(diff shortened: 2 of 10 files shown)


packages/knip/src/ConfigurationValidator.ts
```diff
@@ -77,6 +77,7 @@ export const pluginSchema = z.union([
 ]);
 
 const pluginsSchema = z.object({
+  xo: pluginSchema,
```
uncenter, line 80: Pretty sure this object should be in alphabetical order?

bbeesley: it's generated code, and biome didn't complain about it, but you're absolutely right, it doesn't look right there.

webpro:
Awesome! This is great, the only remark I'd have is that there could be a `package.json#scripts` using `xo` to make it even more complete/realistic, eg:

```json
{
  "scripts": {
    "test": "xo && ava"
  }
}
```

Happy to merge this!

packages/knip/src/plugins/xo/types.ts
```diff
@@ -0,0 +1,15 @@
+import type { ESLintConfig } from '../eslint/types.js';
+
+export type XOConfig = ESLintConfig & {
+  envs?: string[] | undefined;
+  globals?: string[] | undefined;
+  parser?: string | undefined;
+  plugins?: string[];
```
webpro, line 7: Not sure what's already in `ESLintConfig` or should be over there?

bbeesley: I've deduped them, thx

webpro:
Sorry I forgot to tell more about how binaries work, I've just added it to prevent some ping-pong. Hope you don't mind :)

webpro:
:rocket: _This pull request is included in v5.13.0. See [Release 5.13.0](https://github.com/webpro/knip/releases/tag/5.13.0) for release notes._

webpro:
Thanks again @bbeesley!

PR #632, opened 2024-05-12
fix(plugin-commitlint): handling of `parserPreset` when it's not a string

### LINKED ISSUE #631: commitlint plugin update breaks when `parserPreset` isn't a `string`
Bug was introduced in 5.14.0

`parserPreset` can be a record or a promise that returns a record.

RebeccaStevens:
fix #631

### DIFF
--- a/packages/knip/fixtures/plugins/commitlint/.commitlintrc.json
+++ b/packages/knip/fixtures/plugins/commitlint/.commitlintrc.json
@@ -1,4 +1,5 @@
 {
   "extends": ["@commitlint/config-conventional"],
-  "plugins": ["commitlint-plugin-tense"]
+  "plugins": ["commitlint-plugin-tense"],
+  "parserPreset": "conventional-changelog-atom"
 }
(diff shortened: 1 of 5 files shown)


RebeccaStevens:
@webpro Do you want me to include a test with this?
If so, should I just modify [`packages/knip/fixtures/plugins/commitlint/commitlint.config.js`](https://github.com/webpro/knip/blob/main/packages/knip/fixtures/plugins/commitlint/commitlint.config.js) but changing out the `parserPreset` value?

webpro:
Thanks for the PR!

> @webpro Do you want me to include a test with this? If so, should I just modify [`packages/knip/fixtures/plugins/commitlint/commitlint.config.js`](https://github.com/webpro/knip/blob/main/packages/knip/fixtures/plugins/commitlint/commitlint.config.js) but changing out the `parserPreset` value?

Yes, that would be great. And then maybe add that string value to the other config file so we still have both types covered.

webpro:
:rocket: _This pull request is included in v5.16.0. See [Release 5.16.0](https://github.com/webpro/knip/releases/tag/5.16.0) for release notes._

PR #638, opened 2024-05-14
Add UnoCSS plugin

hyoban:

### DIFF
--- a/packages/knip/fixtures/plugins/unocss/unocss.config.ts
+++ b/packages/knip/fixtures/plugins/unocss/unocss.config.ts
@@ -1,0 +1,6 @@
+// uno.config.ts
+import { defineConfig } from 'unocss'
+
+export default defineConfig({
+  // ...UnoCSS options
+})
--- a/packages/knip/fixtures/plugins/unocss/unocss.config.ts
+++ b/packages/knip/fixtures/plugins/unocss/unocss.config.ts
@@ -1,6 +1,30 @@
 // uno.config.ts
-import { defineConfig } from 'unocss'
+import { defineConfig, presetAttributify, presetIcons, presetUno, presetWebFonts } from 'unocss';
 
 export default defineConfig({
-  // ...UnoCSS options
-})
+  shortcuts: [
+    [
+      'btn',
+      'px-4 py-1 rounded inline-block bg-teal-600 text-white cursor-pointer hover:bg-teal-700 disabled:cursor-default disabled:bg-gray-600 disabled:opacity-50',
+    ],
+    [
+      'icon-btn',
+      'text-[0.9em] inline-block cursor-pointer select-none opacity-75 transition duration-200 ease-in-out hover:opacity-100 hover:text-teal-600 !outline-none',
+    ],
+  ],
+  presets: [
+    presetUno(),
+    presetAttributify(),
+    presetIcons({
+      scale: 1.2,
+      warn: true,
+    }),
+    presetWebFonts({
+      fonts: {
+        sans: 'DM Sans',
+        serif: 'DM Serif Display',
+        mono: 'DM Mono',
+      },
+    }),
+  ],
+});
--- a/packages/knip/src/plugins/unocss/index.ts
+++ b/packages/knip/src/plugins/unocss/index.ts
@@ -1,0 +1,33 @@
+import type { EnablerPatterns } from '#p/types/config.js';
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '#p/types/plugins.js';
+import { hasDependency } from '#p/util/plugin.js';
+import type { PluginConfig } from './types.js';
+
+// https://unocss.dev/guide/config-file
+
+const title = 'UnoCSS';
+
+const enablers: EnablerPatterns = ['unocss', /^@unocss\//];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['uno.config.{js,ts,mjs,mts}', 'unocss.config.{js,ts,mjs,mts}'];
+
+const entry: string[] = [];
+
+const production: string[] = [];
+
+const resolveConfig: ResolveConfig<PluginConfig> = async config => {
+  const dependencies = config?.plugins ?? [];
+  return [...dependencies];
+};
+
+export default {
+  title,
+  enablers,
+  isEnabled,
+  config,
+  entry,
+  production,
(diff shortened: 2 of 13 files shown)


webpro (commented):
Thanks @hyoban! No modifications seem to have been made, any chance you could still customize or just remove things specifically tailored to UnoCSS?

hyoban:
Sorry. I don't quite understand what I need to modify, can you be more specific? I basically implemented the unocss plug-in according to the vite plug-in.

packages/knip/src/plugins/unocss/types.ts
```diff
@@ -0,0 +1,4 @@
+export type PluginConfig = {
+  plugins?: string[];
```
webpro, line 2: This type doesn't seem correct?

packages/knip/src/plugins/unocss/types.ts
```diff
@@ -0,0 +1,4 @@
...
+  entryPathsOrPatterns?: string[];
```
webpro, line 3: I think this type does not exist in UnoCSS config

packages/knip/src/plugins/unocss/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const enablers: EnablerPatterns = ['unocss', /^@unocss\//];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['uno.config.{js,ts,mjs,mts}', 'unocss.config.{js,ts,mjs,mts}'];
+
+const entry: string[] = [];
+
+const production: string[] = [];
```
webpro, line 18: Please remove empty arrays

packages/knip/src/plugins/unocss/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+// https://unocss.dev/guide/config-file
+
+const title = 'UnoCSS';
+
+const enablers: EnablerPatterns = ['unocss', /^@unocss\//];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['uno.config.{js,ts,mjs,mts}', 'unocss.config.{js,ts,mjs,mts}'];
```
webpro, line 14: Please format biome, e.g. using `bun format`

packages/knip/src/plugins/unocss/index.ts
```diff
@@ -0,0 +1,33 @@
+import type { EnablerPatterns } from '#p/types/config.js';
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '#p/types/plugins.js';
+import { hasDependency } from '#p/util/plugin.js';
+import type { PluginConfig } from './types.js';
+
+// https://unocss.dev/guide/config-file
+
+const title = 'UnoCSS';
+
+const enablers: EnablerPatterns = ['unocss', /^@unocss\//];
```
webpro, line 10: Let's use only `unocss` and `@unocss/cli` for better precision?

packages/knip/fixtures/plugins/unocss/unocss.config.ts
```diff
@@ -0,0 +1,6 @@
+// uno.config.ts
+import { defineConfig } from 'unocss'
+
+export default defineConfig({
+  // ...UnoCSS options
```
webpro, line 5: I think it would be good to use a more comprehensive config file here to make it more real-world.

packages/knip/src/plugins/unocss/index.ts
hyoban: There are other dependencies to consider, such as `@unocss/webpack`, `@unocss/postcss`

hyoban:
Thank you for your review. Sorry, I thought the default code generated from CLI could be ignored at first.

Another question, can the plug-in tell knip to ignore some dependencies that it thinks do not exist when using UnoCSS? For example `uno.css`, `virtual:uno.css`, `virtual:unocss-devtools`

packages/knip/src/plugins/unocss/index.ts
webpro: The "enabling packages" should only be root/core modules. The rest is referenced elsewhere, in the tool configuration. The plugin returns those to Knip, and packages can then be reported as unlisted or unused.

Having more enablers is not wrong, I just think a smaller list is more precise/easier to understand for consumers. Otherwise we'd have to try and list all possible plugins/dependencies etc. in all plugins which doesn't really make sense imho.

webpro:
> Thank you for your review. Sorry, I thought the default code generated from CLI could be left alone.

No worries! Thanks for contributing to Knip, I really appreciate it.

> Another question, can the plug-in tell Knip to ignore some dependencies that it thinks do not exist when using UnoCSS? For example `uno.css`, `virtual:uno.css`, `virtual:unocss-devtools`

No, but specifiers starting with `virtual:` are ignored by default.

If `uno.css` as not referenced anywhere (not even it's executable/binary), then we could just return it from the `resolveConfig` function. It's a bit rare, but it happens. Here's an example: https://github.com/webpro/knip/blob/main/packages/knip/src/plugins/yorkie/index.ts#L26

packages/knip/src/plugins/unocss/index.ts
hyoban: My previous understanding was wrong, it seems we only need `['unocss']`

hyoban:
> No, but specifiers starting with `virtual:` are ignored by default.

And we need to set `virtual:uno.css` in `ignoreDependencies`?

webpro:
> > No, but specifiers starting with `virtual:` are ignored by default.

Sorry, I was wrong about this, they're not ignored. Here's what happened: https://github.com/webpro/knip/issues/370

> And we need to set `virtual:uno.css` in `ignoreDependencies`?

Yes, that's currently how it works indeed.

> I added a test that will fail in `packages/knip/fixtures/plugins/unocss/main.ts`

There's two options (I don't mind which one):

1. Accept there's an unlisted dependency (`assert(issues.unlisted['main.ts']['virtual:uno.css']);`)
2. Add a Knip config to the fixtures folder with `ignoreDependencies`

webpro:
Thanks @hyoban!

webpro:
:rocket: _This pull request is included in v5.17.0-canary.0. See [Release 5.17.0-canary.0](https://github.com/webpro/knip/releases/tag/5.17.0-canary.0) for release notes._

PR #660, opened 2024-05-28
Add ‘kill’ and ‘ssh’ as globally available binaries

KittyGiraudel:
Hello! 👋  It seems we list a lot of unix commands as available binaries already so I thought `kill` and `ssh` would be fair game. Let me know if not.

### DIFF
--- a/packages/knip/src/constants.ts
+++ b/packages/knip/src/constants.ts
@@ -46,6 +46,7 @@ export const IGNORED_GLOBAL_BINARIES = new Set([
   'git',
   'grep',
   'gzip',
+  'kill',
   'ln',
   'ls',
   'mkdir',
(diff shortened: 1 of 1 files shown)


KittyGiraudel:
Would you accept a PR to add `aws` to this list? :)

webpro:
Drifting off a tiny bit, but I'd accept a PR to add it!

Btw, you can do e.g. `ignoreBinaries: ["aws"]`

KittyGiraudel:
@webpro That’s what we currently do, but I reckon `aws` is popular enough that it would make sense? I understand it’s debatable though.

webpro:
Agreed, let's do it. There's things like `xcodebuild` and `gem` too. I'll think about boundaries better if/when it gets out of hand.

PR #664, opened 2024-05-30
fix: 🐛 Fix regular expression escape error

mahoushoujoarale:
fix: #595


webpro:
Thanks for pushing this, @mahoushoujoarale! When looking into it myself I figured it could be faster and simpler by doing a literal text-based search without regexes at all. So I implemented that including fixtures and test to prevent regressions.

Issue #668, opened 2024-06-03
`ignoreBinaries` config not respected anymore


geigerzaehler:
Starting with version 5.17.0 the `ignoreBinaries` configuration option does not work anymore. I have the following files:

```json
{
  "type": "module",
  "packageManager": "pnpm@9.1.2+sha256.19c17528f9ca20bd442e4ca42f00f1b9808a9cb419383cd04ba32ef19322aba7",
  "dependencies": {
    "@biomejs/biome": "^1.7.3",
    "knip": "5.17.3"
  }
}
```

```js
// knip.js
export default {
  ignoreBinaries: ["biome"],
};
```

When I run `knip` an issue is reported.
```
Unused dependencies (1)
@biomejs/biome  package.json
Configuration issues (1)
Unused item in ignoreBinaries: biome
```

When I use v5.16.0 no issue is reported.

webpro:
Behavior has changed indeed.

From what I can tell here, you can add `@biomejs/biome` to `ignoreDependencies` and remove `biome` from `ignoreBinaries`.

If that doesn't help I'd need more details, ideally a minimal reproduction of the issue.

geigerzaehler:
Yes, I was able to work around the problem using `ignoreDependencies`. Personally, I find the `ignoreBinaries` option a lot clearer, especially in cases where you the package name does not match the binary name. But I can understand if you think the workaround is sufficient.

Regarding the minimal reproduction: The two files I posted above reproduce the issue. If you want me to create a StackBlitz, I can also do that.

webpro:
I'd say it's not a workaround, yet behaving more strict now. The _dependency_ listed in `package.json` is the thing that needs to be ignored, right?

The other way around works too: if you would use the `biome` executable somewhere without having the `@biomejs/biome` listed (e.g. if installed globally) then you can use `ignoreBinaries`.

geigerzaehler:
Ah, I see. That makes sense. I think I got confused by the documentation for `ignoreBinaries` and the fact that adding it to `ignoreBinaries` fixed my issue. Would you be open to a PR to the [docs](https://knip.dev/reference/configuration#ignorebinaries)? Something along the lines of 

> ignoreBinaries
> Exclude binaries that are used but unlisted from the report. Value is an array of binary names or regular expressions.

webpro:
I'm definitely open to PRs. Although specifically "but unlisted" might be confusing since one doesn't list binaries (but the package they're part of instead).

[...]

PR #670, opened 2024-06-03
Better explain `ignoreBinaries` configuration option

geigerzaehler:
As mentioned in #668, the documentation for the `ignoreBinaries` confused me and lead me to use it for an unintended side effect. I hope the change makes it clear what the configuration option is for.


webpro:
Thanks! I like it.

PR #673, opened 2024-06-06
Make TS-style path mappings work for all files with extensions

### LINKED ISSUE #669: Path mapping configuration not respected for .json files
The `path` configuration works for `.js` files but not for `.json` files.

I have the following config:
```json
{
  "paths": {
    "@src/*": ["./src/*"]
  }
}
```

and the following entry point:
```js
import "@src/data.json"
```

Running knip reports an issue
```
Unlisted dependencies (1)
@src/data.json  src/index.js
```

Before v5.17 I was able to work around the problem by specifying `"ignoreDependencies": ["@src"]` but that doesn’t work anymore.

If I use `import "@src/data.js"` or `import "./data.json"` no issues are reported.

You can find a minimal production on [StackBlitz](https://stackblitz.com/edit/github-pkwe7r?file=src%2Findex.js)

geigerzaehler:
When path mappings are specified the mappings are applied to all paths regardless of their extension. We use [tsconfig-paths](https://github.com/dividab/tsconfig-paths) which implements TypeScript’s path mapping algorithm but also works for extensions other than `.js`, `.ts`, etc.

Fixes #669


webpro:
Thanks for the pull request!

The idea is that if `resolveSync` does not resolve to something, we'll fall back to TypeScript. And as you point out already, the TS module resolver doesn't resolve files with non-standard extensions.

That's where the `virtualFileExtensions` come in. They are a merger of the following types of files:

- `FOREIGN_FILE_EXTENSIONS`
- File extensions added as custom compilers (e.g. Vue, Svelte, Astro, MDX files)

Before adding the dependency you're proposing, I'd like to explore those options first:

- We can add `.json` to the list of foreign file extensions. This means the file will basically become a dummy file with an empty response, but also it will not be marked as unused.
- We could add a custom compiler for `.json` files and either return nothing (or the actual parsed object, but this isn't necessary as the files aren't executed and JSON won't have imports or references to other exports).

Haven't really tried either of the options, but I'm leaning towards the former, being the simplest/fastest one. What do you think?

geigerzaehler:
I get your point about not adding another dependency. (Especially if the dependency does something similar to already existing code.) I ran into the issue because I was importing `.json` files and `.md` files with a custom loader. Adding those extensions to `FOREIGN_FILE_EXTENSIONS` fixes my issue.

But the idea for this PR was to make path mapping work for _all_ extensions without listing them explicitly and for files without extensions. While playing around with this, I also noted that the same limitation applies to subpath imports in `package.json`. So I think it would be worthwhile to implement proper support for this.

If you think this is a good idea, I’d be happy to work on it and would appreciate some guidance. (My current approach does not really work. Maybe we can also make the TS compiler handle this properly.) If not, I’ll create a PR that just adds `.json` and `.md` to the list of extensions.

webpro:
Thanks for the reply. Would for sure be nice to have path mappings working for any extension. And there have been reports for e.g. extensionless imports of e.g. Vue files with just `./Component`. This got me thinking, so here it goes..

Maybe it's a bit much to digest, but I'm writing it down anyway. What's currently in Knip's module resolver:

1. `resolveSync` as the first resolver that covers majority of cases.
2. `toSourceFilePath` to handle the cases where `dist/index.js` should instead be resolved to `src/index.ts`
3. `ts.resolveModuleName` is the second, default TS resolver that takes into account `compilerOptions.paths` (but only for accepted/standard extensions)
4. Similar to step 2, when we find a DTS file, we return the related source file instead.
5. `ts.resolveModuleName` with custom `ts.sys` is the third resolver that resolves "virtual" extensions for files that require compilation, and appends the `.js` extension to trick TS into thinking it's a valid file (in `SourceFileManager` we actually do the compilation).
6. If the previous steps fail to resolve to a file, `fileExists` is an extra intermediate step to resolve any file (with an non-standard or non-virtual extension), with a simple `path.join()`, so this only supports relative or absolute import specifiers.
7. If the previous step also fails the last thing to do is return the virtual file path.

I'm not sure what exactly `tsconfig-paths` supports and how feasible my initial idea (below) would be, but perhaps it's interesting to explore replacing all three resolvers with just one (four if we'd consider step 6 as a resolver). The drawback is that the initial `resolveSync` might be faster simply because it's more limited in scope (no path mappings and less extensions), but the advantage is that everything would be resolved in a single pass. And then we can also look into more aggressive caching of successful resolutions (if `tsconfig-paths` doesn't do that already).

So my initial idea would be roughly this:

1. Let `tsconfig-paths` resolve whatever it can with `DEFAULT_EXTENSIONS` + `virtualFileExtensions` + `FOREIGN_FILE_EXTENSIONS`
2. Make sure the `toSourceFilePath` idea is applied to "dist" files and DTS files
3. Return `extension: ts.Extension.Js` (only) for virtual extensions.

Note that everything within `resolveModuleNames` must be sync (limitation of TS).

As far as the implementation goes, I'd say passing the unit tests is a very good start. There's also [integration tests](https://github.com/webpro-nl/knip/blob/main/.github/workflows/integration.yml).

Depending on the number of extensions (Knip) and usage of path mappings (user), the single-pass resolution might be "expensive", but there's only one way to find out.

The module resolver function isn't _timerified_ yet, so the next release (today or tomorrow) will include that so we can do `knip --performance` and the `resolveModuleNames` numbers in the report can be used to compare speed of implementations. We'll see about trade-offs (a better resolver and a simpler implementation vs slight performance decrease?).

WDYT? I understand this might all be a bit overwhelming, so feel free to ask away, break it down into smaller steps (e.g. first replace only `resolveSync` with `tsconfig-paths`) and we can discuss further along the way, and/or simply turn it down.

webpro:
> Thanks for the detailed explanation. It does indeed seem that there are opportunities to consolidate the resolution algorithm and I would like to give it a shot.

Great! Would be super useful for Knip to improve here.

> `tsconfig-paths` only finds mapped paths. So in your outline we would need an additional step after 1. where we resolve relative paths and packages. But that shouldn’t be a big problem.

Oh that's a bummer, but let's see where this goes.

> I’m not clear yet on the purpose of the different types of extension (default, virtual, foreign).

- [default extensions](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/constants.ts#L21) are the ones TypeScript accepts to be added to the program (which it can parse, and compile)
- virtual extensions have the `.js` extension (`ts.Extension.Js`) appended so they can be added to the TS program (and Knip "compiles" them to JS in `SourceFileManager` as requested by the TS program); Knip includes [a few default compilers](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/compilers/index.ts#L35-L40), but more can be added and the existing ones can be overridden, all of those file extensions fall in this category. Because of the `.js` append trick, some handling of such files, hence the custom sys stuff.
- [foreign extensions](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/constants.ts#L80-L102) are for files that don't need to be compiled and don't import anything so can effectively be ignored cq provided to the TS program as an empty string (note: if the `.css` compiler is added by the user config then that should have precedence)

> In any case, I don’t think path mapping should try to append any but the default extensions. I.e. it should resolve `@src/data.ext` to `./src/data.ext` but it shouldn’t resolve `@src/data` to `./src/data.ext` (except where `.ext` is one of `.js`, `.ts` etc.).

Yes, agreed. Also note that e.g. `@org/package.css` is a valid package name.

> This is why I’m not too concerned about performance. The slow part will always be asking the file system whether a path exists. `tsconfig-paths` generates a list of mapped candidates and then tries them in turn and picks the first one. But I will keep an eye on it using `--performance`

Yeah, the most expensive ones are the ones not found.

I've added `resolveModuleNames` to `--performance` in `main`.

> I’m still a bit unclear why there are TypeScript `System` instances

Also see above re. virtual files. The default `sys` wouldn't handle virtual files.

I've tried to optimize by not adding compilers when they're not necessary (e.g. if `svelte` is not a dependency we don't need that compiler and we don't need to look for that extension). However, there's also `FOREIGN_FILE_EXTENSIONS` so still quite the list.

> and how `resolveModuleName` is used in the compiler to support analyzing references and if this creates some unforeseen issues. But we’ll see and I’ll try to figure this out.

When Knip creates a program (`ts.createProgram()`) with the entry files, TS reads and parses them, import specifiers go into ([resolveModuleNames](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/typescript/createHosts.ts#L54)) so TS knows which files to add to the program as well.

This could be compared to what LSP provides to plugins like for Astro or Vue. Such IDE extensions handle compilation and module resolution directly in the editor. Knip does those things as well, but in a different setting (Node.js/CLI), and more universal (for any type of file).

> The next step for me would be to experiment with this a bit to see how we can accomplish this. I’ll try to break it up as much as possible and keep changes small.

Hack around 🔥 Feel free to hit me up such as with DM-ing me on Discord if you prefer.

geigerzaehler:
So, I’ve implemented a basic version that uses `tsconfig-paths` to remap paths with any extension. I kept the change small and simple and avoided touching any other part of the code. This should give us a baseline for refactoring the module resolution and minimizes the performance impact.

My plan is to try to move the `tsconfig-paths` resolver from the bottom to the top of the resolver list and eliminate all the resolvers that we don’t need anymore in a follow-up PR.

webpro:
> So, I’ve implemented a basic version that uses `tsconfig-paths` to remap paths with any extension. I kept the change small and simple and avoided touching any other part of the code. This should give us a baseline for refactoring the module resolution and minimizes the performance impact.

Thanks! Definitely a step in the right direction.

> My plan is to try to move the `tsconfig-paths` resolver from the bottom to the top of the resolver list and eliminate all the resolvers that we don’t need anymore in a follow-up PR.

Yeah, appreciate it is now basically a light-weight "fallback resolver" indeed it doesn't hurt much in terms of performance.

webpro:
:rocket: _This pull request is included in v5.20.0. See [Release 5.20.0](https://github.com/webpro-nl/knip/releases/tag/5.20.0) for release notes._

PR #680, opened 2024-06-14
Eliminiate custom TS System instance

geigerzaehler:
Follow-up to #673

The custom instance was previously responsible for resolving mapped paths to files handled by an external loader (e.g. `.vue` files) or treated as-is (e.g. `.png`). Now, `tsconfig-paths` does the path mapping resolution.

Note that the custom system resolver was really only hit for mapped paths. For non-mapped paths `resolveSync()` would already resolve the file.

That’s my best guess explanation why we can make this change. Please let me know if I overlooked something.

It also looks like this does make resolution faster. I ran it multiple times against the `TanStack/query` repo (because it has the most time spent in `resolveModuleName()`) and I’m seeing a 5-20% decrease in time spent in `resolveModuleName()`. (Take this with a grain of salt, though, since it’s only a rudimentary evaluation.)


geigerzaehler:
Looks like the windows tests fail. I’ll take a look and hopefully it doesn’t invalidate the whole approach

packages/knip/src/typescript/resolveModuleNames.ts
```diff
@@ -73,11 +69,15 @@ export function createCustomModuleResolver(
     // No need to try and resolve builtins or externals, bail out
     if (isBuiltin(sanitizedSpecifier) || isInNodeModules(name)) return undefined;
 
-    const resolvedFileName = resolveSync(sanitizedSpecifier, containingFile, extensions);
+    const resolvedFileName = resolveSync(sanitizedSpecifier, containingFile, [
+      ...DEFAULT_EXTENSIONS,
+      ...customCompilerExtensions,
+    ]);
```
webpro, line 75: I think this can go back as `extensions` in the `createCustomModuleResolver` scope (slight perf cost to unnecessarily keep re-creating this array with a spread operator).

packages/knip/src/typescript/resolveModuleNames.ts
```diff
@@ -91,12 +91,29 @@ export function createCustomModuleResolver(
 
       return {
         resolvedFileName,
-        extension: virtualFileExtensions.includes(ext) ? ts.Extension.Js : ext,
+        extension: getExtension(resolvedFileName),
```
webpro, line 94: Why are we using a function with `customCompilerExtensions.some` and `customCompilerExtensions.includes`? Maybe we can make it consistent/fast with a Set + `Set.has`?

webpro (approved):
Thanks @geigerzaehler, this is great! Just what I was hoping for. Given that it also handles the external repos just fine I'm confident we can merge this.

Just a few nitpicks left.

packages/knip/src/typescript/resolveModuleNames.ts
geigerzaehler: Makes sense and fixed

packages/knip/src/typescript/resolveModuleNames.ts
geigerzaehler: Good point. I removed the function and made it more consistent.

webpro:
This is so good! Thanks, @geigerzaehler.

webpro:
:rocket: _This pull request is included in v5.20.0. See [Release 5.20.0](https://github.com/webpro-nl/knip/releases/tag/5.20.0) for release notes._

PR #682, opened 2024-06-18
feat: add command to ignored binaries

tim-richter:
Hey, firstable thanks for this awesome project!

'command' is part of npm https://www.npmjs.com/package/command, but it's not maintained (last publish 10 years ago). 'command' is also a unix command that, for example, can check if a command is available in the current shell. This creates a false positive whenever 'command' is used.

We use it for example inside of a husky pre-commit hook like this:

```bash
if ! command -v yalc check &> /dev/null
then
    echo "yalc could not be found"
    exit
fi
```

This results in knip flagging this as a unlisted binary:

```
Unlisted binaries (1)
command  .husky/pre-commit
```

I've added a test for this in the husky v9 fixture, but it's actually not really related to husky, because this could be used in any kind of bash script. If there is a better place for this test, let me know.

### DIFF
--- a/packages/knip/fixtures/plugins/husky-v9/.husky/pre-commit
+++ b/packages/knip/fixtures/plugins/husky-v9/.husky/pre-commit
@@ -1,3 +1,9 @@
 npx lint-staged
 npx --no-install commitlint --edit "$1"
 npm test
+
+if ! command -v yalc check &> /dev/null
+then
+    echo "yalc could not be found"
+    exit
+fi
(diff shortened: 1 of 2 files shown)


webpro:
Absolutely! Thanks @tim-richter 

```
❯ last-publish command
The last publish date of command is 6/9/2014
```

webpro:
> I've added a test for this in the husky v9 fixture, but it's actually not really related to husky, because this could be used in any kind of bash script. If there is a better place for this test, let me know.

Btw, I do this more often, "merging" fixtures/tests. Can't have everything separated perfectly, would grow into something too large/slow of a test suite imho.

webpro:
:rocket: _This pull request is included in v5.21.1. See [Release 5.21.1](https://github.com/webpro-nl/knip/releases/tag/5.21.1) for release notes._

PR #693, opened 2024-06-21
Add support for graphql-codegen plugin level config (resolves #692)

### LINKED ISSUE #692: graphql-codegen plugin doesn't handle plugins with plugin-specific config
The graphql-codegen plugin doesn't seem to correctly detect codegen plugins defined in the config using the "[plugin level](https://the-guild.dev/graphql/codegen/docs/config-reference/config-field#plugin-level)" configuration. 

I noted this on [ee83151c812258c04931f231fa8315fb405d2979](https://github.com/webpro-nl/knip/commit/ee83151c812258c04931f231fa8315fb405d2979#commitcomment-143405155), which actually introduced a test case with the data required to verify this use-case - in the fixture data added in the commit, the `add` plugin is declared, but is not detected by knip.

CensoredUser:
This ensures knip finds references to graphql-codegen plugins with plugin-level configuration, and can detect these as used, unused or unlisted dependencies.


webpro:
Thanks @CensoredUser! Short 'n sweet 👌

PR #723, opened 2024-07-11
Support subpath import with arbitrary extensions

geigerzaehler:
Knip stops reporting issues for subpath imports with arbitrary extensions (e.g. `import "#src/foo.ext"`). This already worked for path mappings defined in `tsconfig.json` but now it also works for mappings defined in `package.json`.

When trying different approaches I found it easiest to go back to the initial idea of fooling TypeScript into believing files with non-code extensions exists. But the implementation is much simpler now. And I tried to document the approach in the code.


packages/knip/src/typescript/resolveModuleNames.ts
```diff
@@ -35,6 +34,32 @@ export function createCustomModuleResolver(
   const customCompilerExtensionsSet = new Set(customCompilerExtensions);
   const extensions = [...DEFAULT_EXTENSIONS, ...customCompilerExtensions];
 
+  const virtualDeclarationFiles = new Map<string, { path: string; ext: string }>();
+
+  const tsSys: ts.System = {
+    ...ts.sys,
+    // We trick TypeScript into resolving paths with arbitrary extensions. When
+    // a module "./foo.ext" is imported TypeScript only tries to resolve it to
+    // "./foo.d.ext.ts". TypeScript never checks whether "./foo.ext" itself exists.
```
webpro, line 43: Don't you mean `module.ext.ts` and `module.ext.d.ts` (and same for `.mts` + `.cts`)?

packages/knip/src/typescript/resolveModuleNames.ts
```diff
@@ -147,3 +155,19 @@ export function createCustomModuleResolver(
 
   return timerify(resolveModuleNames);
 }
+
+const declarationPathRe = /^(.*)\.d(\.[^.]+)\.ts$/;
```
webpro, line 159: This matches a lot, but I guess I'm having the same confusion as in the comment: I think you're saying "TypeScript tries to resolve `foo.ext` to `foo.d.ext.ts`", which I doubt is true? Doesn't it resolve `index` to `index.d.ts` or `anything.really` to `anything.really.d.ts`?

webpro (commented):
Thanks for the PR! Love the idea. Had virtual file paths before but this is a more generic one. Good stuff. Some preliminary comments before I'll play a bit with it later today or tomorrow. Tests and integration test are passing so looking great already!

webpro:
> We stop raising issues for subpath imports with arbitrary extensions 

Just to be sure: do you mean Knip doesn't report false positives, or we've stopped opening issues for bugs? :)

packages/knip/src/typescript/resolveModuleNames.ts
geigerzaehler: I was confused by this as well but it is correct. If the import has an unknown extension like `.../foo.ext` typescript looks for `.../foo.d.ext.ts` ([code in TypeScript](https://github.com/microsoft/TypeScript/blob/46410044add2e9f53cea58e445de18dcda53443f/src/compiler/moduleNameResolver.ts#L2161)).

geigerzaehler:
> Just to be sure: do you mean Knip doesn't report false positives, or we've stopped opening issues for bugs? :)

Good point, :sweat_smile:. Knip stops reporting “unresolved” for imports like `#src/foo.ext` where the file after, subpath import mapping, exists

packages/knip/src/typescript/resolveModuleNames.ts
webpro: TIL! Thanks for figuring that out + the link, much appreciated.

packages/knip/src/typescript/resolveModuleNames.ts
```diff
@@ -35,6 +34,32 @@ export function createCustomModuleResolver(
   const customCompilerExtensionsSet = new Set(customCompilerExtensions);
   const extensions = [...DEFAULT_EXTENSIONS, ...customCompilerExtensions];
 
+  const virtualDeclarationFiles = new Map<string, { path: string; ext: string }>();
+
+  const tsSys: ts.System = {
```
webpro, line 39: Unless there's a good reason not to, I think we could move `virtualDeclarationFiles` and `tsSys` out of the function scope, easier for moving/refactors later.

packages/knip/src/typescript/resolveModuleNames.ts
```diff
@@ -147,3 +155,19 @@ export function createCustomModuleResolver(
 
   return timerify(resolveModuleNames);
 }
+
+const declarationPathRe = /^(.*)\.d(\.[^.]+)\.ts$/;
+
+/**
+ * For paths that look like `.../foo.d.yyy.ts` returns path `.../foo.yyy` and
```
webpro, line 162: Trying to use "foo" as little as possible myself, any chance you could rid of that too? I guess "module" could do.

packages/knip/fixtures/subpath-patterns/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "my-package",
   "imports": {
-    "#internals/*": "./src/internals/*.ts"
+    "#internals/*": "./src/internals/*.ts",
+    "#internals-explicit-ext/*": "./src/internals/*"
```
webpro, line 5:
```suggestion
    "#internals-explicit-ext/*": "./src/internals/*",
    "#internals-alias/used.alt": "./src/internals/used.alt"
```

And that file too then please :)

packages/knip/package.json
```diff
@@ -75,7 +75,6 @@
     "smol-toml": "^1.1.4",
     "strip-json-comments": "5.0.1",
     "summary": "2.1.0",
-    "tsconfig-paths": "^4.2.0",
```
webpro, line 78: 🎉

webpro (approved):
Fantastic! Thanks for coming back to this. Tested a bit, works great. Slightly faster in my testing too. What more could I ask for?

packages/knip/src/typescript/resolveModuleNames.ts
geigerzaehler: My thinking behind this was that `virtualDeclarationFiles` is specific to the instance of `resolveModuleName` that is created by `createCustomModuleResolver`. I think it is theoretically possible for two instances with different compiler settings to interfere with each other. Also, keeping `virtualDeclarationFiles`  local and not global allows it to be garbage collected. Whether all this is important in practice, I don’t know. But conceptually I feel like stateful things should be localized.

If you agree, I’ll keep it there, otherwise I can move it out.

webpro: You're right!

webpro:
Thanks @geigerzaehler, excellent stuff!

webpro:
:rocket: _This pull request is included in v5.26.0. See [Release 5.26.0](https://github.com/webpro-nl/knip/releases/tag/5.26.0) for release notes._

Issue #268, opened 2023-09-25
Add Ladle


alecmev:
It's a lightweight Storybook alternative. [Landing page](https://ladle.dev/), [GitHub](https://github.com/tajo/ladle). Its configuration is stored in `.ladle/config.mjs`, along with a few other files, [source](https://ladle.dev/docs/config). It uses Vite, and a custom Vite config is sometimes needed, in a customizable location `viteConfig` (in our case we set it to `.ladle/vite.config.ts`, for example, to keep them together). Currently I get the following, just in case:

```
Unused files (2)
.ladle/config.mjs
.ladle/vite.config.ts
Unused devDependencies (2)
@vitejs/plugin-react-swc  package.json
vite                      package.json
```

webpro:
Maybe you (or someone else) is up for a pull request and [write a plugin](https://github.com/webpro/knip/blob/main/docs/writing-a-plugin.md)? Sounds like there are similar plugins, like the one for Vite?

webpro:
Moved.

PR #728, opened 2024-07-12
Add ladle plugin

justb3a:
[Ladle](https://ladle.dev/) is a lightweight Storybook alternative. Its configuration is stored in `.ladle/config.mjs`, along with some [other files](https://ladle.dev/docs/config). It uses Vite, a custom Vite config is sometimes necessary in a customizable location via `viteConfig`.

Related issue: #268


packages/knip/src/plugins/ladle/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const stories: string[] = ['**/*.@(stories.@(mdx|js|jsx|mjs|ts|tsx))'];
+const restEntry: string[] = ['.ladle/components.{js,jsx,ts,tsx}', 'head.html'];
+const entry: string[] = [...restEntry, ...stories];
+
+const project = ['.ladle/**/*.{js,jsx,ts,tsx}'];
+
+const resolveEntryPaths: ResolveEntryPaths<LadleConfig> = async localConfig => {
+  const localStories = typeof localConfig.stories === 'string' ? [localConfig.stories] : localConfig.stories;
+  const localViteConfig = localConfig.viteConfig ? [localConfig.viteConfig] : [];
```
webpro, line 25: The Vite config should be handled by the Vite plugin, so I think we can just ignore it here? Btw it would need be a `config` file, not an `entry` file.

packages/knip/src/plugins/ladle/index.ts
```diff
@@ -0,0 +1,40 @@
...
+const title = 'ladle';
+
+const enablers: EnablerPatterns = [/^@ladle\//];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['.ladle/config.{mjs,js,ts}'];
+
+const stories: string[] = ['**/*.@(stories.@(mdx|js|jsx|mjs|ts|tsx))'];
+const restEntry: string[] = ['.ladle/components.{js,jsx,ts,tsx}', 'head.html'];
```
webpro, line 18: Knip does not support `head.html` as an entry file (unless a custom `.html` compiler is installed).

packages/knip/src/plugins/ladle/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const title = 'ladle';
+
+const enablers: EnablerPatterns = [/^@ladle\//];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['.ladle/config.{mjs,js,ts}'];
+
+const stories: string[] = ['**/*.@(stories.@(mdx|js|jsx|mjs|ts|tsx))'];
```
webpro, line 17: Please use the same defaults as Ladle itself. Users can override in Knip config if necessary.

webpro (commented):
Thanks @justb3a! A Ladle plugin is a great addition to Knip.

My main point of feedback is that the fixtures are a bit minimal. Ideally there would be more configuration with stories and components to verify the right files are indeed used as entry files (and they'll probably use dependencies that are then no longer unused).

packages/knip/src/plugins/ladle/index.ts
justb3a: For ladle it might be necessary to have a dedicated / different viteConfig file. So in my case I have `~/.vite.config.ts` and for ladle `~/.ladle/vite.config.ts`. Only if you don't use the default one, this parameter is set and then should be respected. Otherwise, I get it listed as "unused file".

webpro: OK in that case I think it's interesting to look into using the `resolveConfig` function of the Vitest plugin (which is basically the same as the Vite plugin).

Simplified inside `resolveConfig`: `const viteDeps = resolveConfig(await load(localConfig.viteConfig))`

Does that make sense?

webpro:
> I added some commits on top and extended the fixtures. If you want to have something else regarding fixtures/tests pls give me some hints, I checked some other plugins and couldn't find something helpful.

- A Ladle config file
- You mentioned a Vite config file (did you see https://github.com/webpro-nl/knip/pull/728#discussion_r1677497393?)
- There should be a `.ladle` directory with specific files?

Most plugins should have fixtures for the config and entry files they're referencing in the implementation.

> I did not add `node_modules` folder because this one would be quite big even with the limited set of package json entries.

Great, that would be too much indeed. Sometimes a basic `definConfig` might come in useful, such as the one for Vite (https://github.com/webpro-nl/knip/blob/main/packages/knip/fixtures/plugins/vite/node_modules/vite/index.js).

packages/knip/src/plugins/ladle/index.ts
justb3a: Changed 👍 I wasn't aware that it is allowed to use other plugins.

justb3a: … I cannot get rid of the message that the file is unused, stuck here.

```
Unused files (1)
.ladle/vite.config.ts
```

I also tried things like 

```
const resolveConfig: ResolveConfig<LadleConfig> = async (localConfig, options) =>
  resolveVitestConfig(localConfig.viteConfig ? await load(join(process.cwd(), localConfig.viteConfig)) : [], options);
```

Maybe someone with more insights can help here? 😇

webpro: Perhaps confusingly, loading the file does not "use" it (because it's not part of the dependency graph). For this you could add it as an entry file (contrary to what I said earlier, sorry, but that was for another reason).

packages/knip/src/plugins/ladle/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const resolveEntryPaths: ResolveEntryPaths<LadleConfig> = async localConfig => {
+  const localStories = typeof localConfig.stories === 'string' ? [localConfig.stories] : localConfig.stories;
+  const patterns = [...restEntry, ...(localStories ?? stories)];
+
+  return patterns.map(toEntryPattern);
+};
+
+const resolveConfig: ResolveConfig<LadleConfig> = async (localConfig, options) =>
+  localConfig.viteConfig ? resolveVitestConfig(await load(localConfig.viteConfig), options) : [];
```
webpro, line 32: Please expand it a bit and use something like `join(options.configFileDir, localConfig.viteConfig)` so that it is resolved relative the Ladle config file dir.

webpro: Oh never mind, now I read https://ladle.dev/docs/config/#viteconfig and see that the path should be absolute anyway (wasn't clear from the fixture).

justb3a: Not sure, just the example uses an absolute path, I could not find something in the documentation stating that is has to be absolute.

I think it would be great to support both, so I now check if the path is already absolute and if not, I join it with `options.cwd`. (I did not check before what's inside options, so helpful 🤣 ).

At least on my machine with our config it now works like a charm :)

packages/knip/src/plugins/ladle/index.ts
justb3a: I re-added it additionally as an entry file 👍

justb3a:
@webpro do you have any idea why the test is failing in CI?

```
test/plugins/ladle.test.ts:
  dyld[2233]: missing symbol called
  error: script "test" was terminated by signal SIGABRT (Abort)
  /Users/runner/work/_temp/e765fe65-ff66-46a8-84ea-064701d7180d.sh: line 1:  2232 Abort trap: 6           bun run test
  Error: Process completed with exit code 134.
 ```
  
  locally it's working

```
❯ bun test test/plugins/ladle.test.ts
bun test v1.1.18 (5a0b9352)

test/plugins/ladle.test.ts:
✓ Find dependencies with the ladle plugin [135.67ms]

 1 pass
 0 fail
Ran 1 tests across 1 files. [473.00ms]
```

webpro:
Yeah, crazy. I'll take a look tonight.

webpro:
The issue seems to be with Bun 1.1.19, 1.1.20 handles it fine.

webpro:
But bun 1.1.20 has [another issue](https://github.com/oven-sh/bun/issues/4899), so I'm afraid we'll have to wait for a fix. Still no idea what eventually causes the issue in this PR 🤷‍♂️

alecmev:
This is great, thanks! 👍

webpro:
Thanks to @justb3a! There was an issue with the release script, apparently the release notes were too long, but this PR is part of [v5.27.0](https://github.com/webpro-nl/knip/releases/tag/5.27.0).

PR #732, opened 2024-07-16
feat(vite): add vite-env file for entrypoints

ivanesik:
Vite provides type definitions file `vite-end.d.ts` - https://vitejs.dev/guide/env-and-mode.html#intellisense-for-typescript.

But now this file gives an error:
Screenshot 2024-07-16 at 13 06 53

Fast search shows that people include this file in ignores: https://github.com/search?q=path%3A**%2Fknip**+vite-env.d.ts&type=code&ref=advsearch
In this PR append path to file in config list

### DIFF
--- a/packages/knip/src/plugins/vite/index.ts
+++ b/packages/knip/src/plugins/vite/index.ts
@@ -12,6 +12,8 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
 
 export const config = ['vite*.config.{js,mjs,ts,cjs,mts,cts}'];
 
+const entry = ['src/vite-env.d.ts'];
+
 const production: string[] = [];
 
 export default {
@@ -19,6 +21,7 @@ export default {
   enablers,
   isEnabled,
   config,
+  entry,
   production,
   resolveEntryPaths,
   resolveConfig,
--- a/packages/knip/src/plugins/vite/index.ts
+++ b/packages/knip/src/plugins/vite/index.ts
@@ -10,9 +10,7 @@ const enablers = ['vite', 'vitest'];
 
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
-export const config = ['vite*.config.{js,mjs,ts,cjs,mts,cts}'];
-
-const entry = ['src/vite-env.d.ts'];
+export const config = ['vite*.config.{js,mjs,ts,cjs,mts,cts}', 'src/vite-env.d.ts'];
 
 const production: string[] = [];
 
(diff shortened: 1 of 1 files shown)


webpro:
Yeah, good one :) Thanks @ivanesik!

packages/knip/src/plugins/vite/index.ts
```diff
@@ -10,7 +10,7 @@ const enablers = ['vite', 'vitest'];
 
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
-export const config = ['vite*.config.{js,mjs,ts,cjs,mts,cts}'];
+export const config = ['vite*.config.{js,mjs,ts,cjs,mts,cts}', 'src/vite-env.d.ts'];
```
webpro, line 13: It should be an `entry` file (to be added as source file in the dependency graph), `config` files are meant to be loaded and parsed.

ivanesik: Entry doesn't works in my project ;(
I'll try to understand why

ivanesik: Maybe because it's a type definition file

ivanesik: I found the reason.
1. In Vite plugin we already have `resolveEntryPaths` - https://github.com/webpro-nl/knip/blob/12d14c6bee72625eb1ded52301acd08585aa2943/packages/knip/src/plugins/vite/index.ts#L23
2. This `resolveEntryPaths` will return the entry path here - https://github.com/webpro-nl/knip/blob/12d14c6bee72625eb1ded52301acd08585aa2943/packages/knip/src/WorkspaceWorker.ts#L302
3. And it is taken by priority here, ignoring entries - https://github.com/webpro-nl/knip/blob/12d14c6bee72625eb1ded52301acd08585aa2943/packages/knip/src/util/plugin.ts#L76

I see two options to fix this.
1. to merge `entry` and result of `resolveEntryPaths` in `getFinalEntryPaths`, but I think thats not correct. 
2. customize `resolveEntryPaths` in Vite plugin description like
```ts
const viteResolveEntryPaths = async (localConfig, options) => {
    const vitestEntryPaths = await resolveEntryPaths(localConfig, options);
    return [...vitestEntryPaths, toEntryPattern('src/vite-env.d.ts')];
};

export default {
    title,
    enablers,
    isEnabled,
    config,
    production,
    resolveEntryPaths: viteResolveEntryPaths,
    resolveConfig,
};
```

And select second one. @webpro what do you think?

webpro:
I sneaked it in in some other way, hope you don't mind :)

ivanesik:
> I sneaked it in in some other way, hope you don't mind :)

Of course I don't mind. Thanks for the fix! 💪

Issue #724, opened 2024-07-11
Cannot run `pnpm create @knip/config` in nx monorepo root


pavinduLakshan:
I was running `pnpm create @knip/config` in my nx monorepo root and the execution fails with the following error.

```
.../Library/pnpm/store/v3/tmp/dlx-15451  |   +1 +
.../Library/pnpm/store/v3/tmp/dlx-15451  | Progress: resolved 1, reused 1, downloaded 0, added 1, done
node:internal/errors:865
  const err = new Error(message);
              ^

Error: Command failed: pnpm add -D knip typescript @types/node
    at checkExecSyncError (node:child_process:885:11)
    at execSync (node:child_process:957:15)
    at main (file:///Users/pavindu/Library/pnpm/store/v3/tmp/dlx-15451/node_modules/.pnpm/@knip+create-config@1.0.1/node_modules/@knip/create-config/index.js:26:3)
    at file:///Users/pavindu/Library/pnpm/store/v3/tmp/dlx-15451/node_modules/.pnpm/@knip+create-config@1.0.1/node_modules/@knip/create-config/index.js:35:1
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25) {
  status: 1,
  signal: null,
  output: [
    null,
    Buffer(352) [Uint8Array] [
      226, 128, 137,  69,  82,  82,  95,  80,  78,  80,  77,  95,
       65,  68,  68,  73,  78,  71,  95,  84,  79,  95,  82,  79,
       79,  84, 226, 128, 137,  32,  82, 117, 110, 110, 105, 110,
      103,  32, 116, 104, 105, 115,  32,  99, 111, 109, 109,  97,
      110, 100,  32, 119, 105, 108, 108,  32,  97, 100, 100,  32,
      116, 104, 101,  32, 100, 101, 112, 101, 110, 100, 101, 110,
       99, 121,  32, 116, 111,  32, 116, 104, 101,  32, 119, 111,
      114, 107, 115, 112,  97,  99, 101,  32, 114, 111, 111, 116,
       44,  32, 119, 104,
      ... 252 more items
    ],
    Buffer(0) [Uint8Array] []
  ],
  pid: 15471,
  stdout: Buffer(352) [Uint8Array] [
    226, 128, 137,  69,  82,  82,  95,  80,  78,  80,  77,  95,
     65,  68,  68,  73,  78,  71,  95,  84,  79,  95,  82,  79,
[...]

PR #738, opened 2024-07-23
fix: pnpm workspace fixes

ChiefORZ:
closes https://github.com/webpro-nl/knip/issues/724

### DIFF
--- a/packages/create-config/index.js
+++ b/packages/create-config/index.js
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 import { execSync } from 'node:child_process';
 import { statSync } from 'node:fs';
+// biome-ignore lint/nursery/noRestrictedImports: ignore
+import path from 'node:path';
 
 const fileExists = filePath => {
   const stat = statSync(filePath, { throwIfNoEntry: false });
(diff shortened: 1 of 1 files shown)


packages/create-config/index.js
```diff
@@ -1,16 +1,21 @@
...
   const stat = statSync(filePath, { throwIfNoEntry: false });
   return stat?.isFile();
 };
 
 const getPackageManager = () => {
-  if (fileExists('bun.lockb')) return 'bun';
-  if (fileExists('yarn.lock')) return 'yarn';
-  if (fileExists('pnpm-lock.yaml')) return 'pnpm';
+  // get the root of the repository
+  const repositoryRoot = execSync('git rev-parse --show-toplevel').toString().trim();
```
webpro, line 14: If we're going to use this here, I guess we should also use it for `fileExists('pnpm-workspace.yaml')` and when running `execSync(cmd)`? If so, please either remove it or apply everywhere.

webpro (commented):
Thanks @ChiefORZ! Happy to merge, but I do have one comment.

packages/create-config/index.js
ChiefORZ: I see where you are coming from, but i think it is ok as it is (from the perspective of pnpm at least).
Because i could intentionally execute this command only in a sub-package of a pnpm workspace:

```sh
cd ./my-repository/packages/my-project
pnpm create @knip/config
```
Which would add the dependency only to the `./my-repository/packages/my-project/package.json`.

Whereas this would add it to the monorepo root (`./my-repository/package.json`):

```sh
cd ./my-repository
pnpm create @knip/config
```

webpro:
:rocket: _This pull request is included in v5.27.1. See [Release 5.27.1](https://github.com/webpro-nl/knip/releases/tag/5.27.1) for release notes._

Issue #614, opened 2024-04-30
Autofix - remove trailing commas


weilbith:
Hey ✌🏾 

I know [the documentation](https://knip.dev/features/auto-fix/#whats-not-included) states that removing empty export declarations are not included in auto-fixes. But I wonder if maybe training commas could be removed. Here an example to better understand what I mean:

```typescript
export { isNotUsed, isActuallyUsed } from "./whatever"
```

If I run `knip --fix --fix-type exports` I'm left with this:

```typescript
export { , isActuallyUsed } from "./whatever"
```

In a more automated environment (i.e. git hooks), I could solve this by running auto-formatting on these files again. But maybe its quite feasible to do this right away. I seriously don't know how hard this is having the AST at hand. 🙈 

Thanks for any feedback to this idea. 🙏🏾

weilbith:
Sorry, disregard that issue. I think my idea with the auto-formatting after `knip --fix` is the most feasible solution and does not bring up any additional conflicts with formatting configuration.
Because I stumbled across more scenarios and I think that's just not _knip_ its responsibility.

Example:
```typescript
export {
  isActuallyUsed,
  isNotUsed,
  alsoNotUsed,
} from "./whatever"
```
becomes
```typescript
export {
  isActuallyUsed,
  ,
  ,
} from "./whatever"
```
but should become after formattting
```typescript
export { isActuallyUsed } from "./whatever"
```

Sorry for the noise 🙈

weilbith:
Though, is there a way to run `knip --fix` with some output, so I get the list of files that have actually been touched and I can just pipe them into the formatter script? 🤔

webpro:
The only "structured" way to do this is `knip --fix --reporter json`, but the JSON reporter doesn't contain the `isFixed` boolean yet. If it's there you could pipe that and use e.g. `jq`, etc.

If this sounds useful to you and you're up for a PR, then this `issue` object has the `isFixed` property that can be added to the `json` output object: https://github.com/webpro/knip/blob/main/packages/knip/src/reporters/json.ts#L74

weilbith:
Thank you very much. Very quick response again. 😊 
I realized that the whole topic is a little more complicated. Tools like `eslint` and `prettier` do not simply clean this up. And then again you might end up in a loop of running `knip --fix` then potentially remove code that has now become detected dead by `eslint` which might again trigger `knip` ♻️ 
[...]

PR #788, opened 2024-09-23
bugfix: Cleaning up floating commas after knip --fix

glemiron:
I would like to fix this issue: https://github.com/webpro-nl/knip/issues/614. 

This problem doesn't allows me to apply knip --fix on scale, because extra commas are breaking typescript check syntax and requires additional work to fix.   

I don't sure that is optimal, but I decided to open this PR to start a discussion.

### DIFF
--- a/packages/knip/src/IssueFixer.ts
+++ b/packages/knip/src/IssueFixer.ts
@@ -88,8 +88,13 @@ export class IssueFixer {
         );
 
         const withoutEmptyReExports = sourceFileText
+          .replaceAll(/export \{[ ,]+\};?\r?\n?/g, '')
           .replaceAll(/export \{[ ,]+\} from ('|")[^'"]+('|");?\r?\n?/g, '')
-          .replaceAll(/export \{[ ,]+\};?\r?\n?/g, '');
+          .replaceAll(/\{(\s*,)+/g, '{')
+          .replaceAll(/(,\s*)+}/g, '}')
+          .replaceAll(/(\s*,\s*,\s*)+/g, ',')
+          // previous replacement leaves 2 commas for odd commas number
+          .replaceAll(/,,/g, ',');
 
         await writeFile(filePath, withoutEmptyReExports);
 
--- a/packages/knip/src/IssueFixer.ts
+++ b/packages/knip/src/IssueFixer.ts
@@ -1,5 +1,5 @@
 import { readFile, rm, writeFile } from 'node:fs/promises';
-import type { Fixes } from './types/exports.js';
+import { type ExportsWithElements, type Fix, type Fixes, isFixForExportWithElements } from './types/exports.js';
 import type { Issues } from './types/issues.js';
 import { load, save } from './util/package-json.js';
 import { join, relative } from './util/path.js';
@@ -19,8 +19,8 @@ export class IssueFixer {
   isFixUnusedTypes = true;
   isFixUnusedExports = true;
 
-  unusedTypeNodes: Map<string, Set<[number, number]>> = new Map();
-  unusedExportNodes: Map<string, Set<[number, number]>> = new Map();
+  unusedTypeNodes: Map<string, Set<Fix>> = new Map();
+  unusedExportNodes: Map<string, Set<Fix>> = new Map();
 
   constructor({ isEnabled, cwd, fixTypes = [], isRemoveFiles }: Fixer) {
     this.isEnabled = isEnabled;
@@ -76,27 +76,36 @@ export class IssueFixer {
   private async removeUnusedExportKeywords(issues: Issues) {
     const filePaths = new Set([...this.unusedTypeNodes.keys(), ...this.unusedExportNodes.keys()]);
     for (const filePath of filePaths) {
-      const exportPositions: Fixes = [
+      const fixes: Fixes = [
         ...(this.isFixUnusedTypes ? (this.unusedTypeNodes.get(filePath) ?? []) : []),
         ...(this.isFixUnusedExports ? (this.unusedExportNodes.get(filePath) ?? []) : []),
-      ].sort((a, b) => b[0] - a[0]);
+      ].sort((a, b) => b!.pos[0] - a!.pos[0]); // TODO how to handle types?
 
+      // TODO add type guard
+      const exportPositions = fixes;
       if (exportPositions.length > 0) {
-        const sourceFileText = exportPositions.reduce(
-          (text, [start, end]) => text.substring(0, start) + text.substring(end),
-          await readFile(filePath, 'utf-8')
-        );
+        console.log(exportPositions);
+        const file = await readFile(filePath, 'utf-8');
+        const sourceFileText = exportPositions.reduce((text, fix) => {
+          if (isFixForExportWithElements(fix)) {
+            // TODO this could be called multiple times
+            const allElementFixesForExport = exportPositions
+              .filter(isFixForExportWithElements)
+              .filter(f => f.pos[0] === fix.pos[0] && f.pos[1] === fix.pos[1])
+              .map(f => f.element);
 
-        const withoutEmptyReExports = sourceFileText
-          .replaceAll(/export \{[ ,]+\};?\r?\n?/g, '')
-          .replaceAll(/export \{[ ,]+\} from ('|")[^'"]+('|");?\r?\n?/g, '')
-          .replaceAll(/\{(\s*,)+/g, '{')
-          .replaceAll(/(,\s*)+}/g, '}')
-          .replaceAll(/(\s*,\s*,\s*)+/g, ',')
-          // previous replacement leaves 2 commas for odd commas number
-          .replaceAll(/,,/g, ',');
+            const allElements = fix.allElements.filter(element => !allElementFixesForExport.includes(element));
+            debugger;
+            if (!allElements.length) {
+              return text.substring(0, fix!.allExportPos[0]) + text.substring(fix!.allExportPos[1]);
+            }
+            return text.substring(0, fix.pos[0]) + `{${allElements.join(', ')}}` + text.substring(fix.pos[1]);
+          } else {
+            return text.substring(0, fix!.pos[0]) + text.substring(fix!.pos[1]);
+          }
+        }, file);
 
-        await writeFile(filePath, withoutEmptyReExports);
+        await writeFile(filePath, sourceFileText);
 
         this.markExportFixed(issues, filePath);
       }
--- a/packages/knip/src/typescript/visitors/exports/exportDeclaration.ts
+++ b/packages/knip/src/typescript/visitors/exports/exportDeclaration.ts
@@ -1,4 +1,4 @@
-import ts from 'typescript';
+import ts, { preProcessFile } from 'typescript';
 import type { Fix } from '../../../types/exports.js';
 import { SymbolType } from '../../../types/issues.js';
 import type { BoundSourceFile } from '../../SourceFile.js';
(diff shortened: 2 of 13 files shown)


glemiron:
It seems there's no way to solve it without a regexp, because we don't have enough information inside `reduce` to be able safely decide if removal of commas is needed.

Our primary targets for this operation are ``export {....}``, ``module.exports = { ... }`` and ``export { Row as KRow, Row } from './reexported'``. In the same time, exports with square brackets must be avoided, because removal commas from them will be unsafe.

Example:
```
// before
export const [c, d] = [1, 2];

// after c removed
export const [d] = [1, 2];

// expected; comma makes a difference! 
export const [,d] = [1, 2];
```

I see the following alternative approach to solve this problem within the reduce function:

I see a following alternative way to solve this problem inside the reduce: 
1. Use a regular expression to find all export positions where a comma fix needs to be applied.
2. For each removed part, check if it belongs to an export and, if so, remove exactly one comma after (before for the last) the variable being removed.

It’s now clear for me that the removal of commas should only be applied within the export context. I'll consider if there's a way to simplify this logic further.

Happy to hear more ideas, thanks for the prompt reply!

webpro:
> It seems there's no way to solve it without a regexp, because we don't have enough information inside reduce to be able safely decide if removal of commas is needed.

We have the full `text` contents of the source file available. Let's write out the function with some psuedo code/high-level ideas:

```ts
const sourceFileText = exportPositions.reduce(
  (text, [start, end]) => {
    const before = text.substring(0, start); 
    const subject = text.substring(start, end); 
    const after = text.substring(end);

    // if subject is e.g. "export " we can return result
    return before + after;

   // otherwise, we could look to the left or right until `{` or `[`
   // if only e.g. `export { }` would remain, we could remove all this
   return before? + after?
  },
  await readFile(filePath, 'utf-8')
);
```

Inside the reducer, we have the full text, so we can "look around" and see if we're inside square or curly brackets, etc. I think this "inside-out" approach is safer than applying regular expressions "after the fact". We've done the hard work during AST traversal so we can assume we're in the right position to remove things.

Would you agree and like to give this a stab?

glemiron:
I wasn't able to make this assumption without full context:  
> We've done the hard work during AST traversal so we can assume we're in the right position to remove things.

Now I have everything to solve it!

webpro:
Ah, great! One example of an AST visitors is https://github.com/webpro-nl/knip/blob/main/packages/knip/src/typescript/visitors/exports/exportDeclaration.ts and most of them should have `// Pattern` comments to show what the matching source code looks like. The returned `fix` value has the `[start, end]` tuple.

glemiron:
@webpro what do you think about fixing this issue on the exportDeclaration level? It'll be much easier. 

Imagine only one element left inside an import, then we'll have one child inside node.exportClause.elements and we need to return a tuple for whole node instead of element. Similar with commas, we have all information there to decide what must go. 

With this approach IssueFixer.ts will be responsible only for cutting right piece of text and exportDeclaration will be fully responsible for witch part must be removed.

webpro:
The result of this first phase (of which AST traversal is an important part) is a module graph. It's serializable and it won't contain AST tree or nodes we could, indeed, use to more easily delete unused exports. It would be (too) expensive to recreate it for the "fixer phase".

The module graph is serializable which has a few advantages:

- It can be [cached](https://knip.dev/reference/cli#--cache) and stored on disk
- In large codebases/monorepos multiple "programs" becomes heavy/slow
  - https://knip.dev/reference/faq#why-doesnt-knip-just-use-tsfindreferences
  - https://knip.dev/blog/slim-down-to-speed-up
- The module graph as a data structure can be used for other features and perhaps even by other libraries, without relying on TypeScript.

However, we might still be able to have some wins when we're still working with the AST. One idea that's perhaps interesting to include the separator (comma) in the `end` pos during AST traversal if the tree/node contains this information (i.e. if it's easier/faster compared to doing this in the reducer in the "fixer phase").

In situations with a single item like `export { oneValue }` we could store `start` and `end` of the whole declaration, but for more than one item we would need the same logic anyway to check for commas and empty `export {}` declaration, so I don't think this helps a lot.

That said, the current `fix: [start, end]` tuple is a first idea I had, if there's more information we need to store to make our lives easier later on then that's certainly possible.

I should also add a different lib that might be interesting for you, but haven't tried myself yet: https://knip.dev/explanations/comparison-and-migration#ts-unused-exports, no idea if it could work in your situation.

Let me know what you think!

webpro:
@glemiron I couldn't resist.. Sorry, not trying to dismiss your work here, but your PR just got me excited to work on it!

This might be step forward. My first attempt is in this branch/commit: https://github.com/webpro-nl/knip/commit/c3248612279eb26d5291f96c114c9da3c468d70e

The parser/cleaner is in a separate function so it's easier to test and work with.

Installation:

```
npm i -D https://pkg.pr.new/knip@c324861
```

Would love to hear what you think and how it behaves on your project. Feel free to try it out and provide feedback if you have any. Let's make it happen :)

glemiron:
Nice one! I don't mind to have your solution if it solves the problem.

I like your approach due to its minimalism, I didn't even think to pass something as 3 argument of the pos tuple.

I've drafted a solution from a bit different perspective, I've pushed unfinished change, so you can check the idea. A while ago I was solving a similar problem and instead of deleting anything, it's possible to construct import from scratch.

It's fun problem to solve and a good example that one problem has a lot of different solutions!

webpro:
That's awesome, tonight I'll some more time and for sure I'll study your solution too. My solution is def not ready, let's combine best of both worlds :)

packages/knip/test/fix.test.ts
```diff
@@ -47,7 +47,7 @@ export type U = number;
       `const identifier = 1;
 const identifier2 = 2;
 
-module.exports = { identifier,  };
+module.exports = { identifier};
```
webpro, line 50: It's fine if we can remove this comma, but if we don't it shouldn't be an issue. My assumption is that eventually a formatter needs to fix up things anyway.

We should also consider that some formatters are set like `trailingComma: "<all|es5|none>"` so might not even want to remove the comma behind the export we're removing.

packages/knip/src/IssueFixer.ts
```diff
@@ -76,22 +76,36 @@ export class IssueFixer {
...
+              .filter(isFixForExportWithElements)
+              .filter(f => f.pos[0] === fix.pos[0] && f.pos[1] === fix.pos[1])
+              .map(f => f.element);
+
+            const allElements = fix.allElements.filter(element => !allElementFixesForExport.includes(element));
+            debugger;
+            if (!allElements.length) {
+              return text.substring(0, fix!.allExportPos[0]) + text.substring(fix!.allExportPos[1]);
+            }
+            return text.substring(0, fix.pos[0]) + `{${allElements.join(', ')}}` + text.substring(fix.pos[1]);
```
webpro, line 102: This looks like we're re-assembling the exported items. How does that affect the original source code with regards to comments and whitespace?

packages/knip/src/typescript/visitors/exports/exportDeclaration.ts
```diff
@@ -1,4 +1,4 @@
-import ts from 'typescript';
+import ts, { preProcessFile } from 'typescript';
```
webpro, line 1: TIL :) But looks like we don't need it?

packages/knip/src/typescript/visitors/exports/exportDeclaration.ts
```diff
@@ -15,15 +15,27 @@ export default visit(
...
-          const fix: Fix = isFixExports || isFixTypes ? [element.getStart(), element.getEnd()] : undefined;
-          return { node: element, symbol, identifier, type, pos, fix };
+
+          const exportClause = node.exportClause;
+          const fix1: Fix =
+            isFixExports || isFixTypes
+              ? {
+                  element: element.getFullText(),
+                  pos: [exportClause!.getStart(), exportClause!.getEnd()], //  TODO wrong, must be coordinates of {}
+                  allElements: allElementStrings,
```
webpro, line 33: Re. `allElementStrings`: the idea is to do the necessary work for the `IssueFixer` only if `isFixExports || isFixTypes` is true.

The bigger question is: do we want to calculate and store things upfront (like here) or do the heavy lifting later on in the `IssueFixer` (like the parser I did).

Here in AST area it's easier and more accurate. But we are probably doing a lot more work than necessary, because we're doing it for each export even though we may not remove it (usually a relatively small percentage is unused). That said, performance isn't the most important thing in `--fix` mode.

In a custom parser "after the fact" we're doing the work efficiently, but it's a bit clumsy. As it stands it also seems a bit less intrusive or safer with regards to comments and such.

webpro (commented):
Thanks again for initiating this @glemiron! Here's a first round of comments.

glemiron:
It's now obvious that idea to rebuild import content from elements isn't viable. It'll be hard to keep formatting, because imports could be separated not only with spaces, but lines breaks as well.   

Now, it’s clear that we should do more heavy lifting on the fixerSide. However, we could simplify the logic a bit by passing information about boundaries of export elements “{}” to the fixerSide, then the search for a nearest comma could be simplified. 

Let me know if I can help with finalising your change. Thanks for your support, I’ve learned a lot about the library! I hope I can use this knowledge and contribute to another features.

webpro:
> It's now obvious that idea to rebuild import content from elements isn't viable. It'll be hard to keep formatting, because imports could be separated not only with spaces, but lines breaks as well.   
> 
> Now, it’s clear that we should do more heavy lifting on the fixerSide. However, we could simplify the logic a bit by passing information about boundaries of export elements “{}” to the fixerSide, then the search for a nearest comma could be simplified.

Whatever we'll end up with, your work and the conversation have been super useful. Happy to have explored multiple angles of problem and solution.

> Let me know if I can help with finalising your change. Thanks for your support, I’ve learned a lot about the library! I hope I can use this knowledge and contribute to another features.

Here's a new version that also supports `type { A }` and `{ type A }`. The implementation is still ugly but it does the job. Still needs a lot of testing. Would be great if you could try it on your project(s) too:

```
npm i https://pkg.pr.new/knip@6b7e30d
```


Feel free to hack on my branch and apply your ideas if you'd like. I won't be able to work on this for at least another day.

glemiron:
Applied to my project and found an edge-case:

```
// initial: 
export const {
  get: someName,
  set: someSetName,
} = getHelpers()

// after the applied fix:
export const {
  ,
  set: someSetName,
} = getHelpers()
```
Looks very promising indeed!  I'll able to work on this a little more only one the weekend, so I don't mind if you finish the feature without my help. I'll be happy to help with testing!

webpro:
> Applied to my project and found an edge-case

Thanks for reporting this! Should be fixed now.

> Looks very promising indeed! I'll able to work on this a little more only one the weekend, so I don't mind if you finish the feature without my help. I'll be happy to help with testing!

Thanks a lot, this is much fun :) Update:

- Fixed: handling of commas
- Added: removal of enum and class members

```
npm i https://pkg.pr.new/knip@e788811
```

glemiron:
The new version is tested and fully works for me. 

Thank you once again!

webpro:
:rocket: _This issue has been resolved in v5.31.0. See [Release 5.31.0](https://github.com/webpro-nl/knip/releases/tag/5.31.0) for release notes._

webpro:
For the record, with regards to the automated message, the changeset of this PR is not actually in this release. HOWEVER, thank you so much for initiating this @glemiron because we've managed to improve Knip a lot together here.

Issue #789, opened 2024-09-24
🔄 Function `isStartsLikePackageName` should accept scoped packages which starts with `.` or `_` too


heavybeard:
### Prerequisites

- [X] I've read the relevant [documentation](https://knip.dev)
- [X] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [X] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://codesandbox.io/p/sandbox/reverent-sammet-yl9drl

### Reproduction access

- [X] I've made sure the reproduction is publicly accessible

### Good version

5.27.3

### Bad version

5.27.4

### Description of the regression

Thanks to [this commit](https://github.com/webpro-nl/knip/commit/89780376c5e0fffb480408dab2895ca6ec9b7244#diff-3b861313c9e2d00e0d0d56c3e1c3fa533731db0097209dc154f992d823b89621R325) introduced in `5.27.4` version, packages like `@_sh/strapi-plugin-ckeditor` are filtered from `ignoreDependencies` array.

This is causing a circular error with "Unused dependecies" and "Unused item in `ignoreDependencies`"

![image](https://github.com/user-attachments/assets/440eafa9-d3bb-4144-b686-daef8dc75194)

In a quick investigation, I landed in `isStartsLikePackageName` regex that missing `_` and `.` symbols. Underscore and point are valid starting character for scoped packages as npm mention [here](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#:~:text=The%20names%20of%20scoped%20packages%20can%20begin%20with%20a%20dot%20or%20an%20underscore.%20This%20is%20not%20permitted%20without%20a%20scope.)

webpro:
Thanks for the report @heavybeard! Would you like to send in a pull request?

PR #790, opened 2024-09-24
accept _ and . in package names

heavybeard:
closes https://github.com/webpro-nl/knip/issues/789

### DIFF
--- a/packages/knip/src/util/modules.ts
+++ b/packages/knip/src/util/modules.ts
@@ -15,7 +15,7 @@ export const getPackageNameFromFilePath = (value: string) => {
   return value;
 };
 
-export const isStartsLikePackageName = (specifier: string) => /^@?[a-z0-9]/.test(specifier);
+export const isStartsLikePackageName = (specifier: string) => /^@?[a-z0-9._]/.test(specifier);
 
 export const isDefinitelyTyped = (packageName: string) => packageName.startsWith(`${DT_SCOPE}/`);
 
--- a/packages/knip/src/util/modules.ts
+++ b/packages/knip/src/util/modules.ts
@@ -15,7 +15,7 @@ export const getPackageNameFromFilePath = (value: string) => {
   return value;
 };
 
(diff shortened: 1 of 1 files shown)


packages/knip/src/util/modules.ts
```diff
@@ -15,7 +15,7 @@ export const getPackageNameFromFilePath = (value: string) => {
   return value;
 };
 
-export const isStartsLikePackageName = (specifier: string) => /^@?[a-z0-9]/.test(specifier);
+export const isStartsLikePackageName = (specifier: string) => /^@?[a-z0-9._]/.test(specifier);
```
webpro, line 18:
```suggestion
export const isStartsLikePackageName = (specifier: string) => /^(@[a-z0-9._]|[a-z0-9])/.test(specifier);
```

Regex could be improved perhaps, but at least we're not allowing non-scoped packages to start with `.` or `_`.

webpro (commented):
Thanks @heavybeard! I think your solution allows non-scoped packages to start with `.` or `_` which isn't valid?

packages/knip/src/util/modules.ts
heavybeard: Thanks to chatGPT, we could improve the regex like this

```suggestion
export const isStartsLikePackageName = (specifier: string) => /^(?:@([._]?[a-z0-9-]{1,213})\/)?[a-z0-9]([a-z0-9-_.]*)(\/[a-z0-9]([a-z0-9-_.]*))*$/gm.test(specifier);
```

Regex tests here https://regex101.com/r/Ukm3Fw/2

Or you could use this package https://www.npmjs.com/package/validate-npm-package-name

heavybeard: It's ok for me use your suggested version

webpro:
Thanks a lot @heavybeard! I'll release it soon.

webpro:
:rocket: _This pull request is included in v5.30.6. See [Release 5.30.6](https://github.com/webpro-nl/knip/releases/tag/5.30.6) for release notes._

Issue #795, opened 2024-10-01
🐛 Knip's gitignore parser does not handle `*#` correctly


Sjlver:
### Prerequisites

- [X] I'm using the latest version
- [X] I've read the relevant [documentation](https://knip.dev)
- [X] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [X] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [X] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

Sorry, don't have one

### Reproduction access

- [X] I've made sure the reproduction is publicly accessible

### Description of the issue

We have the following entry in one of our .gitignore files:

```gitignore
*#
```

I think it's supposed to ignore emacs backup files (or whatever other tool appends `#` to files).

Knip seems to misinterpret it:

```
$ yarn run knip --debug
...
[project-dir] Glob options
{
  patterns: [
    'project-dir/**/*.{js,mjs,cjs,jsx,ts,tsx,mts,cts}',
    '!back/package.json',
    '!back/tsconfig.json',
    '!back/tsconfig.*.json'
  ],
  cwd: '/redacted',
  dir: '/redacted/project-dir',
  gitignore: true,
  absolute: true,
  dot: true,
  ignore: [
     'project-dir/**/*',
  ]
}
[project-dir] Project paths (0)
[]
```

This results in no project files being found.

webpro:
> Sorry, don't have one

Can you explain why it wouldn't be helpful in this case?

mcous:
I hit this exact issue today, made a simple reproduction repository here: https://github.com/mcous/knip-795

mcous:
`.gitignore` contents

```gitignore
node_modules
*#
```

Relevant output of `knip --debug`

```
[*] Parsed gitignore files
{
  gitignoreFiles: [ '.git/info/exclude', '.gitignore' ],
  ignores: Set(5) {
    '.git',
    '**/node_modules/**',
    '.yarn',
    '**/node_modules',
    '**/*'
  },
  unignores: []
}
```

If I remove `*#` from the `.gitignore`:

```
[*] Parsed gitignore files
{
  gitignoreFiles: [ '.git/info/exclude', '.gitignore' ],
  ignores: Set(4) { '.git', '**/node_modules/**', '.yarn', '**/node_modules' },
  unignores: []
}
```

[...]

PR #797, opened 2024-10-02
Do not treat trailing `#` in gitignore as comment

mcous:
Just started using `knip` today and it was instrumental in some really nice cleanup in our codebase. I hit #795 right off the bat, though, and it took me a while to figure out what was going on. I figured I could say "thanks" by opening up a quick PR to fix the issue!

The logic in the gitignore parser that strips trailing `# ...` off of `.gitignore` lines appears to be wrong. In my testing, you cannot add a comment to the end of a `.gitignore` line in this manner (syntax highlighting be damned):

```gitignore
# this is a comment
dist # this is not a comment
```

This tracks with [git's documentation](https://git-scm.com/docs/gitignore) (emphasis mine) as well as my own messing around with git:

> A line **starting** with # serves as a comment. Put a backslash (`\`) in front of the **first** hash for patterns that begin with a hash.

The previous parsing logic (added in https://github.com/webpro-nl/knip/commit/e455527fb2cc58981775ad538f5d4a64a3c70b5b) stripped all non-backslashed `#...` out. I updated the logic to only strip a leading `\`, if it's in front of a hash, so the following cases will be handled properly:

```gitignore
trailing-hash-ignore#
\#leading-hash-ignore
just#some-file
```

### DIFF
--- a/packages/knip/src/util/glob-core.ts
+++ b/packages/knip/src/util/glob-core.ts
@@ -54,7 +54,7 @@ export const parseAndConvertGitignorePatterns = (patterns: string, ancestor?: st
     .split(/\r?\n/)
     .filter(line => line.trim() && !line.startsWith('#'))
     .flatMap(line => {
-      const pattern = line.replace(/(?<!\\)#.*/, '').trim();
+      const pattern = line.replace(/^\\(?=#)/, '').trim();
       if (ancestor && matchFrom) {
         if (pattern.match(matchFrom)) return [pattern.replace(matchFrom, '$1')];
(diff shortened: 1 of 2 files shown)


webpro:
Thanks @mcous! Very much appreciate you made the leap from "dang this thing doesn't work at all" to a PR with a fix with tests! Will be released shortly.

webpro:
:rocket: _This pull request is included in v5.31.0. See [Release 5.31.0](https://github.com/webpro-nl/knip/releases/tag/5.31.0) for release notes._

PR #823, opened 2024-10-29
[feat] Add `--ignore-exports-used-in-file` CLI argument

Sheraff:
This PR adds a new CLI argument `--ignore-exports-used-in-file` that mirrors the config option of the same name (https://knip.dev/reference/configuration#ignoreexportsusedinfile)


webpro:
Thanks for your PR, @Sheraff. We've [discussed this on Discord](https://discord.com/channels/1143209786612125696/1143209787140620310/1300892378747506840), and I don't think we're going to merge this. Because the option in the configuration file also supports a more fine-grained object, this may lead to confusion. Also there seems to be no further interest in this option.

PR #827, opened 2024-11-01
Skip indexing into a gh action if its not an array

butlerx:
Skip checking a github action step if its not an array as its possibly a templated string.
In gh actions a valid config is `${{ toJSON(steps) }}` which breaks the plugins parsing. if we see a string we should just skip those steps


webpro:
Thanks @butlerx! Should be out shortly.

webpro:
:rocket: _This pull request is included in v5.36.1. See [Release 5.36.1](https://github.com/webpro-nl/knip/releases/tag/5.36.1) for release notes._

PR #832, opened 2024-11-12
Support caching results on Win32-based systems

dbehmoaras:
…ct creation


dbehmoaras:
> Thanks for your PR, @dbehmoaras! This is the central function to normalize paths to posix format. So I'm not sure whether this is the right location to fix the issue. What's the exact issue you're trying to solve here?

Hmm - I see your point. Maybe then it should be addressed in file that handles the caching of the results ? The issue we run into is that Knip fails to update the cache on windows systems because it resolves a posix file path when windows env expects a win32 path from node.js

webpro:
Alright, clear.

Maybe you can `import path from 'node:path` in `file-entry-cache.ts` and take it from there? Fix `this.filePath` with `path.resolve`, etc.

dbehmoaras:
Ok let me try that

webpro:
Great, I'm asking so you can directly test it works properly on Windows. If you create a PR, you will automatically get a version that you can `npm install` anywhere, if that helps.

dbehmoaras:
Apologies for the delay on this. My schedule get away from me. Will try to update this over the next couple of days. Thank you for your patience!

dbehmoaras:
@webpro PR is updated and ready for your feedback. Confirmed that it works on posix systems (used Mac OSX 15.1 to confirm). Still running some smoke tests on PC. Will update when complete.

packages/knip/src/util/file-entry-cache.ts
```diff
@@ -38,7 +40,19 @@ export class FileEntryCache<T> {
...
+     * applies correct path transformation based on the current environment
+     * and applies it to the cache location passed in from the cli args
+     */
+    const isWindows = process.platform === 'win32';
+    const { directory: directoryFromCliArg } = parsedArgValues;
+    const currentCwd = directoryFromCliArg || cwd;
+    const currentPath = isWindows ? path.win32.resolve(currentCwd): path.posix.resolve(currentCwd);
+    const pathResolver =  isWindows ? path.win32.join : resolve;
+
+    this.filePath = isAbsolute(currentPath) ? pathResolver(currentPath, cacheId) : pathResolver(currentPath, _path, cacheId);
```
webpro, line 54: - `cwd` should already `[be --directory](https://github.com/webpro-nl/knip/blob/3ea698db8386e994c8df5580c1aea65d21ca39f2/packages/knip/src/util/path.ts#L19) so i don't think we need to read `directory` here from `parsedArgValues` again
- we can use `path.resolve` which should internally do the platform switch, no?
- same for `path.join.

So, thinking out loud, would only this do the same?

```ts
    const currentPath = path.resolve(cwd);
    this.filePath = isAbsolute(currentPath) ? path.join(currentPath, cacheId) : path.join(currentPath, _path, cacheId);
```

dbehmoaras: > * `cwd` should already [be --directory](https://github.com/webpro-nl/knip/blob/3ea698db8386e994c8df5580c1aea65d21ca39f2/packages/knip/src/util/path.ts#L19) so i don't think we need to read `directory` here from `parsedArgValues` again
> * we can use `path.resolve` which should internally do the platform switch, no?
> * same for `path.join`
> 
> So, thinking out loud, would only this do the same?
> 
> ```ts
>     const currentPath = path.resolve(cwd);
>     this.filePath = isAbsolute(currentPath) ? path.join(currentPath, cacheId) : path.join(currentPath, _path, cacheId);
> ```

I will check on those things. We could probably simplify the approach to resolving the path - I'd like to verify that we only need the generic path resolver and not win32 specific one. Regarding directory - I meant to use `cache-location` instead. I will fix this in my next commit.

webpro: Thanks! I think `cache-location` is the `_path` we're receiving as the second argument.

I'm seeing now I made a mistake, here's the correction (did not actually test):

```ts
const currentPath = path.resolve(cwd);
this.filePath = isAbsolute(_path) ? path.join(_path, cacheId) : path.join(currentPath, _path, cacheId);
```

dbehmoaras:
@webpro I think the change is finished from a functional standpoint. `format` fails due to [this biome rule](https://github.com/webpro-nl/knip/blob/ea478f5fc3ad79c916f8302d0e7d45b94892dcee/biome.json#L32). In principle I agree with the rule. Importing [`join`](https://github.com/webpro-nl/knip/blob/ea478f5fc3ad79c916f8302d0e7d45b94892dcee/packages/knip/src/util/path.ts#L15) and [`resolve`](https://github.com/webpro-nl/knip/blob/ea478f5fc3ad79c916f8302d0e7d45b94892dcee/packages/knip/src/util/path.ts#L21), so we would need to expose an env agnostic (or win32) version of each of those functions in that file.

Thoughts on `joinEnvAgnostic` and `resolveEnvAgnostic` ? I'm afraid if we change the current exports it may have unintended side effects.

Alternatively we can add the `// biome-ignore lint/nursery/noRestrictedImports: ignore` rule to this file. Let me know which you prefer.

webpro:
> Alternatively we can add the `// biome-ignore lint/nursery/noRestrictedImports: ignore` rule to this file. Let me know which you prefer.

Sorry I missed that. Please ignore the violation, we should use `node:path` here.

dbehmoaras:
Sure thing - just pushed the latest with the rule ignored.

webpro:
:rocket: _This pull request is included in v5.41.1. See [Release 5.41.1](https://github.com/webpro-nl/knip/releases/tag/5.41.1) for release notes._

webpro:
Thanks again, @dbehmoaras!

dbehmoaras:
Of course! Happy to contribute

PR #861, opened 2024-11-29
docs: update entry/project defaults with production mode

davidlj95:
Defaults that appear in docs are not exactly the [defaults used](https://github.com/webpro-nl/knip/blob/5.38.2/packages/knip/src/ConfigurationChief.ts#L34-L40).

`projects` defaults in docs do not include the final `!` . Hence readers could understand all files are not production files by default.

Updates defaults found in docs to reflect the actual defaults used by the implementation.

> Also sorts `index,cli,main` to match implementation to be nitty :P

Btw, didn't find any place in docs saying that all project files are considered production files by default. Should that be added somewhere? Or it's this new defaults update enough maybe?

### DIFF
--- a/packages/docs/src/content/docs/explanations/entry-files.md
+++ b/packages/docs/src/content/docs/explanations/entry-files.md
@@ -25,10 +25,10 @@ entry files. Here's the default configuration in full:
 ```json
 {
   "entry": [
-    "{index,main,cli}.{js,cjs,mjs,jsx,ts,cts,mts,tsx}",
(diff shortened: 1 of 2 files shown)


webpro:
Thanks, David! Fully on board with the changes.

> Btw, didn't find any place in docs saying that all project files are considered production files by default. Should that be added somewhere? Or it's this new defaults update enough maybe?

Good point, I'll add it to the production mode page shortly.

webpro:
:rocket: _This pull request is included in v5.38.4. See [Release 5.38.4](https://github.com/webpro-nl/knip/releases/tag/5.38.4) for release notes._

PR #865, opened 2024-12-01
Track Angular's `server` builder option

davidlj95:
Tracks Angular's server side rendering server entrypoint file as a production file. 

Defined in Angular CLI's `@angular-devkit/build-angular:application` builder [`server` option](https://github.com/angular/angular-cli/blob/19.0.2/packages/angular/build/src/builders/application/schema.json#L19)

> [!NOTE]
> This is the first of a series of PRs to improve the Angular plugin so that it works out-of-the-box for freshly baked Angular apps.
> 
> Checkout the analysis and following steps in https://github.com/davidlj95/a19-knip
>
> I'll update types in the next PR, as seems many types have changed with recent updates.


webpro:
Thanks, David!

PR #866, opened 2024-12-01
Update Angular workspace types and add update script

davidlj95:
Seems Angular CLI workspace file `angular.json` types has been updated with some additions around. This adds the new types using the same `json-schema-to-typescript` approach. 

Adds a small shell script to help updating that file in the future. Let me know if JS is preffered. I'm more used to writing shell scripts, but can give try writing one using Node.js / `bun`'s compatible `child_process.exec`

### DIFF
--- a/packages/knip/src/plugins/angular/update-types.sh
+++ b/packages/knip/src/plugins/angular/update-types.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env sh
 set -eu
-angular_plugin_dir="$(cd "$(dirname "$0")" && pwd)"
+root_dir="$(git rev-parse --show-toplevel)"
+angular_plugin_dir="$root_dir/packages/knip/src/plugins/angular"
 tmp_dir="$(mktemp -d)"
 cleanup() {
   rm  -rf "$tmp_dir"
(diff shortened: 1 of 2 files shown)


webpro:
Fantastic! Tried the Bash script, flawless. We could rewrite to TS, but personally I really don't mind Bash. LFG.

packages/knip/src/plugins/angular/update-types.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
+set -eu
+angular_plugin_dir="$(cd "$(dirname "$0")" && pwd)"
```
webpro, line 3:
```suggestion
root_dir="$(git rev-parse --show-toplevel)"
angular_plugin_dir="$root_dir/packages/knip/src/plugins/angular"
```

packages/knip/src/plugins/angular/update-types.sh
```diff
@@ -0,0 +1,12 @@
...
+tmp_dir="$(mktemp -d)"
+cleanup() {
+  rm  -rf "$tmp_dir"
+}
+trap cleanup EXIT
+git clone --depth 1 https://github.com/angular/angular-cli "$tmp_dir"
+cd "$tmp_dir/packages/angular/cli/lib/config"
+bunx json-schema-to-typescript -i workspace-schema.json -o "$angular_plugin_dir/types.ts";
```
webpro, line 11: Knip reported unused exports. We have a few options here:

- Add as entry file (meh)
- Ignore in root `knip.json` (fine, but doesn't look super great in knip repo itself)
- Fix up unused exports using knip:

```suggestion
bunx json-schema-to-typescript -i workspace-schema.json -o "$angular_plugin_dir/types.ts";
$root_dir/packages/knip/bin/knip-bun.js --directory "$root_dir" --fix
```

webpro (commented):
Noticed Knip itself complains about unused exports 😜 so added suggestions how we could deal with it, other ideas also welcome of course

packages/knip/src/plugins/angular/update-types.sh
davidlj95: Noice! Didn't know that git magic command ✨

webpro:
Thanks a lot David, very grateful for this 🙏

PR #868, opened 2024-12-02
Track Angular environment files as entry points

davidlj95:
Tracks Angular [environment specific files](https://angular.dev/tools/cli/environments#using-environment-specific-variables-in-your-app) as entry points.

Takes into account:
 - If file replacements are part of a production configuration or not to add them as production entries or not.
 - Previous syntax for Webpack-based browser application builder (`src/replaceWith` instead of `replace/with`).
 - File replacements out of `configurations`. As they're technically correct and work. However, it's not a common use case. The purpose of environment files is to apply different ones depending on the configuration, not apply them for all cases (hence part of the builder options). Also those aren't used if a configuration specifies some `fileReplacements`.

### DIFF
--- a/packages/knip/fixtures/plugins/angular2/angular.json
+++ b/packages/knip/fixtures/plugins/angular2/angular.json
@@ -60,7 +60,13 @@
             "development": {
               "optimization": false,
               "extractLicenses": false,
-              "sourceMap": true
+              "sourceMap": true,
+              "fileReplacements": [
+                {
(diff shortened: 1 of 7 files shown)


webpro:
You're not leaving much on the table to complain about, this is great

webpro:
:rocket: _This pull request is included in v5.39.0. See [Release 5.39.0](https://github.com/webpro-nl/knip/releases/tag/5.39.0) for release notes._

PR #871, opened 2024-12-06
Add Karma plugin

davidlj95:
Add [Karma](https://karma-runner.github.io/) plugin

Features:
 - Track frameworks from configuration file as dev dependencies 891b17083cc211efa59afa69c70cd49d3b9e8149
   - Takes into account that in the case of Jasmine, just `jasmine-core` is needed. `jasmine` package is the CLI package. For instance, Angular projects do not install Jasmine's CLI `jasmine` package, given `ng test` is used instead. This can be [clearly seen in `karma-jasmine` plugin code](https://github.com/karma-runner/karma-jasmine/blob/v5.1.0/lib/index.js#L8)
 - Track test files from configuration file as non-production entry points by reading `basePath` / `files` 3849e47195b47da801d85e46f3b97f7e2748473d
   - ~Not taking into account `exclude` for now. Could be improved later. Jest plugin doesn't take ignore configurations either~
   - Tracks excluded test files as negated entry patterns by reading `exclude` 5c9090a4f0bf4c69ffc76d7d47a93e4378d10933
 - Track dev dependencies specified in `plugins` in the configuration file c025a8959d33ca2ee24b7f508aa633d25d022ae8 24a94412b4ac0f1827e96cae4ae91f99b9c1bbcf
   - Supports default of loading all `karma-*` (dev) dependencies as plugins
   - Supports local plugins (requiring JS files, with or without extension) 
   - Doesn't support requiring dependencies in the config file (`require('karma-coverage')` instead of `karma-coverage`) 

Could be improved:
 - `InMemoryConfig.set` to merge instead of just store the given config. Configuration files may use the `set` method multiple times to merge multiple configurations. However, the (subjectively) usual way is to just call `config.set` with the options to set.
 - Find calls to `karma` binary to find plugins / files used via CLI arguments

Out of scope:
 - Reading CoffeeScript `.coffee` configuration files (or [LiveScript ones](https://github.com/karma-runner/karma/blob/v6.4.4/lib/config.js#L442-L443))
 - Parsing `middlewares`, `beforeMiddleware` and `reporters` to avoid taking into account unused plugins

### DIFF
--- a/packages/knip/src/plugins/karma/index.ts
+++ b/packages/knip/src/plugins/karma/index.ts
@@ -64,6 +64,11 @@ const resolveEntryPaths: ResolveEntryPaths<ConfigFile> = (configFile, options) =
       inputs.add(toEntry(join(options.configFileDir, basePath, fileOrPattern)));
     }
   }
+  if (config.exclude) {
+    for (const fileOrPattern of config.exclude) {
+      inputs.add(toEntry(`!${join(options.configFileDir, basePath, fileOrPattern)}`));
+    }
+  }
 
   return Array.from(inputs);
 };
--- a/packages/knip/src/plugins/karma/index.ts
+++ b/packages/knip/src/plugins/karma/index.ts
@@ -1,6 +1,6 @@
 import type { IsPluginEnabled, Plugin, ResolveConfig, ResolveEntryPaths } from '../../types/config.js';
 import { type Input, toDevDependency, toEntry } from '../../util/input.js';
-import { join } from '../../util/path.js';
+import { extname, isInternal, join } from '../../util/path.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { Config, ConfigOptions } from './types.js';
 
@@ -19,7 +19,7 @@ const entry: string[] = [];
 
 type ConfigFile = (config: Config) => void;
 
-const resolveConfig: ResolveConfig<ConfigFile> = async configFile => {
+const resolveConfig: ResolveConfig<ConfigFile> = async (configFile, options) => {
   const inputs = new Set<Input>();
 
   const config = loadConfigFromFile(configFile);
@@ -27,6 +27,23 @@ const resolveConfig: ResolveConfig<ConfigFile> = async configFile => {
   if (config.frameworks) {
     for (const framework of config.frameworks) {
       inputs.add(toDevDependency(devDepForFramework(framework)));
+    }
+  }
+  if (config.plugins) {
+    for (const plugin of config.plugins) {
+      if (typeof plugin !== 'string') continue;
+      if (isInternal(plugin)) {
+        inputs.add(toEntry(extname(plugin).length === 0 ? `${plugin}.js` : plugin));
+      } else {
+        inputs.add(toDevDependency(plugin));
+      }
+    }
+  } else {
+    const karmaPluginDevDeps = Object.keys(options.manifest.devDependencies ?? {}).filter(name =>
+      name.startsWith('karma-')
+    );
+    for (const karmaPluginDevDep of karmaPluginDevDeps) {
+      inputs.add(toDevDependency(karmaPluginDevDep));
     }
   }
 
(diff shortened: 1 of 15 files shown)


packages/knip/src/plugins/karma/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.config = config;
+  }
+}
+
+const resolveConfig: ResolveConfig<ConfigFile> = async configFile => {
+  const inputs = new Set<Input>();
+
+  const inMemoryConfig = new InMemoryConfig();
+  configFile(inMemoryConfig);
+  const { config } = inMemoryConfig;
```
davidlj95, line 37: A weird way of getting config, which matches the weirdness of the configuration file: a file that exports a function that takes a configuration object which has a `set` method to merge configurations and expects you to call it to set your configuration. So you need to call the exported function with the base configuration where the user will call the `set` method with some configuration options.

Checkout [Karma config file parsing](https://github.com/karma-runner/karma/blob/v6.4.4/lib/config.js#L468-L473)

webpro:
Thanks @davidlj95, looks great.

On board with the comments re. scope of this PR. In general, I'm fine with "unfinished" plugins. All plugins are essentially unfinished as tools and configs change over time. Everything that results in less false positives is a win (without being overly "expensive"/slow or hard to maintain).

packages/knip/src/plugins/karma/index.ts
webpro: Yeah sometimes it gets a bit weird just to not depend on the actual (often large) tool. Thanks for linking it up in the source, it'll def help in the future.

davidlj95:
> Thanks @davidlj95, looks great.

Thanks! 😊 

> 
> On board with the comments re. scope of this PR. In general, I'm fine with "unfinished" plugins. All plugins are essentially unfinished as tools and configs change over time. Everything that results in less false positives is a win (without being overly "expensive"/slow or hard to maintain).

Nice to hear that! I'll do the minimum so the plugin's useful: test files as non production entries. Without taking `exclude` into account for now, as just checked and Jest plugin doesn't take it into account either. And dependencies from `plugins`. Can always be improved later. Thanks for the early review 🙇

packages/knip/src/plugins/karma/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (config.frameworks) {
+    for (const framework of config.frameworks) {
+      inputs.add(toDevDependency(devDepForFramework(framework)));
+    }
+  }
+  if (config.plugins) {
+    for (const plugin of config.plugins) {
+      if (typeof plugin !== 'string') continue;
+      if (isInternal(plugin)) {
+        inputs.add(toEntry(extname(plugin).length === 0 ? `${plugin}.js` : plugin));
```
webpro, line 36: Usually it's better to either:

- if we're sure it's internal and allow it to end up as "unresolved": use `toDeferResolveEntry` - this is more precise/performant, especially combined with `toDevDependency`
- otherwise, use `toDeferResolve` for both internal/external `plugin` strings - this is what most plugins do

Both will defer resolution to later in the process to an internal `enhanced-resolve` wrapper, taking care of file extensions and other module resolution gotchas.

webpro (commented):
> Nice to hear that! I'll do the minimum so the plugin's useful: test files as non production entries. Without taking `exclude` into account for now, as just checked and Jest plugin doesn't take it into account either. And dependencies from `plugins`. Can always be improved later. Thanks for the early review 🙇

Great stuff, really liking the implementation.

Maybe later on in another PR we could add `exclude` as negated `toEntry` patterns.

davidlj95:
> Great stuff, really liking the implementation.
Thx 🙇 😊

> 
> Maybe later on in another PR we could add `exclude` as negated `toEntry` patterns.

Done in latest commit! 🎉

packages/knip/src/plugins/karma/index.ts
davidlj95: Thanks! Went for `toDeferResolveEntry` / `toDevDependency`. And removed the `extname` trick thingie. I came across those but didn't fully understand them so that's why I used the basic ones. Resolving.

webpro:
And another excellent plugin from David, thank you! 🙏

webpro:
:rocket: _This pull request is included in v5.40.0. See [Release 5.40.0](https://github.com/webpro-nl/knip/releases/tag/5.40.0) for release notes._

PR #879, opened 2024-12-15
Add `expo` plugin

jjjjonathan:
This PR adds a plugin for [Expo](https://docs.expo.dev/).

Features
- Adds entry files based on whether `expo-router` (A file-based router akin to NextJS) is used or not.
- ~~Adds `expo`, `react`, and `react-native` as dependencies; required for all Expo projects.~~
- Adds dependencies from plugins listed in Expo config file
- Adds dependencies that are implied from other config options
- Adds dependencies from certain packages that only need to be installed to be picked up by Expo. For example, installing `expo-dev-client` will cause Expo to use the package when building the native app; it does not need to referenced or imported anywhere other than the `package.json`.

### DIFF
--- a/packages/knip/src/plugins/expo/helpers.ts
+++ b/packages/knip/src/plugins/expo/helpers.ts
@@ -1,0 +1,81 @@
+import type { PluginOptions } from '../../types/config.js';
+import { type Input, toDependency, toProductionDependency } from '../../util/input.js';
+import { getPackageNameFromModuleSpecifier } from '../../util/modules.js';
+import type { ExpoConfig } from './types.js';
+
+// https://docs.expo.dev/versions/latest/config/app
+
+export const getDependencies = async (expoConfig: ExpoConfig, { manifest }: PluginOptions) => {
+  const config = 'expo' in expoConfig ? expoConfig.expo : expoConfig;
+
+  const platforms = config.platforms ?? ['ios', 'android'];
+
+  const pluginPackages =
+    (config.plugins
+      ?.map(plugin => {
+        const pluginName = Array.isArray(plugin) ? plugin[0] : plugin;
+        return getPackageNameFromModuleSpecifier(pluginName);
+      })
+      .filter(Boolean) as string[]) ?? [];
+
+  const inputs = new Set<Input>([
+    ...['expo', 'react-native', 'react'].map(toProductionDependency),
+    ...pluginPackages.map(toDependency),
+  ]);
+
+  const allowedPackages = ['expo-atlas', 'expo-dev-client'];
+  const allowedProductionPackages = ['expo-insights'];
+
+  const manifestDependencies = Object.keys(manifest.dependencies ?? {});
+
+  for (const pkg of allowedPackages) {
+    if (manifestDependencies.includes(pkg)) {
+      inputs.add(toDependency(pkg));
+    }
+  }
+
+  for (const pkg of allowedProductionPackages) {
+    if (manifestDependencies.includes(pkg)) {
+      inputs.add(toProductionDependency(pkg));
+    }
+  }
+
+  if (config.updates?.enabled !== false) {
+    inputs.add(toProductionDependency('expo-updates'));
+  }
+
+  if (config.notification) {
+    inputs.add(toProductionDependency('expo-notifications'));
+  }
+
+  const isExpoRouter = manifest.main === 'expo-router/entry';
+
+  // https://docs.expo.dev/router/installation/#setup-entry-point
+  if (isExpoRouter) {
+    inputs.add(toProductionDependency('expo-router'));
+  }
+
+  // https://docs.expo.dev/workflow/web/#install-web-dependencies
+  if (platforms.includes('web')) {
+    inputs.add(toProductionDependency('react-native-web'));
+    inputs.add(toProductionDependency('react-dom'));
+
+    // https://github.com/expo/expo/tree/main/packages/@expo/metro-runtime
+    if (!isExpoRouter) {
+      inputs.add(toDependency('@expo/metro-runtime'));
+    }
+  }
+
+  if (
+    (platforms.includes('android') && (config.userInterfaceStyle || config.android?.userInterfaceStyle)) ||
+    (platforms.includes('ios') && (config.backgroundColor || config.ios?.backgroundColor))
+  ) {
+    inputs.add(toProductionDependency('expo-system-ui'));
+  }
+
+  if (platforms.includes('android') && config.androidNavigationBar) {
+    inputs.add(toProductionDependency('expo-navigation-bar'));
+  }
+
+  return [...inputs];
+};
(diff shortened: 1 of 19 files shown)


packages/knip/src/plugins/expo/helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+  const pluginPackages =
+    (config.plugins
+      ?.map(plugin => {
+        const pluginName = Array.isArray(plugin) ? plugin[0] : plugin;
+        return getPackageNameFromModuleSpecifier(pluginName);
+      })
+      .filter(Boolean) as string[]) ?? [];
+
+  const inputs = new Set<Input>([
+    ...['expo', 'react-native', 'react'].map(toProductionDependency),
```
webpro, line 22: > Adds `expo`, `react`, and `react-native` as dependencies; required for all Expo projects.

Aren't those dependencies referenced in regular ways already? I'd think it isn't necessary to add them explicitly. It's a bit far-fetched, but in theory this might change, and if that's what they are, peer dependencies should picked up properly already.

packages/knip/src/plugins/expo/index.ts
```diff
@@ -0,0 +1,52 @@
...
+const config: string[] = ['app.json', 'app.config.{ts,js}'];
+
+const resolveEntryPaths: ResolveEntryPaths<ExpoConfig> = async (expoConfig, { manifest }) => {
+  const config = 'expo' in expoConfig ? expoConfig.expo : expoConfig;
+
+  let production: string[] = ['index.{js,jsx,ts,tsx}'];
+
+  // https://docs.expo.dev/router/installation/#setup-entry-point
+  if (manifest.main === 'expo-router/entry') {
+    production = ['app/**/*.{js,jsx,ts,tsx}', 'src/app/**/*.{js,jsx,ts,tsx}'];
```
webpro, line 25: Why does this `manifest.main` value result in all those files becoming entry files, isn't that pattern too broad?

packages/knip/src/plugins/expo/helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+    (platforms.includes('ios') && (config.backgroundColor || config.ios?.backgroundColor))
+  ) {
+    inputs.add(toProductionDependency('expo-system-ui'));
+  }
+
+  if (platforms.includes('android') && config.androidNavigationBar) {
+    inputs.add(toProductionDependency('expo-navigation-bar'));
+  }
+
+  return [...inputs];
```
webpro, line 80: That looks like a lot of magic Expo is pulling off, thanks a lot for figuring that all out here. Also appreciate you've added links to docs and used some "heuristics" to best-guess including things like `expo-system-ui`.

webpro (commented):
Thanks, Jonathan! Great PR, happy to merge. A few remarks I have.

packages/knip/src/plugins/expo/helpers.ts
jjjjonathan: Ahh I didn't realize Knip picked up peer dependencies. Very cool! That takes care of `react` and `react-native`, and on second thought `expo` will almost certainly be in the package.json scripts (if not imported in the project), so I'll remove that as well.

packages/knip/src/plugins/expo/index.ts
jjjjonathan: Setting `main` to `expo-router/entry` essentially "turns on" expo-router (if the package is installed, of course). It's a file-system based router, so every file in the `app` directory becomes it own route. (e.g. an `app/profile/settings.tsx` file will become a `/profile/settings` URL) This is all done under the hood by expo-router; these route files are not imported elsewhere in the project source code. So I do think the broad pattern is warranted here, you can name a file anything you want, and expo router will pick it up. It's the same concept as the NextJS "pages" router; anything in that directory is an entry file. https://github.com/webpro-nl/knip/blob/9ce4a49ab09b3e522fa48996f7f06cf44baee1c6/packages/knip/src/plugins/next/index.ts#L22

jjjjonathan: Side note, the `app` directory can live either at the root or under `src`, and `src/app` takes precedence. (https://docs.expo.dev/router/reference/src-directory/) Is it worth it to check for the existence of `src/app` and fall back to `app` if it doesn't exist, instead of using both patterns at the same time?

jjjjonathan: Sounds good, I'll keep it simple

webpro:
Excellent stuff, Jonathan! Really appreciate this, and I'm sure others will as well.

webpro:
:rocket: _This pull request is included in v5.41.0. See [Release 5.41.0](https://github.com/webpro-nl/knip/releases/tag/5.41.0) for release notes._

jjjjonathan:
Thanks for the awesome project, it's tremendously useful! There will be more plugin PRs from me :)

packages/knip/src/plugins/expo/index.ts
wcastand: i also use expo-router but in a index.ts file as my entry and it seems that break that plugin, i can't get it to no pu everything as unused. maybe i configure it badly.

webpro: Thanks @wcastand, here as a comment it might get lost, any chance you could give more details so we have something to look into? What's the issue you're seeing and how to reproduce it?

jerone:
@jjjjonathan Thanks for this plugin! It allowed me to remove a lot of whitelisted dependencies.

I do have a few left, that might still be relevant:
* `babel-preset-expo`
* `expo-updates`
* `react-dom`
* And if your project uses Jest:
  * `babel-jest`
  * `jest-watch-select-projects`
  * `jest-watch-typeahead`

jjjjonathan:
@jerone Glad you're getting some use out of it! Here are my thoughts/suggestions:

- `expo-updates` should already be picked up by this plugin, unless you have `updates: { enabled: false } }` in your Expo config file
- `react-dom` should already be picked up by this plugin, make sure the `platforms` array includes `web` in your Expo config file
- `babel-preset-expo` and the Jest dependencies should be picked up by the Babel/Jest plugins. I'd `knip --debug` and make sure those plugins are active

packages/knip/src/plugins/expo/index.ts
wcastand: i changed the way i handle the entry point so seems to be good for now, issue i can see seems to be with custom modules in `modules` directory being flagged at unused.
other than that hard to say what's an issue on our repo or not so will take a deeper look and if i find anything i will get back to you.

great tool either way thanks :)

arthur-fontaine:
thanks for the plugin @jjjjonathan 

I just set up Knip in my Expo project, but I still get a report for `Unused files (1): metro.config.js`. Is it an expected behavior?

arthur-fontaine:
Nice thank you!

PR #885, opened 2024-12-20
Add Karma configuration from Angular plugin configuration

davidlj95:
Angular allows running tests using Karma via the [`@angular-devkit/build_angular:karma` builder](https://github.com/angular/angular-cli/tree/19.0.6/packages/angular_devkit/build_angular/src/builders/karma). However, configuration in Angular CLI workspace (aka `angular.json`) using that builder to run tests using Karma isn't taken into account as Karma configuration. But this is something that comes built-in when creating a fresh Angular project. This PR aims to take into account that builder configuration for `knip`.

To do so:
 - **Karma plugin has been refactored**: so that common functionality shared by both plugins lives in `helpers.ts`. In the same fashion that XO uses some ESLint helpers. The code to load a configuration file has been moved there too to keep clean the `index.ts` file.
 - **Includes and excludes** specified as builder options are used as included/excluded non-production entry points. If unspecified, Angular CLI defaults are used. Which actually come from the JSON Schema. Will try and see if can the script can be updated to obtain those values too.
 - **Karma config** file specified as `karmaConfig` in builder options is used to:
   - **Load default Karma config** if not specified
   - **Load an alternative Karma config file** if specified, but not one that Karma plugin would take by default.

### DIFF
--- a/packages/knip/src/plugins/angular/index.ts
+++ b/packages/knip/src/plugins/angular/index.ts
@@ -2,7 +2,8 @@ import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.
 import { type Input, toConfig, toDependency, toEntry, toProductionEntry } from '../../util/input.js';
 import { join } from '../../util/path.js';
 import { hasDependency } from '../../util/plugin.js';
-import type { AngularCLIWorkspaceConfiguration, WebpackBrowserSchemaForBuildFacade } from './types.js';
+import * as karma from '../karma/helpers.js';
+import type { AngularCLIWorkspaceConfiguration, KarmaTarget, WebpackBrowserSchemaForBuildFacade } from './types.js';
 
 // https://angular.io/guide/workspace-config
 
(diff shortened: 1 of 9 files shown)


packages/knip/src/plugins/angular/index.ts
```diff
@@ -61,6 +62,34 @@ const resolveConfig: ResolveConfig<AngularCLIWorkspaceConfiguration> = async (co
...
+        const karmaConfig = karmaBuilderOptions.karmaConfig;
+        if (!karmaConfig) {
+          // Hardcoded default Karma config from Angular builder
+          // https://github.com/angular/angular-cli/blob/19.0.6/packages/angular_devkit/build_angular/src/builders/karma/index.ts#L115
+          karma
+            .inputsFromPlugins(
+              ['karma-jasmine', 'karma-chrome-launcher', 'karma-jasmine-html-reporter', 'karma-coverage'],
+              options.manifest.devDependencies
+            )
+            .forEach(inputs.add, inputs);
```
davidlj95, line 86: Don't like the `.forEach(inputs.add, inputs)`. But [doesn't work](https://stackoverflow.com/a/50882116/3263250) without the latest `inputs` argument to specify `this`. And a for loop (as `forEach` is forbidden by Biome linter) looks too much 🤷  happy to change if someone else is unhappy too

webpro: Agreed, this is fine really

webpro:
Great, plugins use each other all over the place and it's chaotic and beautiful :)

webpro:

PR #886, opened 2024-12-20
Track Angular scripts build option

davidlj95:
Several Angular CLI builders allow to specify a [`scripts` option](https://github.com/angular/angular-cli/blob/19.0.6/packages/angular/build/src/builders/application/schema.json#L78) in its configuration. Those files will be included in the build and optionally added to `index.html`. Track them as production entries\*.

> \* Though they may not be production entries! They'll probably be production entries if they are part of the `build` target. 
> Same happens for other configuration options. Like `main`, which can be specified in many `build` targets but also on non-production `test` target. 
> Options `browser`, `ssr.entry` or `server` could also be specified in other targets other than `build` and they won't probably be production entries. Opening another PR for that

### DIFF
--- a/packages/knip/fixtures/plugins/angular/angular.json
+++ b/packages/knip/fixtures/plugins/angular/angular.json
@@ -32,7 +32,9 @@
             "styles": [
               "src/styles.scss"
             ],
-            "scripts": []
+            "scripts": [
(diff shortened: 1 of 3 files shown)


webpro:

PR #887, opened 2024-12-20
Separate Angular production vs non-production entries

davidlj95:
> ~~This is a work in progress. #886 should be merged first to take that into account too. #884 should be merged first too in order for production mode to run.~~ ✅ 

Some of the entry files added as part of Angular plugin are added as production entries. However they may not be production entries. Hence causing production mode to not work as expected.

Configuration for building an Angular project is usually specified under the `build` target inside the `architect` object of the Angular CLI workspace (aka `angular.json`). However, right now the implementation is using configuration options from all targets. Which is good to capture also build options like when building for testing within the `test` target. However, in that scenario, the entries added should be non-production ones.

Also, in any architect target options, the options defined in there are the default values for that target. But they may be overridden when specifying [one or more alternative configurations](https://angular.dev/reference/configs/workspace-config#build-target). From those, the `production` configuration will contain production options and the rest are very probably non-production ones.

Here are the options from which we're adding entries and the Angular CLI's builders they can be used with

| Option | [Webpack](https://github.com/angular/angular-cli/blob/19.0.6/packages/angular_devkit/build_angular/src/builders/browser/schema.json#) | [Karma (test)](https://github.com/angular/angular-cli/blob/19.0.6/packages/angular_devkit/build_angular/src/builders/karma/schema.json) | [ESBuild browser](https://github.com/angular/angular-cli/blob/19.0.6/packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json) | [ESBuild app (`@angular-devkit`)](https://github.com/angular/angular-cli/blob/17.3.11/packages/angular_devkit/build_angular/src/builders/application/schema.json) / [ESBuild app (`@angular/build`)](https://github.com/angular/angular-cli/blob/19.0.6/packages/angular/build/src/builders/application/schema.json) |
|--:|:-:|:-:|:-:|:-:|
| `main` | X | X | | |
| `scripts` | X | X | X | X |
| `fileReplacements` | X | X | X | X |
| `browser` | | | | X |
| `server` | | | | X |
| `ssr.entry` | | | | X |

This PR does two main things:

1. Refactors to read entry files/patterns from both default options and also from each of the configurations defined. This is an additional feature, as previously entries defined in configurations weren't read (apart from `fileReplacement` ones)
2. Add as production entries only the entries for the `build` target in `production` configuration. Takes into account that if an option isn't defined in the `production` configuration (if any), the default option will be used instead. Rest of entries will be development ones

Also the order of options has been changed so that they appear more or less chronologically ascendent. For instance, `main` option comes from the classic / old Webpack builder so appears first than newer `browser` option in application builder.

### DIFF
--- a/packages/knip/src/plugins/angular/index.ts
+++ b/packages/knip/src/plugins/angular/index.ts
@@ -3,7 +3,12 @@ import { type Input, toConfig, toDependency, toEntry, toProductionEntry } from '
 import { join } from '../../util/path.js';
 import { hasDependency } from '../../util/plugin.js';
 import * as karma from '../karma/helpers.js';
-import type { AngularCLIWorkspaceConfiguration, KarmaTarget, WebpackBrowserSchemaForBuildFacade } from './types.js';
+import type {
+  AngularCLIWorkspaceConfiguration,
+  KarmaTarget,
+  Project,
+  WebpackBrowserSchemaForBuildFacade,
+} from './types.js';
 
 // https://angular.io/guide/workspace-config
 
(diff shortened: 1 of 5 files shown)


packages/knip/src/plugins/angular/index.ts
```diff
@@ -23,31 +23,33 @@ const resolveConfig: ResolveConfig<AngularCLIWorkspaceConfiguration> = async (co
 
   for (const project of Object.values(config.projects)) {
     if (!project.architect) return [];
-    for (const target of Object.values(project.architect)) {
+    for (const [targetName, target] of Object.entries(project.architect)) {
+      const isProductionTarget = targetName === 'build';
```
webpro, line 27: I'm totally on board with this. Maybe move this to a function to make it more explicit and then we could pass it e.g. both of `[targetName, target]` to have any sort of heuristics to decide "prod or not".

davidlj95: Nice to hear!  Thanks for the early feedback, that's encouraging 💪 ☺️

Eventually implemented the improvement of taking into account the several  `configurations` a target may have. So code has changed and there's now 2 places where the conditions for production are different. Let me know if can be improved with the final implementation.

packages/knip/test/plugins/angular3.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  assert(nonProdIssues.devDependencies['package.json']['@angular/cli']);
+
+  assert.deepEqual(nonProdCounters, {
+    ...baseCounters,
+    devDependencies: 1,
+    processed: 8,
+    total: 8,
+  });
+
+  const { counters: prodCounters } = await main({
```
webpro, line 25: Any chance this could be a separate tests? Even if same title + `(2)`.

webpro (commented):
Thanks a bunch, David. This is all great, just a nitpick left.

webpro:
Just went ahead and merged to prepare for the next release. Thanks a bunch again, David!

packages/knip/test/plugins/angular3.test.ts
davidlj95: Just saw you did it already 💪  Thanks!

webpro:

Issue #126, opened 2023-05-16
Platform-specific extensions​


neiker:
Hey! I'm using **knip** with **react-native** and after some adjustments it works wonderfully except for one detail. I can't find how to make files with platform-specific extensions​ be seen as the same files by **knip**. 

Eg. I have a file `Component.tsx` and in the same path a `Component.android.tsx`, the first is detected as used but the second doesn't, when in fact they should be considered "the same file". 

As a workaround, I added `"**/*.android.ts{,x}"` on `ignore` but in case `Component.tsx` is removed someone might forget to remove `Component.android.tsx` and **knip** will not detect it

webpro:
Interesting. How could Knip know about such platform-specific conventions? Sounds like it's not a generic file name convention like `index.ts`, but rather any component file such as `Button.tsx` must have `Button.android.tsx`, right?

neiker:
@webpro thanks for your response. This works with any file, not only components. You can start with `index.ts` but then you realize MacOS requires a totally different logic so you create `index.macos.ts`, and then the bundler will pick either `index.ts` or `index.macos.ts` at build time.
I think config can be something like: `"platformSpecificExtensions"` with the default `["ios", "android", "windows", "macos", "web"]`

webpro:
This is all specific to React Native, right? And *any* "base" source file could have one or more platform-specific "sibling" files?

This isn't about calculating the dependency tree and then see what's unused. It's more the other way around: see what unused files are actually used. Kind of feel this isn't something Knip should do. At this point I think it's better/easier to have a separate script walk the file tree and check whether any platform-specific file has an accompanying "base" file.

FYI, Knip plugins add project and entry files, compilers add non-JS/TS files, these are different things than what's requested.

neiker:
Yes, is react-native specific. There could be cases where there is no base file, and only platform-specific files exist. Eg. `something.android.js` and `something.ios.js` might exist but `something.js` don't. It's not common but it might happen. In that case, walking the file tree to see if the base file is used will not work because there's no base file

webpro:
Yes, that's what I meant: walk the file tree and look for `something.[ios|android]js` etc and see if `something.js` is there as well. If not, that's an issue. That's basically all it takes, right? In that case I'm not sure how it fits in what Knip is doing.

neiker:
I'm still having doubts about the case mentioned above when the base file doesn't exist. I can't see how your approach works in that case

[...]

PR #895, opened 2024-12-28
Add `metro` plugin

jjjjonathan:
This adds a plugin for [Metro](https://metrobundler.dev/), the JavaScript bundler for React Native.

### DIFF
--- a/packages/knip/src/plugins/metro/index.ts
+++ b/packages/knip/src/plugins/metro/index.ts
@@ -14,7 +14,7 @@ const isEnabled: IsPluginEnabled = options => hasDependency(options.dependencies
 
 const packageJsonPath = 'metro';
 
-const config: string[] = ['metro.config.{js,cjs,mjs,json}', 'package.json'];
+const config: string[] = ['metro.config.{js,cjs,json}', 'package.json'];
 
 const resolveEntryPaths: ResolveEntryPaths<PluginConfig> = async config => {
   if (!config.projectRoot) return [];
--- a/packages/knip/src/plugins/metro/index.ts
+++ b/packages/knip/src/plugins/metro/index.ts
@@ -8,7 +8,7 @@ import type { PluginConfig } from './types.js';
 
 const title = 'Metro';
 
-const enablers = [/^metro(-.*)?$/, 'react-native'];
+const enablers = ['metro', 'react-native'];
 
 const isEnabled: IsPluginEnabled = options => hasDependency(options.dependencies, enablers);
 
--- a/packages/knip/src/plugins/metro/index.ts
+++ b/packages/knip/src/plugins/metro/index.ts
@@ -1,0 +1,47 @@
+import { join } from '../../util/path.js';
+import type { IsPluginEnabled, Plugin, ResolveConfig, ResolveEntryPaths } from '../../types/config.js';
+import { toDeferResolve, toProductionEntry } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { PluginConfig } from './types.js';
+
+// https://metrobundler.dev/docs/configuration
+
+const title = 'Metro';
+
+const enablers = [/^metro(-.*)?$/, 'react-native'];
+
+const isEnabled: IsPluginEnabled = options => hasDependency(options.dependencies, enablers);
+
+const packageJsonPath = 'metro';
+
+const config: string[] = ['metro.config.{js,cjs,mjs,json}', 'package.json'];
+
+const resolveEntryPaths: ResolveEntryPaths<PluginConfig> = async config => {
+  if (!config.projectRoot) return [];
+
+  const entryFilePattern = 'index.{js,jsx,ts,tsx}';
+  const entryFilePath = join(config.projectRoot, entryFilePattern);
+  return [toProductionEntry(entryFilePath)];
+};
+
+const resolveConfig: ResolveConfig<PluginConfig> = async config => {
+  const { transformerPath, transformer } = config;
+  const inputs: string[] = [];
+
+  if (transformerPath) inputs.push(transformerPath);
+  if (transformer?.assetPlugins) inputs.push(...transformer.assetPlugins);
+  if (transformer?.minifierPath) inputs.push(transformer.minifierPath);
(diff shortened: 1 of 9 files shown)


packages/knip/src/plugins/metro/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import type { IsPluginEnabled, Plugin, ResolveConfig, ResolveEntryPaths } from '../../types/config.js';
+import { toDeferResolve, toProductionEntry } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { PluginConfig } from './types.js';
+
+// https://metrobundler.dev/docs/configuration
+
+const title = 'Metro';
+
+const enablers = [/^metro(-.*)?$/, 'react-native'];
```
jjjjonathan, line 11: Not sure if this is the best way to handle this, but typically Metro is not in the project's package.json, as it's a dependency of React Native

packages/knip/src/plugins/metro/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const isEnabled: IsPluginEnabled = options => hasDependency(options.dependencies, enablers);
+
+const packageJsonPath = 'metro';
+
+const config: string[] = ['metro.config.{js,cjs,mjs,json}', 'package.json'];
```
jerone, line 17: It looks like the official list of supported extensions is a little bit different: https://github.com/facebook/metro/blob/main/packages/metro-config/src/loadConfig.js#L48-L53

jjjjonathan: Thanks @jerone ! I removed `mjs` as a valid extension.

packages/knip/src/plugins/metro/index.ts
webpro: In that case, let's just use only `react-native`? Is that ever not listed?

webpro (commented):
Thanks Jonathan, this is great!

The reason I held off introducing this plugin before is that it's not trivial to properly support Metro's module resolution, i.e. support the various `*.{android,ios,web}.ts` etc. files. It basically means added complexity and creating the graph for each target separately (also see https://github.com/webpro-nl/knip/issues/126 if you haven't already).

How are you handling this in your project(s)?

There's a similar situation with [the Nuxt plugin](https://knip.dev/reference/plugins/nuxt), which might lead to (lots of) false positives/confusion. We could add a note for this plugin too, should probably stand out more (not necessarily in this PR though).

I think overall this is definitely worth merging, I just like to also make sure users understand what "works with RN/Metro" means and how they should reason/go about it.

packages/knip/src/plugins/metro/index.ts
jjjjonathan: @webpro It's definitely possible to have a project without `react-native` listed. Metro can also bundle for web, although it'd be pretty rare to use it for exclusively that purpose over other web bundlers. Or the project could be a custom dev server that uses Metro's bundling API. Again, a rare use case, but possible.

jjjjonathan:
@webpro Thanks for the review! I have looked through #126 in the past and I totally respect that it would add a lot of complexity to Knip. I don't have a lot of platform-specific extensions in my projects, but I have experimented with a little script that simply finds files with platform-specific extensions, and if they have a matching file without a platform-specific extension, it adds the file to my `knip.config.ts` as an entry point.

For me personally, there's a great benefit to a simple plugin to add entry points and dependencies even without the platform-specific extensions support.

If a note about RN module resolution/platform-specific extensions is added to the docs, I'd suggest adding it for the Expo plugin too. Metro is a little more technical and new RN users will likely encounter Expo first (it's the recommended way to get started with RN from the RN docs).

packages/knip/src/plugins/metro/index.ts
webpro: We can/should obviously leave some `metro` specific enabler(s) here, but this regex matches anything that starts with `metro` which is too loose?

webpro:
> I don't have a lot of platform-specific extensions in my projects, but I have experimented with a little script that simply finds files with platform-specific extensions, and if they have a matching file without a platform-specific extension, it adds the file to my knip.config.ts as an entry point.

> For me personally, there's a great benefit to a simple plugin to add entry points and dependencies even without the platform-specific extensions support.

That's cool but if we market Knip as "supports React Native" I'd expect users in RN projects to run Knip and then face a ton of unused (platform-specific) files reported. For them I need a clear path out (tbh actually it's already too late for a good first impression then, but that's the risk I might be willing to take).

jjjjonathan:
@webpro I understand; it definitely makes sense from a "good first impression" perspective. Let me know if I can help move things along in this PR or another!

webpro:
Yeah not that Knip is perfect by any means, but I can easily imagine situations where many false positives might be overwhelming for new users. Was thinking we could add a note to the plugin page and whip up a page dedicated to RN, perhaps a bit like https://knip.dev/features/integrated-monorepos. Maybe configuration hints directly in the CLI output.

If you could make this PR green then we can wrap this up soon I guess :)

webpro:
Thanks Jonathan! Planning to prepare a release in the coming days.

webpro:
Tried a few projects and using those default production entry patterns work pretty well:

https://github.com/webpro-nl/knip/blob/093dcca11155072159175830cfbc9c3cca0dde8d/packages/knip/src/plugins/metro/index.ts#L23-L41

E.g. in https://github.com/pass-culture/pass-culture-app-native in don't need to change the defaults in others there's only a small number of false positives. With the simple plugin docs hopefully it's pretty clear what's going on:

https://4f7938f0.knip.pages.dev/reference/plugins/metro

If you'd like to try it out install with something like `npm i -D https://pkg.pr.new/knip@093dcca`

In case you have any concerns or feedback I'd love to hear about it :)

webpro:
E.g. in https://github.com/Expensify/App this seems to work well:

```json
{
  "metro": {
    "entry": ["src/**/*.{ios,android,windows,web,native,default,desktop,website}.{js,jsx,json,ts,tsx}"],
  },
}
```

webpro:

PR #904, opened 2025-01-08
Add support for Babel plugins used in Vite/React config

### LINKED ISSUE #903: 🐛Babel plugin passed to @vitejs/plugin-react reported as unused
### Prerequisites

- [X] I'm using the latest version
- [X] I've read the relevant [documentation](https://knip.dev)
- [X] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [X] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [X] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/mbacalan/knip-vite-react-emotion-repro

### Reproduction access

- [X] I've made sure the reproduction is publicly accessible

### Description of the issue

Plugins can be declared for Babel through `@vitejs/plugin-react` configuration object in Vite config file.

```js
export default defineConfig({
  plugins: [react({
    babel: {
      plugins: ['@emotion/babel-plugin']
    }
  })],
})
```

Knip reports these plugins as unused devDependencies:
image

mbacalan:
Fixes #903

### DIFF
--- a/packages/knip/src/plugins/vite/index.ts
+++ b/packages/knip/src/plugins/vite/index.ts
@@ -1,6 +1,12 @@
-import type { IsPluginEnabled, Plugin } from '../../types/config.js';
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import { type Input, toEntry } from '../../util/input.js';
+import { join } from '../../util/path.js';
 import { hasDependency } from '../../util/plugin.js';
-import { resolveConfig, resolveEntryPaths } from '../vitest/index.js';
+import { getDependenciesFromConfig } from '../babel/index.js';
+import type { BabelConfigObj } from '../babel/types.js';
+import { findConfigDependencies, getConfigs, resolveEntryPaths } from '../vitest/index.js';
+import type { ViteConfig, ViteConfigOrFn, VitestWorkspaceConfig } from '../vitest/types.js';
+import type { BabelPlugin } from './types.js';
 
 // https://vitejs.dev/config/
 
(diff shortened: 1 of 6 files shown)


mbacalan:
I tried to keep the changes minimal but had to copy over the `resolveConfig` fn from `vitest` plugin to add the additional plugin checks to it. I'm not sure if this is the way tbh. Looking forward to your feedback!

packages/knip/src/plugins/vitest/index.ts
```diff
@@ -48,7 +48,7 @@ const findConfigDependencies = (localConfig: ViteConfig, options: PluginOptions)
   return [...[...environments, ...reporters, ...coverage].map(toDependency), ...setupFiles, ...globalSetup];
 };
 
-const getConfigs = async (localConfig: ViteConfigOrFn | VitestWorkspaceConfig) => {
+export const getConfigs = async (localConfig: ViteConfigOrFn | VitestWorkspaceConfig) => {
```
webpro, line 51: Perhaps we can import the babel function here and handle it from the vitest plugin?

packages/knip/src/plugins/vite/index.ts
```diff
@@ -12,6 +18,52 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
...
+      }
+    } else if (typeof plugin === 'object' && plugin !== null) {
+      const pluginObj = plugin as BabelPlugin;
+      if (pluginObj?.babel?.plugins) {
+        return true;
+      }
+    }
+  }
+
+  return false;
```
webpro, line 45: I didn't realize at first the babel object is sent to the react plugin, but if this works then... all good! :)

packages/knip/src/plugins/vite/index.ts
```diff
@@ -12,6 +18,52 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
 
 export const config = ['vite.config.{js,mjs,ts,cjs,mts,cts}'];
 
+const hasReactBabelPlugins = (config: ViteConfig) => {
```
webpro, line 21: If we can move it over to the vitest plugin, then this helper could go the `helpers.ts`

webpro (commented):
Thanks @mbacalan, much appreciated!

I'll enable CI and then it'll run in a couple of projects automatically as a "ecosystem smoke test".

mbacalan:
Thanks for the feedback. Upon further inspection, it looks like these changes don't solve the issue.
I think I misunderstood the assertion in the test file. It actually means that we are getting 1 devDependencies error while the goal is to not receive any since we have `@emotion/babel-plugin` in `package.json` and `vite.config.ts`, right?

I'll get back to it soon and take into account the points you raised.

webpro:
Initially I saw `babel: { plugins: ['@emotion/babel-plugin'] }` as a prop of the Vite config and that would make things easy.

However, if the `react()` plugin doesn't just return that object argument (or anything useful) then we currently just don't have a solution for this. So, could you try and see if there's anything useful returned in the `config` object, just to be 100% sure?

It's basically the same problem as #856 where we also ideally would have an easy way to - instead of loading the configuration file and read as an object - use the AST and traverse it to find strings like `"@emotion/babel-plugin"`. But that comes with a cost of performance impact and extra complexity.

At the very least this serves as an extra use case that probably involves reading and parsing config file ASTs.

webpro:
Thanks for digging through. And no worries about my time, mostly glad to hear you got something out of Knip!

PR #910, opened 2025-01-12
Unused exported type & interface members

webpro:
WIP. From: https://github.com/webpro-nl/knip/pull/848#issuecomment-2585735560

Initial docs: https://09833a2d.knip.pages.dev/guides/type-members


tylersayshi:
This sounds like a helpful thing to warn on. 

At the same time, it seems perfectly valid to have unused type members for things like the result of an external REST api request. I'd type out all members of the expected response and maybe not use all of the fields at first.

webpro:
Iteration speed is definitely a thing to consider, yet could be mitigated using local JSDoc tags or config e.g. `ignore: ["src/{api,generated}/**"]` to (temporarily) exclude files or parts of the codebase. But won't suit everyone, takes a little bit of extra work/config. In the end the issue type is opt-in to begin with.

Issue #907, opened 2025-01-10
💡 Add support for dependency-cruiser 


filipw01:
### Suggest an idea for Knip

https://github.com/sverweij/dependency-cruiser

It should consider config file passed to it used

webpro:
> It should consider config file passed to it used

Can you please elaborate? I guess it should not report `.dependency-cruiser.js` as unused if `dependency-cruiser` is listed in `package.json, right? Or is there more to it?

Would you be up for a PR? Sounds like a small plugin :)

filipw01:
Yes, it should detect the default `.dependency-cruiser.js` as well as support custom config passed in using  `--config` and `-c` options

PR #911, opened 2025-01-12
feat: add dependency-cruiser plugin

filipw01:
Fixes: #907

### DIFF
--- a/packages/knip/src/plugins/dependency-cruiser/index.ts
+++ b/packages/knip/src/plugins/dependency-cruiser/index.ts
@@ -1,0 +1,27 @@
+import type { IsPluginEnabled, Plugin } from '../../types/config.js';
+import { hasDependency } from '../../util/plugin.js';
+
+// https://github.com/sverweij/dependency-cruiser
+
+const title = 'dependency-cruiser';
+
+const enablers = ['dependency-cruiser'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['.dependency-cruiser.{js,cjs,mjs,json}'];
+
+const args = {
+  binaries: ['depcruise'],
+  alias: { config: ['c'] },
+  string: ['config'],
+  config: true,
+};
+
+export default {
+  title,
+  enablers,
+  isEnabled,
+  config,
+  args,
+} satisfies Plugin;
--- a/packages/knip/src/plugins/dependency-cruiser/index.ts
+++ b/packages/knip/src/plugins/dependency-cruiser/index.ts
@@ -12,9 +12,7 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
 const config = ['.dependency-cruiser.{js,cjs,mjs,json}'];
 
 const args = {
-  binaries: ['depcruise'],
-  alias: { config: ['c'] },
-  string: ['config'],
+  binaries: ['depcruise', 'dependency-cruise', 'dependency-cruiser', 'depcruise-baseline'],
   config: true,
 };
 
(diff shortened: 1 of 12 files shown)


packages/knip/src/plugins/dependency-cruiser/index.ts
```diff
@@ -0,0 +1,27 @@
...
+const title = 'dependency-cruiser';
+
+const enablers = ['dependency-cruiser'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['.dependency-cruiser.{js,cjs,mjs,json}'];
+
+const args = {
+  binaries: ['depcruise'],
```
webpro, line 15: looks like they have many: https://github.com/sverweij/dependency-cruiser/blob/c48aca959c9c9c1b1068c6d85cd4ed368de10d27/package.json#L74-L80

any chance you could add those as well?

packages/knip/src/plugins/dependency-cruiser/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  alias: { config: ['c'] },
+  string: ['config'],
+  config: true,
```
webpro, line 18: setting `config: true` is a shorthand that sets the `alias` and `string` options as well: https://github.com/webpro-nl/knip/blob/5a77dcc47c2cdfc026a5296f1dc3cf54c786c562/packages/knip/src/types/args.ts#L59-L69

just a heads-up, no need to change if you prefer explicit

packages/knip/test/plugins/dependency-cruiser.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+test('Find dependencies with the dependency-cruiser plugin', async () => {
+  const { issues, counters } = await main({
+    ...baseArguments,
+    cwd,
+  });
+
+  // Check that dependencies are properly detected
+  assert(issues.devDependencies['package.json']['dependency-cruiser']);
+  assert(issues.binaries['package.json']['depcruise']);
```
webpro, line 18: what i usually do is add the most minimal package in an actual `node_modules` folder with e.g. `{ "name": "dependency-cruiser", "bin": { ... } }` in `package.json` and an empty `index.js` - that way, Knip does the full machinery and there should be no issues left (random example: https://github.com/webpro-nl/knip/tree/main/packages/knip/fixtures/plugins/angular/node_modules/@angular/cli).

webpro (commented):
Awesome! Love that you also did the `args` thing :)

packages/knip/src/plugins/dependency-cruiser/index.ts
filipw01: Nice, I didn't know that

packages/knip/test/plugins/dependency-cruiser.test.ts
filipw01: That's awesome. I was a bit lost working on the test

packages/knip/src/plugins/dependency-cruiser/index.ts
filipw01: Yes, added the relevant ones

filipw01:
@webpro Is there an option to colocate tests and fixtures? It was a bit annoying how far they were in the file structure when working on the test. I'm happy to open a PR moving it around

webpro:
> @webpro Is there an option to colocate tests and fixtures? It was a bit annoying how far they were in the file structure when working on the test. I'm happy to open a PR moving it around

Gotcha. I can relate, but overall this structure is very convenient to me.

filipw01:
Thanks for taking care of that. Was a bit busy this week

webpro:
:rocket: _This pull request is included in v5.42.1. See [Release 5.42.1](https://github.com/webpro-nl/knip/releases/tag/5.42.1) for release notes._

webpro:
No worries, was tiny change. Thanks again for the PR!

PR #915, opened 2025-01-17
Add pm2-runtime to IGNORED_GLOBAL_BINARIES constant

meowsus:
I'm proposing we add `pm2-runtime` to the list of `IGNORED_GLOBAL_BINARIES` based on [this documentation](https://pm2.keymetrics.io/docs/usage/docker-pm2-nodejs/) which indicates that `pm2` should be installed globally inside of Docker containers and invoked using the `pm2-runtime` binary.

In my specific use case we have a script in our `package.json` file called `workers:prod` which uses this binary, e.g. `pm2-runtime start pm2.config.js`, which will fail the `knip --include binaries` check since its not added to our package and is only necessary in our production environment.

I'm curious if others have a strong feeling for or against including this entry.

### DIFF
--- a/packages/knip/src/constants.ts
+++ b/packages/knip/src/constants.ts
@@ -92,6 +92,7 @@ export const IGNORED_GLOBAL_BINARIES = new Set([
   'nproc',
   'npx',
   'paste',
+  'pm2-runtime',
   'pnpm',
   'pnpx',
   'pr',


webpro:
Although I understand the use case it might set a precedent for similar cases (I mean, what's next?). The thing is, Knip recognizes `"start"` as a "production script" in `package.json` - but not `"workers:prod"`.

I think it would be better to somehow make Knip understand what the "production scripts" are in `package.json`. Maybe through some new config option, not sure yet.

For now please consider using `ignoreBinaries: ["pm2-runtime"]`

meowsus:
Absolutely agree. Thanks for the rundown, @webpro. I must have missed the little blurb [here](https://knip.dev/guides/handling-issues#unreachable-code), but thanks for pointing it out :D

PR #922, opened 2025-01-21
Fix vitest reporters with options

### LINKED ISSUE #921: 🐛 Vitest plugin doesn't handle reporters with options
### Prerequisites

### Reproduction url

https://github.com/dakro/knip-issues/tree/vitest-reporters

### Reproduction access

### Description of the issue

Vitest reporters can be in the form:
`['reporter-name', {options:{}}]`

Knip reports unused 'reporter-name' dependency in this case

dakro:
Fix #921

### DIFF
--- a/packages/knip/src/plugins/vitest/helpers.ts
+++ b/packages/knip/src/plugins/vitest/helpers.ts
@@ -45,5 +45,6 @@ export const getExternalReporters = (reporters?: ViteConfig['test']['reporters']
   reporters
     ? [reporters]
         .flat()
+        .flat()
         .filter((reporter): reporter is string => typeof reporter === 'string' && !builtInReporters.includes(reporter))
     : [];
--- a/packages/knip/src/plugins/vitest/helpers.ts
+++ b/packages/knip/src/plugins/vitest/helpers.ts
@@ -45,6 +45,6 @@ export const getExternalReporters = (reporters?: ViteConfig['test']['reporters']
   reporters
     ? [reporters]
         .flat()
-        .flat()
+        .map(reporter => (Array.isArray(reporter) ? reporter[0] : reporter))
         .filter((reporter): reporter is string => typeof reporter === 'string' && !builtInReporters.includes(reporter))
     : [];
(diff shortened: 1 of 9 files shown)


packages/knip/src/plugins/vitest/helpers.ts
```diff
@@ -44,6 +44,7 @@ const builtInReporters = [
 export const getExternalReporters = (reporters?: ViteConfig['test']['reporters']) =>
   reporters
     ? [reporters]
+        .flat()
         .flat()
         .filter((reporter): reporter is string => typeof reporter === 'string' && !builtInReporters.includes(reporter))
```
webpro, line 49: Extra flattening may lead to unexpected reporters maybe if there's a string "anywhere"? How about instead of the extra `flat()` perhaps it would make sense to return `reporter` if it's a string or `reporter[0]` for an array (or ignore otherwise)?

packages/knip/test/plugins/vitest7.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  assert(issues.unlisted['vitest.config.ts']['ReporterString']);
+  assert(issues.unlisted['vitest.config.ts']['ReporterArray']);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    files: 0,
+    processed: 3,
+    total: 3,
+    unlisted: 2,
```
webpro, line 24: This is fine, yet ideally we'd add `ReporterString` and `ReporterArray` to dependencies in `package.json` so they'll be both listed and referenced, and there should be no issues left here in the test ("full circle").

webpro (commented):
Thanks! Happy to merge, but two remarks I have.

packages/knip/src/plugins/vitest/helpers.ts
dakro: Yes, at the moment should not be possible to have a string, but in this way will be more future proof, I will change as suggested

packages/knip/test/plugins/vitest7.test.ts
dakro: Do you prefer just to change this test or to add another one?

webpro: This one test should be enough!

webpro:
Thanks @dakro! Great stuff.

webpro:
:rocket: _This pull request is included in v5.42.3. See [Release 5.42.3](https://github.com/webpro-nl/knip/releases/tag/5.42.3) for release notes._

PR #864, opened 2024-11-30
`yarn.config.cjs` support

jwoo0122:
Close #597

PR #924, opened 2025-01-22
feat: make knip understand yarn pnp

jwoo0122:
Make knip understand [yarn pnp](https://yarnpkg.com/features/pnp). It can now look into manifest files inside yarn cache, and correctly check peer dependency or bin field.

Maybe we can make this feature to be enabled with yarn plugin made from #864 but I couldn't find how to patch manifest helpers with plugin.

You should see `manifest/helpers.ts`.


packages/knip/fixtures/yarn-pnp/.yarnrc.yml
jwoo0122, line 1: I made a real yarn workspace to test real pnp api, but it'll be good to see any ideas for this fixture.

jwoo0122:
I've checked test code with `bun test test/yarn-pnp.test.ts` in `packages/knip` directory and it passed, but seems like it failed when runs with `npm test`. I'll fix it.

packages/knip/src/manifest/helpers.ts
```diff
@@ -1,12 +1,87 @@
...
+    const parentDir = dirname(currentDir);
+    if (parentDir === currentDir) {
+      break; // Reached root
+    }
+    currentDir = parentDir;
+  }
+  pnpStatus.dir = startDir;
+  pnpStatus.pnpPath = '';
+  pnpStatus.enabled = false;
+};
```
jwoo0122, line 37: I made this by assuming that knip needs to be able to understand yarn pnp, even it isn't under yarn pnp environment, e.g `yarn dlx knip`. If it's not, we can just simply import pnpapi directly, instead of manually finding it.

webpro:
Thanks for the PR, @jwoo0122. Without knowing exactly how PnP works, please let my try to understand what support for PnP means to Knip, and what features (i.e. issue types) are still (un)available/relevant.

Some initial questions that pop up in my mind:

- The fixture has only "internal" packages and ignores a bunch of stuff, but perhaps we could see a more realistic example?
- Is there support for monorepos/workspaces (i.e. `package.json#workspaces`)? Or is that not relevant cq nothing to worry about?
- Since there are no `node_modules` anymore, what about:
  - Binaries (`node_modules/.bin`) e.g. those referenced in `package.json#scripts`
  - Knip reads `package.json#types` etc. to see if types are included with the package (so e.g. `@​types/eslint` is unused if `eslint` is listed).

webpro:
Closing this due to inactivity and lack of responses. Feel free to reopen, discuss, or upvote as you see fit.

PR #925, opened 2025-01-24
feat(jest): add __mocks__ to entry files

TkDodo:
jest allows defining manual mocks within a special __mocks__ directory, so those files should be seen as used by knip. See: https://jestjs.io/docs/manual-mocks

### DIFF
--- a/packages/knip/src/plugins/jest/index.ts
+++ b/packages/knip/src/plugins/jest/index.ts
@@ -16,7 +16,7 @@ const isEnabled: IsPluginEnabled = ({ dependencies, manifest }) =>
 
 const config = ['jest.config.{js,ts,mjs,cjs,json}', 'package.json'];
 
-const entry = ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).[jt]s?(x)', '**/__mocks__/**'];
+const entry = ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).[jt]s?(x)', '**/__mocks__/**/*.[jt]s?(x)'];
 
 const resolveDependencies = async (config: JestInitialOptions, options: PluginOptions): Promise<Input[]> => {
   const { configFileDir } = options;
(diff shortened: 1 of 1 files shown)


packages/knip/src/plugins/jest/index.ts
```diff
@@ -16,7 +16,7 @@ const isEnabled: IsPluginEnabled = ({ dependencies, manifest }) =>
 
 const config = ['jest.config.{js,ts,mjs,cjs,json}', 'package.json'];
 
-const entry = ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).[jt]s?(x)'];
+const entry = ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).[jt]s?(x)', '**/__mocks__/**'];
```
webpro, line 19: Would be nice to have a file in that folder won't be reported as unused anymore

webpro (commented):
Thanks, Dominik! One nit I have.

packages/knip/src/plugins/jest/index.ts
webpro: I mean in one of the jest fixtures folders

TkDodo: right, how can we achieve that?

what I did before was exclude them in `project`:

```
!/**/__mocks__/**
```

but I don’t think that’s what we’d generally want?

webpro: We could an empty file such as `__mocks__/used.js` in of the fixtures folders. Before the change that was an unused file.

Btw, maybe also better to use something more specific like `**/__mocks__/**/*.{js,tx}`  - would that suffice?

TkDodo: like this?

256aa4a9

webpro: I see my own typo now 🤦‍♂️ but yeah that's fine (i wouldn't expect jsx to be necessary here)

Any chance you could still add that file?

TkDodo: right, yeah, I think I got now what you mean: 615dd83a

webpro:
:rocket: _This pull request is included in v5.43.3. See [Release 5.43.3](https://github.com/webpro-nl/knip/releases/tag/5.43.3) for release notes._

PR #931, opened 2025-02-02
feat: Add plugin for React router 7 framework mode

lasseklovstad:
Hi, I was hoping to add support for React Router 7 framework mode 😄 I have tried my best to implement a plugin based on the docs. In React Router 7 the `routes.ts` file defines all the routes so I map over every route entry and their children, I think this is the correct way to do it. This plugin should also handle only js files aswell.

If you use react router as library mode (only having `react-router` as dependency) the plugin should not be enabled. So I only enabled it when `@react-router/dev` is present.

I was a bit unsure what to include in `packages/knip/fixtures/plugins/react-router/**` directory. I asumed that I only include mock files to be used when running the tests, so I omited the vite config for instance.

### DIFF
--- a/packages/knip/fixtures/plugins/react-router/app/routes.ts
+++ b/packages/knip/fixtures/plugins/react-router/app/routes.ts
@@ -1,0 +1,7 @@
+export default [
+  { file: "routes/home.tsx", index: true },
+  {
+    file: "routes/layout.tsx",
+    children: [{ file: "routes/another-route.tsx" }],
+  },
+];
--- a/packages/knip/fixtures/plugins/react-router/app/routes.ts
+++ b/packages/knip/fixtures/plugins/react-router/app/routes.ts
@@ -2,6 +2,6 @@ export default [
   { file: "routes/home.tsx", index: true },
   {
     file: "routes/layout.tsx",
-    children: [{ file: "routes/another-route.tsx" }],
+    children: [{ file: "./routes/another-route.tsx" }],
   },
 ];
--- a/packages/knip/src/plugins/react-router/index.ts
+++ b/packages/knip/src/plugins/react-router/index.ts
@@ -1,41 +1,33 @@
-import { existsSync } from "node:fs";
-import type {
-  IsPluginEnabled,
-  Plugin,
-  ResolveEntryPaths,
-} from "../../types/config.js";
-import { toEntry } from "../../util/input.js";
-import { join } from "../../util/path.js";
-import { hasDependency, load } from "../../util/plugin.js";
-import type { PluginConfig, RouteConfigEntry } from "./types.js";
+import { existsSync } from 'node:fs';
+import type { IsPluginEnabled, Plugin, ResolveEntryPaths } from '../../types/config.js';
+import { toEntry } from '../../util/input.js';
+import { join } from '../../util/path.js';
+import { hasDependency, load } from '../../util/plugin.js';
+import type { PluginConfig, RouteConfigEntry } from './types.js';
 
 // https://reactrouter.com/start/framework/routing
 
-const title = "react-router";
+const title = 'react-router';
 
-const enablers = ["@react-router/dev"];
+const enablers = ['@react-router/dev'];
 
-const isEnabled: IsPluginEnabled = ({ dependencies }) =>
-  hasDependency(dependencies, enablers);
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
-const config: string[] = ["react-router.config.{js,ts}"];
+const config: string[] = ['react-router.config.{js,ts}'];
 
 const entry: string[] = [];
 
 const production: string[] = [];
 
-const resolveEntryPaths: ResolveEntryPaths<PluginConfig> = async (
-  localConfig,
-  options
-) => {
+const resolveEntryPaths: ResolveEntryPaths<PluginConfig> = async (localConfig, options) => {
   const { configFileDir } = options;
-  const appDirectory = localConfig.appDirectory ?? "app";
+  const appDirectory = localConfig.appDirectory ?? 'app';
   const appDir = join(configFileDir, appDirectory);
 
   let routeConfig: RouteConfigEntry[] = [];
 
-  const routesPathTs = join(appDir, "routes.ts");
-  const routesPathJs = join(appDir, "routes.js");
+  const routesPathTs = join(appDir, 'routes.ts');
+  const routesPathJs = join(appDir, 'routes.js');
 
   if (existsSync(routesPathTs)) {
     routeConfig = await load(routesPathTs);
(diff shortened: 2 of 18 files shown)


packages/knip/src/plugins/react-router/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  let routeConfig: RouteConfigEntry[] = [];
+
+  const routesPathTs = join(appDir, 'routes.ts');
+  const routesPathJs = join(appDir, 'routes.js');
+
+  if (existsSync(routesPathTs)) {
+    routeConfig = await load(routesPathTs);
+  } else if (existsSync(routesPathJs)) {
+    routeConfig = await load(routesPathJs);
```
webpro, line 35: maybe good idea to check if `routeConfig` is actually set later on?

packages/knip/fixtures/plugins/react-router/app/routes.ts
```diff
@@ -0,0 +1,7 @@
+export default [
+  { file: "routes/home.tsx", index: true },
+  {
+    file: "routes/layout.tsx",
+    children: [{ file: "routes/another-route.tsx" }],
+  },
+];
```
webpro, line 7: seeing things in the docs like `route("/", "./home.tsx")` - do we cater for that as well? or how's that returned when loading `app/routes.ts`?

webpro:
> Hi, I was hoping to add support for React Router 7 framework mode 😄 I have tried my best to implement a plugin based on the docs. In React Router 7 the `routes.ts` file defines all the routes so I map over every route entry and their children, I think this is the correct way to do it. This plugin should also handle only js files aswell.

Thanks! Good stuff indeed 👍

> If you use react router as library mode (only having `react-router` as dependency) the plugin should not be enabled. So I only enabled it when `@react-router/dev` is present.

Do you think we might get in trouble if library mode will be supported here? Can we distinguish that easily, or would we need separate plugins, if at all necessary?

> I was a bit unsure what to include in `packages/knip/fixtures/plugins/react-router/**` directory. I asumed that I only include mock files to be used when running the tests, so I omited the vite config for instance.

This looks fine. Maybe you know of a public project we could run this on for some additional smoke testing?

lasseklovstad:
> > If you use react router as library mode (only having `react-router` as dependency) the plugin should not be enabled. So I only enabled it when `@react-router/dev` is present.
> 
> Do you think we might get in trouble if library mode will be supported here? Can we distinguish that easily, or would we need separate plugins, if at all necessary?

Do you mean if library mode wil be supported by `@react-router/dev` or  if framwork mode will be supported by `react-router` package? In the [`@react-router/dev`](https://github.com/remix-run/react-router/tree/main/packages/react-router-dev) package they describe it as `Dev tools and CLI for React Router that enables framework features through bundler integration like server rendering, code splitting, HMR, etc.`. So for now its safe to say that your using react router as framwork mode if you have that package installed. 
> 
> > I was a bit unsure what to include in `packages/knip/fixtures/plugins/react-router/**` directory. I asumed that I only include mock files to be used when running the tests, so I omited the vite config for instance.
> 
> This looks fine. Maybe you know of a public project we could run this on for some additional smoke testing?

Yes. We can test it on [The epic stack](https://github.com/epicweb-dev/epic-stack) which uses `remix-flat-routes` package. They also use `remixRoutesOptionAdapter` which is a function that eases migration from remix to react router 7.

I have a public project which uses react router flat routes (`@react-router/fs-routes`) package: https://github.com/lasseklovstad/gata
And I have a private repository which just uses the utility functions from react router index, layout, prefix and so on...

I can do some smoke testing tonight.

webpro:
> > > If you use react router as library mode (only having `react-router` as dependency) the plugin should not be enabled. So I only enabled it when `@react-router/dev` is present.
> > 
> > 
> > Do you think we might get in trouble if library mode will be supported here? Can we distinguish that easily, or would we need separate plugins, if at all necessary?
> 
> Do you mean if library mode wil be supported by `@react-router/dev` or if framwork mode will be supported by `react-router` package? In the [`@react-router/dev`](https://github.com/remix-run/react-router/tree/main/packages/react-router-dev) package they describe it as `Dev tools and CLI for React Router that enables framework features through bundler integration like server rendering, code splitting, HMR, etc.`. So for now its safe to say that your using react router as framwork mode if you have that package installed.

What I meant to ask was: is it possible we might need another plugin for react-router? Ideally we can keep just one plugin, unless it becomes too messy e.g. for having to support multiple modes.

packages/knip/fixtures/plugins/react-router/app/routes.ts
lasseklovstad: `route("/", "./home.tsx")` returns a simple object containing `{ file: "./home.tsx", path: "/" }`. Since we are joining the path like this: `join(appDir, route.file)` and the `routes.ts` file is defined in `appDir` this is catered for.

packages/knip/src/plugins/react-router/index.ts
lasseklovstad: I initialize `routeConfig` as an empty array above. So if no routes.ts file is defined it is ignored.

packages/knip/src/plugins/react-router/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+// https://reactrouter.com/start/framework/routing
+
+const title = 'react-router';
+
+const enablers = ['@react-router/dev'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['react-router.config.{js,ts}', ...vite.config];
```
lasseklovstad, line 17: I added vite.config aswell because the `react-router.config` is optional.

webpro: All `config` paths are loaded and if there's `resolveConfig` its executed with the default exported value of those config files. Also they're automatically added as `entry` files (so their imports are registered as dependencies like any other source file).

Since this plugin doesn't have `resolveConfig` I think `'react-router.config.{js,ts}'` should be moved to the `entry` array (guess I've missed this before), and looks like there's no value in adding `vite.config`.

Just saying, the `config` array items do not influence whether the plugin is enabled or not.

packages/knip/src/plugins/react-router/index.ts
webpro: Ah, gotcha. Good!

webpro:
> I did some smoketesting on epic-stack and my personal projects and it seemed to work fine. Found some edge cases in the process which I fixed.

Thanks a lot!

> One thing to mention is that react router generates types with the command `react-router typegen` if these types are not generated knip will complain about `Unresolved imports`. Is this something we should handle?

Good question, tbh I'm not sure. If it's recommended to generate those I guess not. Maybe if used in JavaScript-only context that would be annoying, and we could add an `entry` item.

packages/knip/src/plugins/react-router/index.ts
webpro: Any chance you could move `config` to `entry` and remove `vite.config` - I'd guess that gives the same results, right?

lasseklovstad: Moving `react-router.config.ts` to `entry` and removing vite doesn't seem to enable the plugin. When console logging inside `resolveEntryPaths` nothing happends. Not sure whats happening.

webpro: Ah sorry I looked over the fact that `localConfig` is an argument to `resolveEntryPaths` and thus a `config` item is required. Took the liberty to clean up and merge :)

webpro:
:rocket: _This pull request is included in v5.44.0. See [Release 5.44.0](https://github.com/webpro-nl/knip/releases/tag/5.44.0) for release notes._

PR #933, opened 2025-02-05
fix: ensure augmentWorkspace honors tsConfig rootDir

rmunch:
I _think_ this fixes logic that has a bug but due to incorrect parenthesis in a ternary with a nullish coalescing operator. When a `rootDir` is specified in tsconfig.json that is something other than the value `"src"`, the `rootDir` config is not honored. In my case, I _do_ have a `src` dir, but it's unrelated to my TypeScript source which is located elsewhere (where my `rootDir` poins to). Knip ends up forcing the workspace `srcDir` value to `"src"` due to this bug combined with the presence of a `src` directory.

I attempted to add a test, but I'm not so familiar with the codebase and could not seem to exercise the right code paths 😞, but can give it another try if needed.

### DIFF
--- a/packages/knip/src/util/to-source-path.ts
+++ b/packages/knip/src/util/to-source-path.ts
@@ -13,7 +13,7 @@ const matchExt = /(\.d)?\.(m|c)?(j|t)s$/;
 
 export const augmentWorkspace = (workspace: Workspace, dir: string, compilerOptions: CompilerOptions) => {
   const srcDir = join(dir, 'src');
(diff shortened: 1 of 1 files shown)


webpro:
Thanks @rmunch! Simply by looking at the logic/syntax I'm pretty sure you're right. The parentheses placement made no sense.

webpro:
:rocket: _This pull request is included in v5.44.0. See [Release 5.44.0](https://github.com/webpro-nl/knip/releases/tag/5.44.0) for release notes._

PR #937, opened 2025-02-08
feat: Support Stylelint custom syntax object

TJNhxMZHmqGytuWT:
The [`customSyntax`](https://stylelint.io/user-guide/configure/#customsyntax) option in Stylelint config can accept a module name, but it can also accept a [Syntax object](https://postcss.org/docs/how-to-write-custom-syntax#syntax):

```js
// .stylelintrc.js
module.exports = {
  customSyntax: require('postcss-less'),
};
```

Some examples in the wild:
https://github.com/search?q=%22customSyntax%3A+require%28%22&type=code

Since Knip assumes that `config.customSyntax` can only be a string, running it with this kind of config results in the following error:

```
export const isInNodeModules = (filePath) => filePath.includes('node_modules');
                                                      ^
TypeError: filePath.includes is not a function
```

I updated the Stylelint plugin to avoid trying to resolve syntax objects.

I also added tests for JS config files.


packages/knip/test/plugins/stylelint2.test.ts
```diff
@@ -0,0 +1,29 @@
...
+test('Find dependencies with the stylelint plugin (2)', async () => {
+  const { issues, counters } = await main({
+    ...baseArguments,
+    cwd,
+  });
+
+  assert(issues.dependencies['package.json']['stylelint-order']);
+  assert(issues.unresolved['stylelint.config.js']['stylelint-config-standard']);
+  assert(issues.unlisted['stylelint.config.js']['stylelint-config-recommended']);
+  assert(issues.unresolved['.stylelintrc.mjs']['stylelint-config-standard']);
```
webpro, line 19: The idea of most of the plugin tests is to not leave any issues, so only the `processed` and `total` counters should not be zero (0). In other words, the dependencies referenced in the plugin config files in the fixtures should also be listed in `package.json`.

webpro (commented):
Thanks for the PR! Happy to merge, but please first see the feedback.

packages/knip/src/plugins/stylelint/types.ts
```diff
@@ -1,5 +1,5 @@
 export type BaseStyleLintConfig = {
-  customSyntax?: string;
+  customSyntax?: string | object;
```
webpro, line 2: Maybe `unknown` would be better than `object` here?

webpro (commented):
Thanks!

Idea: if we'd remove the `node_modules` folder in the fixture folder, would we still get the same test results? This is usually necessary to have in certain situations with binaries and such. If it's not doing much we might as well delete it.

webpro:
Perfect, thank you!

webpro:
:rocket: _This pull request is included in v5.44.1. See [Release 5.44.1](https://github.com/webpro-nl/knip/releases/tag/5.44.1) for release notes._

PR #970, opened 2025-03-06
feat: add Gitlab Code Quality Reporter

atheck:
This adds a built-in reporter for the GitLab Code Quality Report format.

I took the json reporter as template and adapted it to print json in the GitLab format.

Because this format does not support reporting multiple issues in one entry, I decided to create multiple entries for the 'duplicates' issue type.

### DIFF
--- a/packages/knip/src/reporters/codeclimate.ts
+++ b/packages/knip/src/reporters/codeclimate.ts
@@ -11,18 +11,29 @@ interface CodeClimateEntry {
   check_name: string;
   description: string;
   categories: string[];
-  location: {
-    path: string;
-    positions: {
-      begin: {
-        line: number;
-        column: number;
-      };
-    };
-  };
+  location: Location;
   severity: CodeClimateSeverity;
   fingerprint: string;
 }
+
+type Location = {
+  path: string;
+} & (
+  | {
+      positions: {
+        begin: {
+          line: number;
+          column: number;
+        };
+      };
+    }
+  | {
+      lines: {
+        begin: number;
+        end: number;
+      };
+    }
+);
 
 export default async ({ report, issues }: ReporterOptions) => {
   const entries: CodeClimateEntry[] = [];
@@ -44,15 +55,7 @@ export default async ({ report, issues }: ReporterOptions) => {
             check_name: getTitle(fixedType),
             description: getSymbolDescription({ symbol, parentSymbol: issue.parentSymbol }),
             categories: ['Duplication'],
-            location: {
-              path: toRelative(filePath),
-              positions: {
-                begin: {
-                  line: symbol.line ?? 0,
-                  column: symbol.col ?? 0,
-                },
-              },
-            },
+            location: createLocation(filePath, symbol.line, symbol.col),
             severity: convertSeverity(issue.severity),
             fingerprint: createFingerprint(filePath, symbol.symbol, symbol.pos),
           }))
@@ -63,15 +66,7 @@ export default async ({ report, issues }: ReporterOptions) => {
           check_name: getTitle(fixedType),
           description: getIssueDescription(issue),
           categories: ['Bug Risk'],
-          location: {
-            path: toRelative(filePath),
-            positions: {
-              begin: {
-                line: issue.line ?? 0,
-                column: issue.col ?? 0,
-              },
-            },
-          },
+          location: createLocation(filePath, issue.line, issue.col),
           severity: convertSeverity(issue.severity),
           fingerprint: createFingerprint(filePath, issue.symbol, issue.pos),
         });
@@ -112,6 +107,28 @@ function getIssueDescription({ symbol, symbols, parentSymbol }: Issue) {
 
 function getSymbolDescription({ symbol, parentSymbol }: { symbol: IssueSymbol; parentSymbol?: string }) {
   return `${symbol.symbol}${parentSymbol ? ` (${parentSymbol})` : ''}`;
+}
+
+function createLocation(filePath: string, line?: number, col?: number): Location {
+  if (col !== undefined) {
+    return {
+      path: toRelative(filePath),
+      positions: {
+        begin: {
+          line: line ?? 0,
+          column: col,
+        },
+      },
+    };
+  }
+
+  return {
+    path: toRelative(filePath),
+    lines: {
+      begin: line ?? 0,
+      end: line ?? 0,
+    },
+  };
 }
 
 function createFingerprint(filePath: string, message: string, pos?: number): string {
--- a/packages/knip/src/reporters/codeclimate.ts
+++ b/packages/knip/src/reporters/codeclimate.ts
@@ -32,18 +32,16 @@ export default async ({ report, issues }: ReporterOptions) => {
       continue;
     }
 
-    if (type === 'files') {
-      continue;
-    }
+    const fixedType = type === 'files' ? '_files' : type;
 
-    for (const issue of flatten(issues[type] as IssueRecords)) {
+    for (const issue of flatten(issues[fixedType] as IssueRecords)) {
       const { filePath } = issue;
 
-      if (type === 'duplicates' && issue.symbols) {
+      if (fixedType === 'duplicates' && issue.symbols) {
         entries.push(
           ...issue.symbols.map(symbol => ({
             type: 'issue' as const,
-            check_name: getTitle(type),
+            check_name: getTitle(fixedType),
             description: getSymbolDescription({ symbol, parentSymbol: issue.parentSymbol }),
             categories: ['Duplication'],
             location: {
@@ -62,7 +60,7 @@ export default async ({ report, issues }: ReporterOptions) => {
       } else {
         entries.push({
           type: 'issue' as const,
-          check_name: getTitle(type),
+          check_name: getTitle(fixedType),
           description: getIssueDescription(issue),
           categories: ['Bug Risk'],
           location: {
--- a/packages/knip/src/reporters/codeclimate.ts
+++ b/packages/knip/src/reporters/codeclimate.ts
@@ -1,34 +1,38 @@
 import { createHash } from 'node:crypto';
 import type { Entries } from 'type-fest';
 import type { Issue, IssueRecords, IssueSeverity, IssueSymbol, Report, ReporterOptions } from '../types/issues.js';
+import { toRelative } from '../util/path.js';
 import { getTitle } from './util.js';
 
-/**
- * Entry for GitLab Code Quality report. See: https://docs.gitlab.com/ci/testing/code_quality/#code-quality-report-format
- */
-interface CodeQualityEntry {
+type CodeClimateSeverity = 'info' | 'minor' | 'major' | 'critical' | 'blocker';
+
+interface CodeClimateEntry {
+  type: 'issue';
   check_name: string;
   description: string;
-  severity: 'info' | 'minor' | 'major' | 'critical' | 'blocker';
-  fingerprint: string;
+  categories: string[];
   location: {
     path: string;
-    lines: {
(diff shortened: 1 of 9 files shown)


packages/knip/src/reporters/gitlab.ts
```diff
@@ -0,0 +1,128 @@
...
+  description: string;
+  severity: 'info' | 'minor' | 'major' | 'critical' | 'blocker';
+  fingerprint: string;
+  location: {
+    path: string;
+    lines: {
+      begin: number;
+    };
+  };
+}
```
webpro, line 20: Looking at the specs, I would love to name it `codeclimate` while keeping it compatible for GitLab usage. Would that work for you as well?

If so, could you please add `"type": "issue"` to the interface/object?

I'm not sure about `categories` yet, so let's leave it out for now.

packages/knip/src/reporters/gitlab.ts
```diff
@@ -0,0 +1,128 @@
...
+  const hashes = new Set<string>();
+
+  for (const [type, isReportType] of Object.entries(report) as Entries<Report>) {
+    if (!isReportType) {
+      continue;
+    }
+
+    if (type === 'files' || type === '_files') {
+      continue;
+    }
```
webpro, line 33: Maybe we could leave in the unused files type?

I know it's ugly, but it'd need something like this:

```ts
type = type === 'files' ? '_files' : type;
```

And ignore `files`. No worries if you don't want/need it now, I could work it in later on.

packages/knip/src/reporters/gitlab.ts
```diff
@@ -0,0 +1,128 @@
...
+
+      if (type === 'duplicates' && issue.symbols) {
+        entries.push(
+          ...issue.symbols.map(symbol => ({
+            check_name: getTitle(type),
+            description: getSymbolDescription({ symbol, parentSymbol: issue.parentSymbol }),
+            severity: convertSeverity(issue.severity),
+            fingerprint: createFingerprint(filePath, symbol.symbol, hashes),
+            location: {
+              path: filePath,
```
webpro, line 46: According to the specs, the `filePath` should be relative, you can import the [toRelative](https://github.com/webpro-nl/knip/blob/7e22547039df179a47335c4be25736acf0e5ad33/packages/knip/src/util/path.ts#L30) helper for this.

packages/knip/src/reporters/gitlab.ts
```diff
@@ -0,0 +1,128 @@
...
+              lines: {
+                begin: symbol.line ?? 0,
```
webpro, line 48: Knip has `line` and `col` (and `pos`) available, and looks like the specs allow to include this like so:

```json
{
  "path": "path/to/file.css",
  "positions": {
    "begin": {
      "line": 3,
      "column": 10
    }
  }
}
```

The gitlab description mentions `location.positions.begin.line` so let's use that?

packages/knip/src/reporters/gitlab.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  md5.update(filePath);
+  md5.update(message);
+
+  let md5Copy = md5.copy();
+  let hash = md5Copy.digest('hex');
+
+  // To avoid hash collision, we keep generating new hashes until we get a unique one.
+  while (hashes.has(hash)) {
+    md5.update(hash);
```
webpro, line 119: Perhaps if we add `symbol` and/or `pos` (or `line` and `col`) to the data we should get unique ones (without having to keep track of hashes)?

packages/knip/src/reporters/gitlab.ts
```diff
@@ -0,0 +1,128 @@
...
+};
+
+function flatten(issues: IssueRecords): Issue[] {
+  return Object.values(issues).flatMap(Object.values);
+}
+
+function convertSeverity(severity?: IssueSeverity): CodeQualityEntry['severity'] {
+  switch (severity) {
+    case undefined:
+      return 'info';
```
webpro, line 85: Maybe instead of this initial `undefined` case we could move it to the end as the default case?

webpro (commented):
Thanks Henrik! I'm excited to add this new reporter to Knip, but I do have a few remarks.

packages/knip/src/reporters/gitlab.ts
atheck: I also add the categories array and use "Duplication" or "Bug Risk" for it.

packages/knip/src/reporters/gitlab.ts
atheck: What is the difference between "files" and "_files"?

webpro: `files` is a Set of file paths, but I've added `_files` later on so that all issue types have the same object structure for consistency/ease of use (and I can wrap this up a in next major with a breaking change).

atheck:
I pushed a new commit addressing the remarks.

packages/knip/src/reporters/codeclimate.ts
```diff
@@ -0,0 +1,127 @@
...
+            severity: convertSeverity(issue.severity),
+            fingerprint: createFingerprint(filePath, symbol.symbol, symbol.pos),
+          }))
+        );
+      } else {
+        entries.push({
+          type: 'issue' as const,
+          check_name: getTitle(type),
+          description: getIssueDescription(issue),
+          categories: ['Bug Risk'],
```
webpro, line 67: I wonder what's the best way to go about this one. While I agree "bug risk" is best option if we must choose one from https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#categories, but it doesn't really cut it. But thanks, let's roll with it as it looks we don't really have another option atm.

webpro (commented):
Thanks for the changes, looks great.

One thing I forgot to ask first time around: could you please add a test? One way to go about it is to copy https://github.com/webpro-nl/knip/blob/main/packages/knip/test/cli/reporter/cli-reporter-json.test.ts and adjust the `json` output variable.

atheck:
Yes, I'll write some tests. At the moment I have some issues running the tests. I hope to get them running.

webpro:
Feel free to AMA or hit me up on e.g. Discord. Tip: it's possible to run a single test only using `bun test tests/feat.test.ts`

webpro:
> At the moment I get an error from /bin/sh not finding the node command. But node is installed via nvm. I think this is a setup issue on my machine.

If it helps, you could do it all with only Bun (although `npm run test:node` should give no issues, you don't need to worry about getting this running properly in Node).

> Also, I cannot compare the output and expectation as strings, because of the fingerprint property. Instead I have to parse the output and compare the objects.

I think the idea of the fingerprint is that it is reproducible. You made the right choice for using md5 here, because it's deterministic and should produce the same output repeatedly given the same input. So I'd expect the `stdout` of running `knip --reporter codeclimate` to match `JSON.stringify(json)` without extra steps.

> Coming from jest, there is a function `expect.any(String)`. I need to figure out how to do this with bun.

We're using the `node:assert/strict` module for assertions, they also work in Bun.

webpro:
@atheck Almost there! A few issues around fingerprints left. I wouldn't mind wrapping up the last bits here, but happy to leave it to you as well. WDYT?

atheck:
I didn't take look into it yet, but I will. If I need help, I will ask you.

atheck:
I use `expect` and `expect.any(String)` to fix the issue with the fingerprints. But now, the tests are not correctly transformed by transform-test.js. Any ideas?

webpro:
Sorry for the hassle. While I like the richness of the `expect` matchers, so far I've tried to keep things simple and compatible with Node.js. I guess we still have some options here:

- Introduce a new assertion lib (not really in favor atm, but open to this idea)
- Or, we could run the tests in Bun only (there's a few `skipIfBun` already, we could do the opposite here)
- Or, to make things compatible with Windows, maybe add a function like this (untested):

```js
function assertButRemoveFingerprints(issues) {
  return issues.map(issue => {
    const { fingerprint, ...rest } = issue;
    assert.match(fingerprint, /[a-f0-9]{32}/);
    return rest;
  });
}
```

Not pretty, but benefit is a slightly better assertion of the `fingerprint` value

atheck:
Ok, I think the last option is the best for now. I try it.

atheck:
I didn't want to copy that function in all tests, so I decided to write a reusable function and put it into the helpers directory.

packages/knip/test/cli/reporter/cli-reporter-codeclimate4.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      categories: ['Bug Risk'],
+      location: { path: 'package.json', positions: { begin: { line: 0, column: 0 } } },
+      severity: 'major',
+    },
+    {
+      type: 'issue',
+      check_name: 'Unlisted binaries',
+      description: 'start-server',
+      categories: ['Bug Risk'],
+      location: { path: 'package.json', positions: { begin: { line: 0, column: 0 } } },
```
webpro, line 50: I think it's better to not add `positions` if we don't have those, otherwise it could be confusing (like here). Looks like the spec allows to omit it?

packages/knip/src/reporters/codeclimate.ts
```diff
@@ -0,0 +1,127 @@
...
+export default async ({ report, issues }: ReporterOptions) => {
+  const entries: CodeClimateEntry[] = [];
+
+  for (const [type, isReportType] of Object.entries(report) as Entries<Report>) {
+    if (!isReportType) {
+      continue;
+    }
+
+    if (type === 'files') {
+      continue;
```
webpro, line 36: Just noticed that the tests don't show unused files, while there should be a few instances of those. So I did some work in `main` to make the `_files` issue type consistent with the others.

Any chance you could rebase from `main` and do [something like this](https://github.com/webpro-nl/knip/blob/171c86a392bb4f0046e7e9e71a9f311fbc8ce309/packages/knip/src/reporters/symbols.ts#L56) so we'll also have unused files in the report?

webpro (commented):
Thanks a bunch for the helper, very useful! Noticed two more things when reviewing, that should get us through 🙏

packages/knip/test/cli/reporter/cli-reporter-codeclimate4.test.ts
atheck: That's true. I'll fix this.

atheck: Hm, the GitLab documentation says that `positions` or `lines` properties are required. In the Codeclimate specs I can't find anything either that it's optional.

When there is no column, I use the `lines` property instead of `positions`.

webpro: Thank you!

https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#locations:

Here it is mentioned that `path` is required, not `positions`. But indeed from the GitLab spec it sounds like it's required. Bummer. Let's roll with it 🛼

atheck:
@webpro Thank you!

webpro:
:rocket: _This pull request is included in v5.46.0. See [Release 5.46.0](https://github.com/webpro-nl/knip/releases/tag/5.46.0) for release notes._

webpro:
Thanks again Henrik (maybe interesting thread: https://bsky.app/profile/webpro.nl/post/3lkai4eozv22d)

atheck:
@webpro Thank you, very interesting.

I also need some improvement. The description should not only contain the dependency or type, but also what kind of issue this is, e.g. something like "Unused dependency foo" instead of only "foo".

Maybe we can create a new issue to improve the description and the fingerprint.

PR #985, opened 2025-03-13
Add SST plugin

BryanCrotazGivEnergy:
Added a v1 sst plugin.

Loads lambda handlers from:
/handlers/*
/lambdas/*
/src/handlers/*
/src/lambdas/*

Work left for a v2 plugin to read the sst lambda setup and explicitly load the lambda handlers as entry points

### DIFF
--- a/packages/knip/src/plugins/sst/index.ts
+++ b/packages/knip/src/plugins/sst/index.ts
@@ -1,0 +1,24 @@
+import type { IsPluginEnabled, Plugin } from "../../types/config.js";
+import { hasDependency } from "../../util/plugin.js";
+
+// link to sst docs
+
+const title = "sst";
+
+const enablers = ["sst"];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) =>
+  hasDependency(dependencies, enablers);
+
+const entry = [
+  "sst.config.{js,cjs,mjs,ts}",
+  "{handlers,lambdas}/*.{js,cjs,mjs,ts}",
+  "src/{handlers,lambdas}/*.{js,cjs,mjs,ts}",
+];
+
+export default {
+  title,
+  enablers,
+  isEnabled,
+  entry,
+} satisfies Plugin;
--- a/packages/knip/src/plugins/sst/index.ts
+++ b/packages/knip/src/plugins/sst/index.ts
@@ -12,8 +12,8 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) =>
 
 const entry = [
   "sst.config.{js,cjs,mjs,ts}",
-  "{handlers,lambdas}/*.{js,cjs,mjs,ts}",
-  "src/{handlers,lambdas}/*.{js,cjs,mjs,ts}",
+  "{handlers,lambdas}/**/*.{js,cjs,mjs,ts}",
+  "src/{handlers,lambdas}/**/*.{js,cjs,mjs,ts}",
 ];
 
 export default {
--- a/packages/knip/src/plugins/sst/types.ts
+++ b/packages/knip/src/plugins/sst/types.ts
@@ -1,0 +1,4 @@
+export type PluginConfig = {
+  plugins?: string[];
+  entryPathsOrPatterns?: string[];
+};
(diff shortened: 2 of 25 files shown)


packages/knip/src/plugins/sst/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const enablers = ["sst"];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) =>
+  hasDependency(dependencies, enablers);
+
+const entry = [
+  "sst.config.{js,cjs,mjs,ts}",
+  "{handlers,lambdas}/*.{js,cjs,mjs,ts}",
+  "src/{handlers,lambdas}/*.{js,cjs,mjs,ts}",
```
webpro, line 16: Do we have any docs/resources that indicate these could be considered defaults locations?

packages/knip/src/plugins/sst/index.ts
```diff
@@ -0,0 +1,24 @@
+import type { IsPluginEnabled, Plugin } from "../../types/config.js";
+import { hasDependency } from "../../util/plugin.js";
+
+// link to sst docs
```
webpro, line 4: Please add a relevant link to SST docs re. config.

packages/knip/test/plugins/sst.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  test('Find dependencies with the sst plugin with one handler in src/lambdas folder', async () => {
+    const { issues, counters } = await main({
+      ...baseArguments,
+      cwd: cwdOneHandlerInSrcLambdas,
+    });
+  
+//    console.log('issues', issues);
+    
+    assert.partialDeepStrictEqual(issues, {
+        unlisted: {
```
webpro, line 85: The idea that most plugins follow is to have zero issues left, so we only need to check all issue counters are zero, e.g.:

```ts
assert.deepEqual(counters, {
  ...baseCounters,
    processed: 4,
    total: 4,
})
```

In case there are still issues left we bump the relevant counter and use additional separate assertions for issues like so:

```ts
assert(issues.devDependencies['package.json']['some-unused-dep']);
```

packages/knip/test/plugins/sst.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import assert from 'node:assert/strict';
+import { main } from '../../src/index.js';
+import { resolve } from '../../src/util/path.js';
+import baseArguments from '../helpers/baseArguments.js';
+import baseCounters from '../helpers/baseCounters.js';
+
+const cwdEmptyConfig = resolve('fixtures/plugins/sst/empty-config');
+const cwdOneHandler = resolve('fixtures/plugins/sst/one-handler');
+const cwdOneHandlerInSrcLambdas = resolve('fixtures/plugins/sst/one-handler-in-lambdas');
+const cwdOneHandlerInSrcHandlers = resolve('fixtures/plugins/sst/one-handler-in-src');
```
webpro, line 11: Maybe good to move each `cwd` to each test context individually to hold them together.

packages/knip/src/plugins/sst/types.ts
```diff
@@ -0,0 +1,4 @@
+export type PluginConfig = {
+  plugins?: string[];
+  entryPathsOrPatterns?: string[];
```
webpro, line 3: This is probably coming from the template, any chance it could reflect the actual SST config interface? If there's nothing interesting from the plugin perspective it could be left empty or even removed.

packages/knip/test/plugins/sst.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+test('Find dependencies with the sst plugin with empty config', async () => {
+  const { issues, counters } = await main({
+    ...baseArguments,
+    cwd: cwdEmptyConfig,
+  });
+
+//   console.log('issues', issues);
+//   console.log('counters', counters);
```
webpro, line 20: Please remove `console.log` statements

webpro (commented):
Thanks Bryan! Happy to merge the SST plugin, but I do have a few remarks.

BryanCrotazGivEnergy:
Sounds good, I'll take a look at #1005

webpro:
Closing this in favor of #1005.

PR #988, opened 2025-03-18
fix: escape react-router route paths

rossipedia:
The `react-router` plugin currently seems to ignore route files with `(` or `)` in their path, since these are [interpreted as special characters](https://www.npmjs.com/package/fast-glob#advanced-syntax) by `micromatch` (and thus `fast-glob`).

I put together a small repo that shows the issue here: https://github.com/rossipedia/knip-optional-path-routes

The simplest fix I could think of was to just use a regex to escape any of those characters found in the loaded route file paths, since they were discovered via the file-system I think it makes sense to treat them as literal paths with any special characters having been escaped.

### DIFF
--- a/packages/knip/fixtures/plugins/react-router/app/routes.ts
+++ b/packages/knip/fixtures/plugins/react-router/app/routes.ts
@@ -4,4 +4,7 @@ export default [
     file: "routes/layout.tsx",
     children: [{ file: "./routes/another-route.tsx" }],
   },
+  {
+    file: "routes/route.(with).$special[.chars].tsx"
+  }
 ];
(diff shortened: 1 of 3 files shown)


webpro:
Thanks for the PR! That's an unfortunate issue, and the fix looks fine to me. Any chance you could add some files to the fixtures in `fixtures/plugins/react-router` to prevent future regressions? Once they're picked up the counters in the test file `react-router.test.ts` can be bumped to match.

rossipedia:
done 👍

rossipedia:
sorry for the late response here, but yeah. I don't think I'm doing anything particularly platform-specific, so I'd say skipping the test for now on Windows is the easiest (since there's already prior art here in the test suite).

webpro:
Thank you!

webpro:
:rocket: _This pull request is included in v5.48.0. See [Release 5.48.0](https://github.com/webpro-nl/knip/releases/tag/5.48.0) for release notes._

AlemTuzlak:
@rossipedia can confirm this broke it on Windows 🤣

PR #991, opened 2025-03-18
Correctly resolve git hooks directory using `git rev-parse`

### LINKED ISSUE #990: 🐛 Knip fails in a git worktree + lefthook
### Prerequisites

### Reproduction url

https://github.com/VanTanev/knip-lefthook-bug-reproduction

### Reproduction access

### Description of the issue

Reproduction is kind of tedious, because it requires cloning the repo and then creating a [git worktree branch checkout](https://git-scm.com/docs/git-worktree).

The issue is that Knip assumes the hooks directory is always `.git/hooks`, but that is not the case. In a worktree checkout, `.git` is not a folder but instead a pointer to the real folder.

VanTanev:
Fixes #990

### DIFF
--- a/packages/knip/src/util/git.ts
+++ b/packages/knip/src/util/git.ts
@@ -14,7 +14,7 @@ const hookFileNames = [
 
 const getGitHooksPath = (defaultPath = '.git/hooks') => {
   try {
-    return execSync('git config --get core.hooksPath', { encoding: 'utf8', stdio: ['pipe', 'pipe', 'ignore'] }).trim();
+    return execSync('git rev-parse --git-path hooks', { encoding: 'utf8', stdio: ['pipe', 'pipe', 'ignore'] }).trim();
   } catch (_error) {
     return defaultPath;
   }


VanTanev:
Tangential, but this doesn't work with `knip --directory ../some-folder-one-level-up` - This is because the `gitHooksPath()` function is called in a context, where the `--directory` parameter doesn't exist, and we can't pass `cwd` option to `child_process.execSync()`

webpro:
> Fixes #990

Thanks for the PR!

> There is no reasonable way to add tests for this, as far as I can figure.

We do have https://github.com/webpro-nl/knip/blob/main/packages/knip/test/plugins/lefthook-v1.test.ts + https://github.com/webpro-nl/knip/tree/main/packages/knip/fixtures/plugins/lefthook-v1 to prevent regression for regular `.git/hooks` lookups.

VanTanev:
Which now fails because it depended on the old static behavior... OK, let me look at it and see if I can come up with a test

VanTanev:
@webpro The tests are fairly ugly, but that is because git hook paths are resolved at knip import time, and do not resolve relatively to the `cwd` argument. This means the only way to set `process.cwd()` is to do `process.chdir()` before the import statement :-/

VanTanev:
Leaving aside the biome formatting issue, this probably fails because the tests do not have execution isolation.

So there is a module import that happens earlier, and the incorrect paths get cached here: https://github.com/webpro-nl/knip/blob/529a734833416ba2172b14d2936e68028b94c76f/packages/knip/src/plugins/lefthook/index.ts#L16

This is not easy to fix, unless plugin `config` variable option turns into a function that gets knip options and returns an array.

packages/knip/src/plugins/lefthook/index.ts
```diff
@@ -54,6 +50,10 @@ export default {
   title,
   enablers,
   isEnabled,
-  config,
+  get config() {
+    const gitHookPaths = getGitHookPaths();
+
+    return ['lefthook.yml', ...gitHookPaths];
+  },
```
VanTanev, line 57: This is a workaround for configs getting calculated way too early with a static `process.cwd()

webpro:
How about adding a separate Actions workflow to assert git worktrees work properly? Could that simplify things, or maybe you have other ideas?

webpro:
There don't even necessarily need to be a test runner per se, only shell commands to set up the situation, add some files perhaps (maybe prepare a fixtures folder for this, but without a test file), and run knip. If all commands (including `knip`) exit with code zero then the workflow (which itself is the test!) succeeds.

VanTanev:
Yep! Will look into doing that if not tomorrow by the end of the week. Thanks for helping me get this to a good state!

webpro:
Is there still interest in finishing this PR? No worries if not, but I'm going to clean things up a bit around here.

VanTanev:
I definitely want to push this through, as it's still a problem in my day-to-day work. It just seemed daunting every time I thought to return to it, as writing the correct tests seems quite complicated, even though the underlying fix is so straightforward: https://github.com/webpro-nl/knip/pull/991/commits/2fe0a85331ae638d43c5b51f375654469f2fb095

But please, keep it open for a while longer, and I'll make a best effort to get this fixed properly.

webpro:
Sure, no hurries, just wanted to check if it's still interesting/relevant.

VanTanev:
Hey @webpro, I added a workflow that verifies worktree behavor and can correctly fail:
https://github.com/VanTanev/knip/actions/runs/16552992919/job/46810512675

This PR is good for re-review.

Let me know if you'd like me to improve this PR in any way before it's good for merge, and thank you for the patience!

VanTanev:
Huh, wonder what causes these failures, tests were running fine locally. Will investigate.

webpro:
Think we made it! LFG 🚀

webpro:
:rocket: _This issue has been resolved in v5.78.0. See [Release 5.78.0](https://github.com/webpro-nl/knip/releases/tag/knip%405.78.0) for release notes._

PR #995, opened 2025-03-20
Use codeclimate-types

remcohaszing:
This replaces the custom CodeClimate types with the `codeclimate-types` package. Reusing existing types is less error prone that maintaining them internally.

This change uncovered that end positions were missing. This doesn’t affect GitLab, which doesn’t use this information (yet), but it might affect other tooling that expects it.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,6 +5,7 @@
       "name": "@knip/root",
       "devDependencies": {
         "@biomejs/biome": "1.9.4",
+        "codeclimate-types": "^0.3.1",
         "installed-check": "^9.3.0",
       },
     },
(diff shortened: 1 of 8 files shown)


webpro:
Thanks for the PR, Remco! Much appreciated. There was an issue (missing `isTypeOnly` for namespace type-only imports) that I've fixed in main, if you could please rebase I'm happy to merge.

remcohaszing:
AFAICT the failed CI job appears to be an actual downstream issue in TanStack/query: They should remove the now redundant dependency on `@types/eslint`

webpro:
Yes, nothing to do with any changes today. Thanks again!

webpro:
:rocket: _This pull request is included in v5.46.1. See [Release 5.46.1](https://github.com/webpro-nl/knip/releases/tag/5.46.1) for release notes._

PR #1002, opened 2025-03-23
Avoid position to calculate CodeClimate fingerprint

remcohaszing:
CodeClimate fingerprints should uniquely identify the issue. The problem with using positional information is that though it adds to the uniqueness, it significantly reduces the identification part.

To clarify, let’s say we have the following file:

```js
// file.js
export function used() {}

export function unused() {}
```

The function `unused` is detected as unused by knip. The hash is calculated from:
- `'/runner/file.js'` (filePath)
- `'unused'` (message)
- `54` (pos)

Let’s say this results in the hash `abc`.

This hash has two problematic factors:
1. It depends on an absolute path, which is system dependant. For example a GitLab users may switch to a different runner or Docker image.
2. If any characters are inserted or deleted before `unused`, the fingerprint changes.

Let’s say the user now makes a change:

```diff
  // file.js
  export function used() {}
+ // Some inserted text

  export function unused() {}
```

The new hash is calculated from:
- `'/runner/file.js'` (filePath)
- `'unused'` (message)
- `76` (pos)

Let’s say this results in the hash `def`.

Now the GitLab merge request  UI will tell the user that the first problem with `hash` is resolved. Also a new problem is introduced with a hash `def`.

By omitting the positional information, the same hash would be preserved, and the GitLab UI will tell the user the merge request didn’t affect the results.

I also added an implementation to ensure hashes are always unique in a single run. This is based on
[eslint-formatter-gitlab#48](https://gitlab.com/remcohaszing/eslint-formatter-gitlab/-/merge_requests/48).

Also noteworthy is that it’s best to avoid changing the fingerprint formula between versions if possible. That would cause GitLab to also declare all previous issues were resolved, but new similar ones are detected.

No tests were modified, because the hashes are explicitly omitted from testing.

### DIFF
--- a/packages/knip/src/reporters/codeclimate.ts
+++ b/packages/knip/src/reporters/codeclimate.ts
@@ -7,7 +7,7 @@ import { getTitle } from './util.js';
 
 export default async ({ report, issues }: ReporterOptions) => {
   const entries: codeclimate.Issue[] = [];
-  const hashes = new Set<string>()
+  const hashes = new Set<string>();
 
   for (const [type, isReportType] of Object.entries(report) as Entries<Report>) {
     if (!isReportType) {
@@ -109,23 +109,23 @@ function createLocation(filePath: string, line?: number, col?: number): codeclim
 function createFingerprint(filePath: string, message: string, hashes: Set<string>): string {
   const md5 = createHash('md5');
 
-  md5.update(filePath);
+  md5.update(toRelative(filePath));
   md5.update(message);
 
   // Create copy of hash since md5.digest() will finalize it, not allowing us to .update() again
-  let md5Tmp = md5.copy()
-  let hash = md5Tmp.digest('hex')
+  let md5Tmp = md5.copy();
+  let hash = md5Tmp.digest('hex');
 
   while (hashes.has(hash)) {
     // Hash collision. This happens if we encounter the same ESLint message in one file
     // multiple times. Keep generating new hashes until we get a unique one.
-    md5.update(hash)
+    md5.update(hash);
 
-    md5Tmp = md5.copy()
-    hash = md5Tmp.digest('hex')
+    md5Tmp = md5.copy();
+    hash = md5Tmp.digest('hex');
   }
 
-  hashes.add(hash)
+  hashes.add(hash);
 
   return md5.digest('hex');
 }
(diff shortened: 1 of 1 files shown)


packages/knip/src/reporters/codeclimate.ts
```diff
@@ -105,12 +106,26 @@ function createLocation(filePath: string, line?: number, col?: number): codeclim
...
+  while (hashes.has(hash)) {
+    // Hash collision. This happens if we encounter the same ESLint message in one file
+    // multiple times. Keep generating new hashes until we get a unique one.
+    md5.update(hash);
+
+    md5Tmp = md5.copy();
+    hash = md5Tmp.digest('hex');
+  }
+
+  hashes.add(hash);
```
remcohaszing, line 128: This should definitely be removed if you’re not worried about duplicate messages.

webpro: Yeah, don't think we can have duplicate `symbol` instances within the same file. Even less so after we'd have implemented https://github.com/webpro-nl/knip/pull/970#issuecomment-2721108208 (cc @atheck).

webpro (commented):
Thanks, great PR. Good to know about the hash preservation.

> This hash has two problematic factors:
> 
> 1. It depends on an absolute path, which is system dependant. For example a GitLab users may switch to a different runner or Docker image.

Good catch, this was an oversight on my end when reviewing.

> 2. If any characters are inserted or deleted before `unused`, the fingerprint changes.

Did not think of this, happy to go with the new approach.

> Also noteworthy is that it’s best to avoid changing the fingerprint formula between versions if possible. That would cause GitLab to also declare all previous issues were resolved, but new similar ones are detected.

Alright!

packages/knip/src/reporters/codeclimate.ts
remcohaszing: I believe you have the better knowledge on this subject :)

I imagined maybe if the user adds duplicate exports:

```js
export function unused() {}
export function unused() {}
```

or even TypeScript overloads:

```
export declare function add(a: 1, b: 1): 2
export declare function add(a: 2, b: 2): 4
export declare function add(a: number, b: number): number
export function add(a: number, b: number): number {
  return a + b
}
```

This is an edge case though, and not super relevant. I’ll remove this logic.

webpro: If such an export would be unused a single issue would be reported, same for unused dependencies in `package.json` or unlisted dependencies in a file. Members of enums and classes should be prefixed with the name of the class/enum. Same for unused exports attached to a namespace, etc.

Thanks!

webpro:
:rocket: _This pull request is included in v5.46.1. See [Release 5.46.1](https://github.com/webpro-nl/knip/releases/tag/5.46.1) for release notes._

PR #1003, opened 2025-03-23
feat: handle vitest workspace configuration

codepunkt:
The `vitest` plugin is not properly handling [workspace configurations](https://vitest.dev/guide/workspace.html).

This PR should enable that, and thus detect various things (e.g. `globalSetup`, `setupFiles`, `reporters` or `coverage.provider`) inside a `workspace` array - like in this example:

```ts
import { defineConfig } from 'vitest/config';

export default defineConfig({
  test: {
    workspace: [
      {
        test: {
          name: 'integration',
          setupFiles: './vitest.integration.setup.mjs',
        },
      },
      {
        test: {
          globalSetup: './vitest.unit.setup.ts',
          name: 'unit',
        },
      },
    ],
  },
});
```

### DIFF
--- a/packages/knip/src/plugins/vitest/index.ts
+++ b/packages/knip/src/plugins/vitest/index.ts
@@ -45,7 +45,19 @@ const findConfigDependencies = (localConfig: ViteConfig, options: PluginOptions)
   const setupFiles = [testConfig.setupFiles ?? []].flat().map(specifier => ({ ...toDeferResolve(specifier), dir }));
   const globalSetup = [testConfig.globalSetup ?? []].flat().map(specifier => ({ ...toDeferResolve(specifier), dir }));
 
-  return [...[...environments, ...reporters, ...coverage].map(id => toDependency(id)), ...setupFiles, ...globalSetup];
+  const workspaceDependencies: Input[] = [];
+  if (testConfig.workspace !== undefined) {
+    for (const workspaceConfig of testConfig.workspace) {
+      workspaceDependencies.push(...findConfigDependencies(workspaceConfig, options));
+    }
+  }
+
+  return [
+    ...[...environments, ...reporters, ...coverage].map(id => toDependency(id)),
+    ...setupFiles,
+    ...globalSetup,
+    ...workspaceDependencies,
+  ];
 };
 
 const getConfigs = async (localConfig: ViteConfigOrFn | VitestWorkspaceConfig) => {
--- a/packages/knip/test/plugins/vitest8.test.ts
+++ b/packages/knip/test/plugins/vitest8.test.ts
@@ -1,0 +1,26 @@
+import { test } from 'bun:test';
+import assert from 'node:assert/strict';
+import { main } from '../../src/index.js';
+import { resolve } from '../../src/util/path.js';
+import baseArguments from '../helpers/baseArguments.js';
+import baseCounters from '../helpers/baseCounters.js';
+
+const cwd = resolve('fixtures/plugins/vitest8');
+
+test('Find dependencies with Vitest plugin (8)', async () => {
+  const { counters, issues, report, rules } = await main({
+    ...baseArguments,
+    cwd,
+  });
+
+  assert(issues.unresolved['vitest.config.ts']['./vitest.integration.setup.mjs']);
+  assert(issues.unresolved['vitest.config.ts']['./vitest.unit.setup.ts']);
+
+  assert.deepEqual(counters, {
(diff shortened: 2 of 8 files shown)


packages/knip/test/plugins/vitest8.test.ts
```diff
@@ -0,0 +1,26 @@
...
+const cwd = resolve('fixtures/plugins/vitest8');
+
+test('Find dependencies with Vitest plugin (8)', async () => {
+  const { counters, issues, report, rules } = await main({
+    ...baseArguments,
+    cwd,
+  });
+
+  assert(issues.unresolved['vitest.config.ts']['./vitest.integration.setup.mjs']);
+  assert(issues.unresolved['vitest.config.ts']['./vitest.unit.setup.ts']);
```
webpro, line 17: Thanks for the PR! Happy to merge. If you could please `touch` those (empty) files there? Then this test file and the `counters` are tidy too 🍃

packages/knip/src/plugins/vitest/index.ts
```diff
@@ -45,7 +45,19 @@ const findConfigDependencies = (localConfig: ViteConfig, options: PluginOptions)
   const setupFiles = [testConfig.setupFiles ?? []].flat().map(specifier => ({ ...toDeferResolve(specifier), dir }));
   const globalSetup = [testConfig.globalSetup ?? []].flat().map(specifier => ({ ...toDeferResolve(specifier), dir }));
 
-  return [...[...environments, ...reporters, ...coverage].map(id => toDependency(id)), ...setupFiles, ...globalSetup];
+  const workspaceDependencies: Input[] = [];
+  if (testConfig.workspace !== undefined) {
```
webpro, line 49: Please update the types for this plugin (https://github.com/webpro-nl/knip/pull/1003/checks#step:6:18)

packages/knip/test/plugins/vitest8.test.ts
codepunkt: https://github.com/webpro-nl/knip/pull/1003/commits/c2ce2934a6a907e3ffef85d83e147b373b743019

packages/knip/src/plugins/vitest/index.ts
codepunkt: https://github.com/webpro-nl/knip/pull/1003/commits/27564a4a14a97a7103d604ff21e36fa9a8d1324c

webpro:
Thanks! Only a few lint issues left it seems.

webpro:
Here are the commands that can be used for QA and that should go green: https://github.com/webpro-nl/knip/blob/main/.github/DEVELOPMENT.md#qa

There's currently one job in Run Knip against external projects that fails, we can ignore that.

codepunkt:
`bun knip:production` wasn't found.
`bun lint` had an error due to unused variables and a small spacing issue, which I've fixed.

`bun run test` failed on the new testcase that I've added, because I'm specifically testing the non-existance of files, but you asked me to add them. I've sinced removed them, and now everything listed in the QA link you've sent (except `knip:production`) should pass.

Is there anything else I can help with?

packages/knip/test/plugins/vitest8.test.ts
codepunkt: I'm not 100% sure that I understand the counters and issues returned by `await main()`. When i `touch` the files, the test fails because they're not unresolved then.

How exactly would `touch`ing the files make this test file and `counters` tidy?

webpro: Sorry, I wasn't clear. With "tidy" I meant that the issue counters end up being zero (`0`). When the files are there (i.e. not unresolved), there are no issues left (only a number of processed/total files). Then we don't need to assert numbers for `unresolved` etc. which in my mind is "tidy".

webpro:
> `bun knip:production` wasn't found. 

Forgot to update that one, should be `bun knip --strict` now.

codepunkt:
My thanks go to you! 👍🏻

webpro:
:rocket: _This pull request is included in v5.47.0. See [Release 5.47.0](https://github.com/webpro-nl/knip/releases/tag/5.47.0) for release notes._

fshowalter:
This doesn’t account for `extends` so if you define your coverage engine as Istanbul in the root config and set `extends: true` in the workspace, Knip will think the workspaces have no  coverage definition and default to v8, leading Knip to flag it as a missing dep.

PR #1005, opened 2025-03-24
Add support for resolveFromAST in plugins

webpro:
Add support for `resolveFromAST` in plugins so we can use the source file AST directly (i.e. without loading the file and receiving the resolved `localConfig`).

Updated plugin guide: https://feat-plugin-ast.knip.pages.dev/guides/writing-a-plugin#example-4-use-the-ast-directly

New plugins:

- TanStack Router (https://github.com/webpro-nl/knip/pull/856)
- SST (https://github.com/webpro-nl/knip/pull/985)
- Starlight

Extended plugins:

- ~Astro~ (new plugin: Starlight)
- Next.js

For the existing `resolveEntryPaths` and `resolveFromConfig` functions that plugins can implement, Knip loads the tool's configuration file and passes the resulting object to this plugin function. This works well for most cases, but has some drawbacks:

- Potentially slow as it may import lots of dependencies itself
- Error-prone as the config files may depend on context such as environment variables being set
- Information may be missing once resolved i.e. the returning object might no longer contain references we want

An example `vite.config.ts`:

```ts
import { defineConfig } from 'vite';
import { TanStackRouterVite } from '@tanstack/router-plugin/vite';

export default defineConfig({
  plugins: [
    TanStackRouterVite({
      routesDirectory: './routes',
      generatedRouteTree: './gen/routeTree.gen.ts',
    }),
  ],
});
```

We want to know the values of `routesDirectory` and `generatedRouteTree`, but if we `import('vite.config.ts')` then this information is missing from the default-exported value that we pass on to `resolveEntryPaths` and `resolveFromConfig`.

The `vite.config.ts` is already part of the module graph and it's `import` statements need to be parsed anyway (so we know `vite` and `@tanstack/router-plugin/vite` are dependencies that should be listed in `package.json`), so why not make this available to plugins as well?

This way, plugins can create a custom function to extract the information, in this case the (new) TanStack Router plugin can return non-default entry paths (`./routes/**/*.{ts,tsx}`) and a non-default path for `generatedRouteTree` so users don't need to configure such `entry` patterns manually in `knip.json`. This was previously impossible!

A downside is that the "DSL" here is the TS API, which is powerful but also tedious. Another downside is that the plugins can start using `typescript` dependency which has a large API surface and reduces their compactness and simplicity, which in turn requires more governance.

Here's example code to do what we just described:

```ts
import ts from 'typescript';

const CONFIG_KEYS = new Set([
  'routeFilePrefix',
  'routeFileIgnorePrefix',
  'routeFileIgnorePattern',
  'routesDirectory',
  'generatedRouteTree',
]);

const FUNCTIONS = new Set(['TanStackRouterVite', 'TanStackRouterRspack', 'TanStackRouterWebpack']);

export const getCustomConfig = (sourceFile: ts.SourceFile) => {
  const config: Record<string, string> = {};

  function visit(node: ts.Node) {
    if (ts.isCallExpression(node)) {
      const callee = node.expression;
      if (ts.isIdentifier(callee) && FUNCTIONS.has(callee.text)) {
        const firstArg = node.arguments[0];
        if (ts.isObjectLiteralExpression(firstArg)) {
          for (const prop of firstArg.properties) {
            if (ts.isPropertyAssignment(prop) && ts.isIdentifier(prop.name)) {
              if (CONFIG_KEYS.has(prop.name.text)) {
                if (ts.isStringLiteral(prop.initializer)) {
                  config[prop.name.text] = prop.initializer.text;
                }
              }
            }
          }
        }
      }
    }

    ts.forEachChild(node, visit);
  }

  visit(sourceFile);

  return config;
};
```

Also see the other `resolveFromAST.ts` files in this PR for more examples.

There are no utilities for this type of querying yet, but I can imagine we'll often want to do things like "get the string values of certain literal object properties" so that could be a next step once we figure out our needs better.

Edit: added a few helpers already 😇 so also the example above can be implemented more concise now

Edit 2: also see https://github.com/webpro-nl/knip/pull/1009 that has the same plugins using [tsquery](https://github.com/phenomnomnominal/tsquery)


packages/knip/src/WorkspaceWorker.ts
```diff
@@ -331,7 +342,13 @@ export class WorkspaceWorker {
         getInputsFromScripts: createGetInputsFromScripts(containingFilePath),
       };
 
-      const configEntryPaths: Input[] = [];
+      if (config.entry) {
+        const toInput = isProduction && plugin.production && plugin.production.length > 0 ? toProductionEntry : toEntry;
+        for (const input of config.entry.map(id => toInput(id))) addInput(input);
+      } else if ((!plugin.resolveEntryPaths && !plugin.resolveFromAST) || remainingConfigFilePaths.length === 0) {
+        if (plugin.entry) for (const input of plugin.entry.map(id => toEntry(id))) addInput(input);
+        if (plugin.production) for (const input of plugin.production.map(id => toProductionEntry(id))) addInput(input);
```
tryggvigy, line 350: a complete nit but I guess these maps could be `.map(toInput)` etc.

packages/knip/src/WorkspaceWorker.ts
```diff
@@ -341,45 +358,47 @@ export class WorkspaceWorker {
...
+          if (!loadedConfig) loadedConfig = await loadConfigForPlugin(configFilePath, plugin, opts, pluginName);
+          if (loadedConfig) {
+            const inputs = await plugin.resolveEntryPaths(loadedConfig, opts);
+            for (const input of inputs) addInput(input, configFilePath);
+          }
+        }
+
+        if (plugin.resolveConfig) {
+          if (!loadedConfig) loadedConfig = await loadConfigForPlugin(configFilePath, plugin, opts, pluginName);
+          if (loadedConfig) {
```
tryggvigy, line 374: Seeing that both `plugin.resolveEntryPaths` and `plugin.resolveConfig` attempt to define `loadedConfig` makes me think they are mutually exclusive properties (don't make sense when specified together in `plugin`). Is that true? If so, maybe it could be helpful to model the type of `plugin` to eliminate the invalid state where both can be passed on the same time. I'm just spitballing here really and could be completely off, also this could not matter at all :) I kind of lack the bigger picture. I'll comment on everything where I have some thought. Feel free to disregard any  of my comments

packages/knip/src/plugins/sst/resolveFromAST.ts
```diff
@@ -0,0 +1,64 @@
...
+  options,
+  getSourceFile,
+  getReferencedInternalFilePath
+) => {
+  const entries: Input[] = [];
+  const importMap = new Map<string, string>();
+
+  // Unbound sourceFile.imports isn't available, maybe create helper
+  for (const statement of sourceFile.statements) {
+    if (ts.isImportDeclaration(statement)) {
```
tryggvigy, line 17: Not sure if we want to care about `require` call expressions here?

packages/knip/fixtures/plugin-ast/vite.config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default defineConfig({
+  plugins: [
+    TanStackRouterVite({
+      target: 'react',
+      autoCodeSplitting: true,
+      routesDirectory: './routes',
+      generatedRouteTree: './gen/routeTree.gen.ts',
+      // routeFileIgnorePrefix: '-',
+      // quoteStyle: 'single',
```
tryggvigy, line 13: I'm reviewing commit by commit. Maybe this was removed in a later commit?

packages/knip/src/plugins/astro/index.ts
```diff
@@ -19,6 +23,11 @@ const production = [
   'src/actions/index.{js,ts}',
 ];
 
+const resolveFromAST: ResolveFromAST = (sourceFile: ts.SourceFile) => {
+  const componentPaths = getComponentPathsFromSourceFile(sourceFile);
+  return [...entry.map(toEntry), ...[...production, ...componentPaths].map(id => toProductionEntry(id))];
```
tryggvigy, line 28: A few places where we are spreading into one array. Any chance we could have duplicate paths in the resulting array? I wonder if it's worth deduping this before returning.

packages/knip/src/plugins/next/index.ts
```diff
@@ -9,28 +11,40 @@ const enablers = ['next'];
...
 ];
 
-const production = [
-  ...productionEntryFilePatternsWithoutSrc,
-  ...productionEntryFilePatternsWithoutSrc.map(pattern => `src/${pattern}`),
-];
+const getEntryFilePatterns = (pageExtensions = defaultPageExtensions) => {
+  const pages = pageExtensions.map(ext => `pages/**/*.${ext}`);
+  const patterns = [...productionEntryFilePatterns, ...pages];
+  return [...patterns, ...patterns.map(pattern => `src/${pattern}`)];
```
tryggvigy, line 31: same here, wondering about if duplicates could be a problem

packages/knip/src/WorkspaceWorker.ts
webpro: `resolveEntryPaths` and `resolveConfig` can both be implemented as they server different purposes. They might get merged one day now that we have all the "inputs" (`onEntry` etc), but that'd be for another day. So what I was trying to do here is "load if need, and skip if not".

packages/knip/src/plugins/sst/resolveFromAST.ts
webpro: This is specific to SST, so probably not.

packages/knip/src/plugins/sst/resolveFromAST.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  ts.forEachChild(sourceFile, addHandlerSpecifiers); // Only for v3?
+
+  function visit(node: ts.Node) {
+    if (
+      ts.isCallExpression(node) &&
+      ts.isPropertyAccessExpression(node.expression) &&
+      node.expression.name.text === 'stack'
+    ) {
```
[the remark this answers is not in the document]
webpro: I've actually looked at tsquery, but what we need is a few matchers, so at this point I'm thinking to create those as we go (indeed, without pulling in another dep).

packages/knip/fixtures/plugin-ast/vite.config.ts
webpro: Nice catch, thanks

packages/knip/src/plugins/astro/index.ts
webpro: Eventually things should be deduped before the actual globbing or reporting takes place.

packages/knip/src/plugins/next/index.ts
webpro: This is all local, I don't think there's a huge risk of dupes.

packages/knip/src/plugins/sst/resolveFromAST.ts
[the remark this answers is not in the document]
webpro: Thanks for sharing those, I often use https://ts-ast-viewer.com

packages/knip/src/plugins/sst/resolveFromAST.ts
BryanCrotazGivEnergy: well SST can be used with `require` or with `import`...

webpro: One does not generally use CommmonJS syntax in TS files (`sst.config.ts`). Or is e.g. `sst.config.cjs` also supported?

Would be easy to add, though :)

webpro:
:rocket: _This pull request is included in v5.48.0. See [Release 5.48.0](https://github.com/webpro-nl/knip/releases/tag/5.48.0) for release notes._

PR #1011, opened 2025-03-28
Add support for markdownlint-cli2

pkuczynski:

### DIFF
--- a/packages/knip/src/plugins/markdownlint-cli2/index.ts
+++ b/packages/knip/src/plugins/markdownlint-cli2/index.ts
@@ -1,0 +1,29 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import { toDependency } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { MarkdownlintCli2Config } from './types.js';
+
+// https://github.com/DavidAnson/markdownlint-cli2
+
+const title = 'markdownlint-cli2';
+
+const enablers = ['markdownlint-cli2'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['.markdownlint-cli2.{jsonc,yaml,cjs,mjs}'];
+
+const resolveConfig: ResolveConfig<MarkdownlintCli2Config> = (config, options) => {
+  const formatters = config?.outputFormatters
+      ? [...config.outputFormatters.map(([formatter]) => formatter)] : [];
+
+  return [...formatters].map(id => toDependency(id));
+};
+
+export default {
+  title,
+  enablers,
+  isEnabled,
+  config,
+  resolveConfig,
+} satisfies Plugin;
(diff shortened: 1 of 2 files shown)


webpro:
Thanks Piotr! Much appreciated. Please add a test with a fixture to clarify what the plugin does and to prevent future regressions. Each and every existing Knip plugin has this.

Also see https://knip.dev/guides/writing-a-plugin#create-a-new-plugin

pkuczynski:
> Please add a test with a fixture to clarify what the plugin does and to prevent future regressions. Each and every existing Knip plugin has this.

Sure thing! Will try to do it by end of this week...

pkuczynski:
hey @webpro I made a first attempt to write a test, but does not seem to be working as expected. Can you maybe have a look an tell me what I am doing wrong?

1. My intention was to check it there is a plugin mentioned in the config, but misisng in dependencies.
2. Later on I would like to test exactly opposite case.
3. Then maybe add a test to check if config file exists but no reference in package.json (however the tool can be installed via brew or gha and run without being mentioned in package.json - so not sure if such test makes sense?)

I pushed my changes in the PR. Looking forward to hear your feedback...

webpro:
> hey @webpro I made a first attempt to write a test, but does not seem to be working as expected. Can you maybe have a look an tell me what I am doing wrong?

Thank you for working on this, feel free to ask away!

> 1. My intention was to check it there is a plugin mentioned in the config, but misisng in dependencies.

Everything looks good at the moment. So `markdownlint-cli2` is the `"lint:md"` script in `package.json`. This should result in the lint issue "missing binary" which is in the `issues` object as expected.

The `"markdownlint-rule-relative-links"` specifier is not unresolved, because the plugin is not enabled/running at all. The plugin is configured to be enabled if the `"markdownlint-cli2"` is listed in either dependencies or devDependencies.

> 2. Later on I would like to test exactly opposite case.

That should result in the lint issue "unused devDependency" which should be in `issues.devDependencies`.

By the way, you'd have to make a new fixture directory for this (e.g. `markdownlint-cli2-2` 😃)

> 3. Then maybe add a test to check if config file exists but no reference in package.json (however the tool can be installed via brew or gha and run without being mentioned in package.json - so not sure if such test makes sense?)

If this situation happens, I'd expect the config file to be reported as unused (in `issues.files`)

> I pushed my changes in the PR. Looking forward to hear your feedback...

One workflow I like, if you're using VS Code (I guess you could configure the same on others), run the debugger from a test file like `markdownlint-cli2.test.ts`. Press `F5` and the "Debug Bun test" launch config will run and you can set breakpoints anywhere, output will go to the in-IDE terminal.

webpro:
Is there still interest in finishing this PR? No worries if not, but I'm going to clean things up a bit around here.

pkuczynski:
> Is there still interest in finishing this PR? No worries if not, but I'm going to clean things up a bit around here.

There is! Let me have a look this week. If I won't make it, I will close it...

webpro:
Sure, no hurries, just wanted to check if it's still interesting/relevant.

webpro:
Closing this issue as part of a general cleanup to keep this project sustainable and optimize my time working on it. If you think this is inappropriate or if there is no workaround and you feel stuck, feel free to [open a new issue](./new/choose). Thanks for your understanding.

PR #1027, opened 2025-04-10
Mark simple-git-hooks as used dependency when config exists

azat-io:
Hello!

I get an error when using knip:

```
Unused devDependencies (1)
simple-git-hooks  package.json:66:6
```

The plugin does not seem to be working correctly at the moment. This PR marks the simple-git-hooks dependency as used when the config exists.

### DIFF
--- a/packages/knip/src/plugins/simple-git-hooks/index.ts
+++ b/packages/knip/src/plugins/simple-git-hooks/index.ts
@@ -1,5 +1,6 @@
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
 import type { Input } from '../../util/input.js';
+import { toDependency } from '../../util/input.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { SimpleGitHooksConfig } from './types.js';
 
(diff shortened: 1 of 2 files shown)


webpro:
The idea of plugins is that they're enabled if a certain dependency is listed in `package.json`, but the dependency itself should also be referenced somewhere. This tool seems to be an exception, there's only `npx simple-git-hooks` to install the hooks once.

So your solution is the way to go indeed, any chance you could also remove the assertion from the test?

webpro:
No, sorry I wasn't clear. I meant to ask if you could remove this line:

https://github.com/webpro-nl/knip/blob/b39ac44651f662cdd545d854011d107ab260baa2/packages/knip/test/plugins/simple-git-hooks.test.ts#L18

Since after we've marked `simple-git-hooks` as a used dependency in this PR, it is no longer an unused dependency in [this package.json](https://github.com/webpro-nl/knip/blob/b39ac44651f662cdd545d854011d107ab260baa2/packages/knip/fixtures/plugins/simple-git-hooks/package.json#L4) so we need to reflect that in the test to make it pass.

webpro:
Thanks Azat! 🙏

webpro:
:rocket: _This pull request is included in v5.50.4. See [Release 5.50.4](https://github.com/webpro-nl/knip/releases/tag/5.50.4) for release notes._

PR #1036, opened 2025-04-18
Set env via cross-env

elapshynov:
Adding `cross-env` to set `NX_DAEMON=false` for nx integration


webpro:
Thanks for the pull request, @elapshynov. This is documentation and the snippet is an example of how the known issue could be dealt with. Not everyone has `cross-env` globally installed on their machine so I think this solution introduces more issues than it fixes.

PR #1037, opened 2025-04-18
Add support for AWS CDK

thoroc:
# Add support for [aws-cdk](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib-readme.html)

* Created new plugin using the cli provided
* documented the `deps` and `devDeps`
* written basic tests for it

I need help on how to get `knip` to find the dependencies in the most likely path (`bin`, `cdk`, `lib`) and to test for that. I also need to check for the `cdk.json` and the use of `ts-node` in there (although it is already part of the dependencies check.

Could you give me some pointers to resolve the above issues, and let me know if I am missing anything. 

Ran the test for the new plugin using `bun` v1.2.9

### DIFF
--- a/packages/knip/src/plugins/aws-cdk/index.ts
+++ b/packages/knip/src/plugins/aws-cdk/index.ts
@@ -1,0 +1,30 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import { toDeferResolve } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { PluginConfig } from './types.js';
+
+// https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib-readme.html
+
+const title = 'aws-cdk';
+
+const enablers = ['aws-cdk'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ["cdk.json"];
+
+const production: string[] = [];
+
+const resolveConfig: ResolveConfig<PluginConfig> = async config => {
+  const inputs = config?.plugins ?? [];
+  return [...inputs].map(toDeferResolve);
+};
+
+export default {
+  title,
+  enablers,
+  isEnabled,
+  config,
+  production,
+  resolveConfig,
+} satisfies Plugin;
--- a/packages/knip/src/plugins/aws-cdk/index.ts
+++ b/packages/knip/src/plugins/aws-cdk/index.ts
@@ -1,7 +1,5 @@
-import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
-import { toDeferResolve } from '../../util/input.js';
+import type { IsPluginEnabled, Plugin } from '../../types/config.js';
 import { hasDependency } from '../../util/plugin.js';
-import type { PluginConfig } from './types.js';
 
 // https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib-readme.html
 
@@ -13,12 +11,12 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
 
 const config: string[] = ["cdk.json"];
 
-const production: string[] = [];
+const production: string[] = [
+  '{src/,cdk/,}bin/**/*.{js,ts}',
+  '{src/,cdk/,}lib/**/*.{js,ts}',
+];
 
-const resolveConfig: ResolveConfig<PluginConfig> = async config => {
-  const inputs = config?.plugins ?? [];
-  return [...inputs].map(toDeferResolve);
-};
+
 
 export default {
   title,
@@ -26,5 +24,4 @@ export default {
   isEnabled,
   config,
   production,
-  resolveConfig,
 } satisfies Plugin;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,13399 @@
+{
+  "name": "@knip/root",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@knip/root",
+      "version": "0.0.0",
+      "license": "ISC",
+      "workspaces": [
+        "packages/*"
+      ],
+      "devDependencies": {
+        "@biomejs/biome": "1.9.4",
+        "installed-check": "^9.3.0"
+      }
+    },
+    "node_modules/@astro-community/astro-embed-youtube": {
+      "version": "0.5.6",
+      "resolved": "https://registry.npmjs.org/@astro-community/astro-embed-youtube/-/astro-embed-youtube-0.5.6.tgz",
+      "integrity": "sha512-/mRfCl/eTBUz0kmjD1psOy0qoDDBorVp0QumUacjFcIkBullYtbeFQ2ZGZ+3N/tA6cR/OIyzr2QA4dQXlY6USg==",
+      "license": "MIT",
+      "dependencies": {
+        "lite-youtube-embed": "^0.3.3"
+      },
+      "peerDependencies": {
+        "astro": "^2.0.0 || ^3.0.0-beta || ^4.0.0-beta || ^5.0.0-beta"
+      }
+    },
+    "node_modules/@astrojs/check": {
+      "version": "0.9.4",
+      "resolved": "https://registry.npmjs.org/@astrojs/check/-/check-0.9.4.tgz",
+      "integrity": "sha512-IOheHwCtpUfvogHHsvu0AbeRZEnjJg3MopdLddkJE70mULItS/Vh37BHcI00mcOJcH1vhD3odbpvWokpxam7xA==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@astrojs/language-server": "^2.15.0",
+        "chokidar": "^4.0.1",
+        "kleur": "^4.1.5",
+        "yargs": "^17.7.2"
+      },
+      "bin": {
+        "astro-check": "dist/bin.js"
+      },
+      "peerDependencies": {
+        "typescript": "^5.0.0"
+      }
+    },
+    "node_modules/@astrojs/compiler": {
+      "version": "2.11.0",
+      "resolved": "https://registry.npmjs.org/@astrojs/compiler/-/compiler-2.11.0.tgz",
+      "integrity": "sha512-zZOO7i+JhojO8qmlyR/URui6LyfHJY6m+L9nwyX5GiKD78YoRaZ5tzz6X0fkl+5bD3uwlDHayf6Oe8Fu36RKNg==",
+      "license": "MIT"
+    },
+    "node_modules/@astrojs/internal-helpers": {
+      "version": "0.6.1",
+      "resolved": "https://registry.npmjs.org/@astrojs/internal-helpers/-/internal-helpers-0.6.1.tgz",
+      "integrity": "sha512-l5Pqf6uZu31aG+3Lv8nl/3s4DbUzdlxTWDof4pEpto6GUJNhhCbelVi9dEyurOVyqaelwmS9oSyOWOENSfgo9A==",
+      "license": "MIT"
+    },
+    "node_modules/@astrojs/language-server": {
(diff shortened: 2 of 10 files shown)


webpro:
> # Add support for [aws-cdk](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib-readme.html)
> 
> * Created new plugin using the cli provided
> * documented the `deps` and `devDeps`
> * written basic tests for it

Great start!

> I need help on how to get `knip` to find the dependencies in the most likely path (`bin`, `cdk`, `lib`)

I see `production` entry patterns so matching files will be added as entry files. Their file contents will be analyzed and may contain internal and external imports, and the latter will be marked as dependencies. You'd need to add such files with import statements to the fixtures folder add matching locations. Does that help?

> and to test for that.

The external dependencies can be added to `package.json` in the fixture folder so they won't be reported as unlisted.

> I also need to check for the `cdk.json` and the use of `ts-node` in there (although it is already part of the dependencies check.

The `config` file pattern is there, `resolveConfig` can be implemented and exported as part of the plugin and the contents of `cdk.json` will sent as the first argument.

thoroc:
* I have added source files to the fixtures for what I would normally expect to see in such a project under what I'd consider the normal place
* I have also added the `cdk.json` with values I am familiar with

currently the tests are reporting the following:

```sh
bun test v1.2.9 (9a329c04)

test/plugins/aws-cdk.test.ts:
{
  files: Set {},
  dependencies: {},
  devDependencies: {
    "package.json": {
      "aws-cdk": [Object ...],
    },
  },
  optionalPeerDependencies: {},
  unlisted: {
    "cdk.json": {
      cdk: [Object ...],
    },
  },
  binaries: {},
  unresolved: {},
  exports: {},
  nsExports: {},
  types: {},
  nsTypes: {},
  enumMembers: {},
  classMembers: {},
  duplicates: {},
  _files: {},
}
✓ Find dependencies with the aws-cdk plugin [97.81ms]

 1 pass
 0 fail
Ran 1 tests across 1 files. [589.00ms]
```

packages/knip/src/plugins/aws-cdk/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ["cdk.json"];
+
+const resolveConfig: ResolveConfig<AwsCdkConfig> = async config => {
+  if (!config) return [];
+
+  const app = config.app.split(" ")[0];
+  const watch = config.watch?.include ?? [];
```
webpro, line 21: My guess would be that the `watch` values aren't necessarily dependencies, nor entry files. Maybe we should ignore those. Maybe there are other options that represent entry points?

packages/knip/src/plugins/aws-cdk/index.ts
```diff
@@ -0,0 +1,38 @@
...
+const enablers = ['aws-cdk'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ["cdk.json"];
+
+const resolveConfig: ResolveConfig<AwsCdkConfig> = async config => {
+  if (!config) return [];
+
+  const app = config.app.split(" ")[0];
```
webpro, line 20: The `app` option looks like a command, we should use `getInputsFromScripts` for this.

This function is available on the second `options` argument that `ResolveConfig` is invoked with.

Look for `getInputsFromScripts` in the plugins folder for other usage examples.

packages/knip/src/plugins/aws-cdk/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  const watch = config.watch?.include ?? [];
+  const context = Object.keys(config.context ?? {}).map(key => key.split("/")[0]) ?? [];
+  return compact([app, ...watch, ...context]).map(id => toDependency(id));
```
webpro, line 23: I like that we're using `compact` here and already convert string to so-called "inputs" using `toDependency`. However, even though the `context` items are such dependencies indeed, returned items could also be entry points. You'll see what I mean when using `getInputsFromScripts` - its return value could contain various types of inputs.

webpro (commented):
OK this is great progress, thanks for working on this!

thoroc:
Thank you the pointers. I'll have a look asap.

packages/knip/src/plugins/aws-cdk/index.ts
thoroc: Addressed in df97e056d37cad6787c8c2aad8465d5adf9700c2. Just a question though: what is the `knownBinsOnly` option for?

packages/knip/src/plugins/aws-cdk/index.ts
thoroc: Are you meaning that I should be checking for the type of `id` in the `map` function? I saw typescript complaining about mismatched type and used something I saw somewhere else in the codebase:

```typescript
compact([...app, ...watch, ...context]).map(id => (typeof id === 'string' ? toDependency(id) : id));
```

packages/knip/src/plugins/aws-cdk/index.ts
thoroc: To the best of my knowledge, the `watch.include` entries are file that should form the codebase for the IaC when using AWS CDK. From the name, it would indicate those are source files being **watched** for changes when running `cdk watch` which would allow an immediate changes of the generated CloudFormation template.

packages/knip/src/plugins/aws-cdk/index.ts
webpro: Yes, something like that, e.g. the return value of `getInputsFromScripts` is input, not strings.

packages/knip/src/plugins/aws-cdk/index.ts
webpro: By default, binaries are expected to be shipped alongside the dependencies in `node_modules/.bin` - it's a setting that allows to be less strict and accept "any" binary name in CI environments—which are usually provisioned with unknown binaries—to prevent false positives for "unlisted binaries". However, the variable should def be improved.

packages/knip/src/plugins/aws-cdk/index.ts
webpro: If we would add the `watch.include` patterns, then they should be returned as `project` files (not `entry` files), i.e. `toInput('bin/**/*.ts')`. Adding "all" files as entry files would have two disadvantages: by default, unused exports of entry files are not reported, and Knip wouldn't be able to find unused files (as entry files are inherently used files).

However, the default `project` pattern is `**/*.{js,ts}`, so adding more specific patterns from a plugin wouldn't be effective.

My suggestion would be to ignore the `watch.include` patterns and see whether the defaults Knip finds are sufficient, then we could add default entry patterns in the plugin if there are any, and perhaps there are specific entry patterns to be found in aws-cdk configuration files.

thoroc: Ah ok that makes sense. I'll remove that 'watch' from the returned list.

thoroc: Addressed in 7efb03dbe2d9175efed5647b28610bbd0f2be98b

thoroc:
Where is the documentation for the plugin generated btw? I couldn't see where to add the meagre info that would allow someone to use/enable it.

webpro:
> Where is the documentation for the plugin generated btw? I couldn't see where to add the meagre info that would allow someone to use/enable it.

The script to generate plugin docs is here: https://github.com/webpro-nl/knip/blob/main/packages/docs/scripts/generate-plugin-docs.ts

Not sure when CloudFlare generates/deploys new docs for PRs. Maybe only for me as it did in e.g. https://github.com/webpro-nl/knip/pull/1029. But existing ones are all accessible from https://knip.dev/reference/plugins.

webpro:
> @webpro Not sure what's the best approach to resolve the test failing. My understanding is that the test are failing because I have put a potential alternative path to the regular expected `aws cdk` project; The fixtures only have one type of directory struct (`bin` + `lib`).
> 
> Should we consider this is not wanted, and revise it if people are coming back with the request for flexibility in defining the IaC path to the source code? Or do I need to create fixtures to get this test going?

Not sure I understand the question. But if you would add `ts-node` package in `node_modules` (exactly like `aws-cdk` package) then that would fix both the `binaries/ts-node` and `devDepencies/ts-node` () issues.

The other unused devDependencies could also be removed from `package.json` as they're not referenced?

thoroc:
Just updated the branch and fix the tests:

```sh
bun test v1.2.9 (9a329c04)

test/plugins/aws-cdk.test.ts:
{
  files: Set {},
  dependencies: {},
  devDependencies: {
    "package.json": {
      "aws-cdk": [Object ...],
      jest: [Object ...],
      "ts-jest": [Object ...],
      "ts-node": [Object ...],
    },
  },
  optionalPeerDependencies: {},
  unlisted: {},
  binaries: {
    "cdk.json": {
      "ts-node": [Object ...],
    },
  },
  unresolved: {},
  exports: {},
  nsExports: {},
  types: {},
  nsTypes: {},
  enumMembers: {},
  classMembers: {},
  duplicates: {},
  _files: {},
}
✓ Find dependencies with the aws-cdk plugin [119.78ms]

 1 pass
 0 fail
Ran 1 tests across 1 files. [587.00ms]
```

How do I run all the tests to ensure we are good right now?

thoroc:
> Not sure I understand the question

I was referring to the possibility that the code base where `knip` would be used could have a different directory structure than what `aws cdk` is advising (ie: `lib`/`bin` directories to hold all the IaC). 

I don't think it would make much of a difference right now.

webpro:
> How do I run all the tests to ensure we are good right now?

I often use this, from the `packages/knip` directory:

```
bun run build && bun run test  # OR
bun run test:smoke             # skip fs-related tests (should be enough for this PR)
```

Could you please remove the `package-lock.json`?

Do you know of a public repo we could use to test drive this plugin? To me it seems we have some good fixtures, but an actual repository would be great to verify we'd not be missing out on anything major.

thoroc:
Apologies - I have not forgotten about this piece of work.

---

I have just updates the branch and running the tests, there seems to be a good few errors now.

I'll hunt down someone's public repo. All the example I have at hand are work related and have sensitive information about the company's setup.

thoroc:
I have just re-ran the test as there was a change in the import:

```shell
bun test test/plugins/aws-cdk.test.ts
bun test v1.2.15 (df017990)

test/plugins/aws-cdk.test.ts:
{
  files: Set(1) {
    "/Users/thomas.roche/Projects/github/webpro-nl/knip/packages/knip/fixtures/plugins/aws-cdk/bin/infra.ts",
  },
  dependencies: {},
  devDependencies: {
    "package.json": {
      "aws-cdk": [Object ...],
      jest: [Object ...],
      "ts-jest": [Object ...],
      "ts-node": [Object ...],
    },
  },
  optionalPeerDependencies: {},
  unlisted: {},
  binaries: {
    "cdk.json": {
      "ts-node": [Object ...],
    },
  },
  unresolved: {},
  exports: {},
  nsExports: {},
  types: {},
  nsTypes: {},
  enumMembers: {},
  classMembers: {},
  duplicates: {},
  _files: {
    "fixtures/plugins/aws-cdk/bin/infra.ts": [
      [Object ...]
    ],
  },
}
✓ Find dependencies with the aws-cdk plugin [128.66ms]

 1 pass
 0 fail
Ran 1 tests across 1 files. [822.00ms]
```

webpro:
@thoroc just added a commit to assist a bit with the `entry` and `production` options

> Let me know how I could run my locally modified version to test the change.

Please see https://github.com/webpro-nl/knip/blob/main/.github/DEVELOPMENT.md

thoroc:
Just checking in; What are we missing here?

webpro:
> Just checking in; What are we missing here?

Afaic there has not been actual exercises on public repositories yet:

> Which seems to be a good candidate. Let me know how I could run my locally modified version to test the change.

Were you able to run your local version of Knip with [the dev guide](https://github.com/webpro-nl/knip/blob/main/.github/DEVELOPMENT.md)?

webpro:
Is there still interest in wrapping this up?

webpro:

PR #1041, opened 2025-04-20
Ignore imports inside tagged template literals

azat-io:
Hello!

I am currently experiencing a problem when using Knip.

The issue is that I have code samples in my code with imports that are string and not real imports.

Check out the example:
https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/docs/components/Demo/DemoBlock.astro

When I use imports inside a string, Knip tries to analyze them.

```ts
const myCodeExample = dedent`
  import Button from '~/components/Button'
`
```

Errors:

```
Unresolved imports
~/components/Button
```

This PR fixes the issue by ignoring imports inside tagged template literals, which are often used for code examples rather than actual imports.

### DIFF
--- a/packages/knip/fixtures/tagged-template-literal/index.astro
+++ b/packages/knip/fixtures/tagged-template-literal/index.astro
@@ -1,0 +1,11 @@
+---
+import dedent from 'dedent';
+import Component from './Component.astro';
+
+const dedented = dedent`
+  import Button from './Button'
+  import type { FC } from 'react'
+`;
+---
+
+<Component>
--- a/packages/knip/fixtures/tagged-template-literal/index.ts
+++ b/packages/knip/fixtures/tagged-template-literal/index.ts
@@ -1,0 +1,10 @@
+import 'astro';
+import dedent from 'dedent';
+import './app.ts';
+
+const dedented = dedent`
+  import Button from './Button'
+  import type { FC } from 'react'
+`;
+
+dedented;
--- a/packages/knip/fixtures/tagged-template-literal/package.json
+++ b/packages/knip/fixtures/tagged-template-literal/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "@fixtures/tagged-template-literal",
+  "dependencies": {
+    "astro": "*",
+    "dedent": "*"
+  }
+}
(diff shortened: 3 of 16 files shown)


webpro:
Thanks for the PR, Asat! I like the idea of ignoring imports and exports in such tagged templates.

The thing that bothers me a little bit is that we're traversing down the tree and then going back up to see if we're inside a template literal, in most visitors. Maybe I'm exaggerating the performance cost of this, walking up parents is not expensive. But perhaps we can try to see if we optimize and can bail out early, similar to `isTopLevel` here: https://github.com/webpro-nl/knip/blob/4fe4099f1ee24c58883e2388362a540fcb981127/packages/knip/src/typescript/get-imports-and-exports.ts#L289

Or do the `isInsideStringLiteral` check there, except for the `visitors.script` block.

Because we need to make sure not to bail out early for [script visitors](https://github.com/webpro-nl/knip/tree/main/packages/knip/src/typescript/visitors/scripts) that handle scripts inside code like this:

```ts
$`run script.ts`
$sync`run script.ts`
```

Wouldn't it be enough to check for `ts.isTaggedTemplateExpression` only?

And something else: we're creating the `isTopLevel` boolean (in the code I link to above), while the `dedent` example you're giving shows that the import statements are not top level. So maybe we should look into improving the `isTopLevel` guard which might fix the whole issue as well?

It may help to add a few tests to verify those assumptions and prevent future regressions.

azat-io:
Thanks for your feedback! I've implemented the changes as suggested.

azat-io:
Are you suggesting to leave the changes in the Astro compiler (`taggedTemplateMatcher`) and remove all other `isInsideStringLiteral` checks?

webpro:
> Are you suggesting to leave the changes in the Astro compiler (`taggedTemplateMatcher`) and remove all other `isInsideStringLiteral` checks?

Yes, that's what it comes down to :)

webpro:
Thank you very much, Azat! What a ride 😅

webpro:
:rocket: _This pull request is included in v5.51.0. See [Release 5.51.0](https://github.com/webpro-nl/knip/releases/tag/5.51.0) for release notes._

PR #1047, opened 2025-04-22
add relay plugin

PatrykWalach:
I'd like to add a [relay](https://relay.dev/) plugin. There are a lot of ways relay can be used so it's a bit complicated.

## How relay works
1. When using relay developers use special template literals in their code
```tsx
graphql`/* ... */`
```
2. For each of these `relay-compiler` generates artifacts in `artifactDirectory`.

3. There's a babel transform that replaces literals with imports to corresponding artifacts.

This makes it impossible for `knip` to detect this relationship by default.

## `artifactDirectory`
There's a relation between a template literal and the corresponding artifact. When `knip` detects file with literal is unused, developer can delete it and the artifact will disappear as well. 

- [x] This led me to conclusion, that **all artifacts should be entrypoints**.

## enablers
- Some people use artifacts with fetch client other than `react-relay` so this is not a good enabler.
- `react-compiler` is also not a good enabler because some people don't don't use the babel transform and import artifacts directly, in that case marking all artifacts as entrypoints can lead to false positives.

- [x] I think it's **safest to use list of common transforms as `enablers`** ( `babel-transform` and `vite-transform`)
- [ ] Detecting transform in nextjs https://nextjs.org/docs/architecture/nextjs-compiler#relay (I might skip this)
- [x] Also give an option to **enable this plugin manually in case any other transform is used**. It's a simple transform so it's common to use a custom transform.

## `customScalarTypes`
The relay config can include `customScalarTypes` - pointing to typescript `types`. It depends on the generated artifacts which types are used and which not.

**We can do two things here**:
1. Treat all types as entrypoints.
2. Leave them be. This will mark unused files as intended, even though these might be needed for future artifacts.

In my project I'd prefer option nr. 1. Fortunately, in relay config there's a property `requireCustomScalarTypes`, 
- [x] when set to true, `customScalarTypes` needs to be exhaustive so we should use strategy nr. 1, 
- [x] otherwise we can assume user cares about a subset of the types and we can mark the unused ones.  

## No `artifactDirectory`
- [x] There's an option `artifactDirectory` is not defined, in this case the artifacts are generated relative to literals, in that case we use wildcard to find entrypoints `**/__generated__/*`


packages/knip/src/plugins/relay/index.ts
```diff
@@ -0,0 +1,49 @@
...
+      : [];
+
+    return [toEntry(path.join(artifactDirectory, '**'))].concat(scalars.map(toDeferResolve));
+  });
+};
+
+const args = {
+  binaries: ['relay-compiler'],
+  // config: // first positional argument is config file
+  // (eg. relay-compiler ./myconfig.js)
```
PatrykWalach, line 39: Is there an easy way to do this?
Do I have to use `args` to transform `relay-compiler ./myconfig.js` into `relay-compiler -c ./myconfig.js` and add `config: true`?

webpro: > Is there an easy way to do this?

No, it's an edge case that a config file is a positional argument

> Do I have to use `args` to transform `relay-compiler ./myconfig.js` into `relay-compiler -c ./myconfig.js` and add `config: true`?

Yes. But you could `import parseArgs from 'minimist'` and then grab `parsedArgs._[0]` for a bit more robust way to go about it.

webpro:
Thanks for the PR. Not 100% sure where we are at. I feel like you're asking for cq ready to get some feedback? If that's true then I think it will help if you add some fixtures cq example (config) files for me to better understand how to go about things.

PatrykWalach:
I have two non working fixtures right now, the first one should have no errors, the second should report `Unused.ts` as unused

webpro:
Looking great :)

webpro:
Btw a workflow I like in VS Code is to press F5/"Debug Bun test" when in a test file

packages/knip/src/plugins/relay/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const config: string[] = ['relay.config.json', 'relay.config.js'];
+
+const resolveEntryPaths: ResolveEntryPaths<RelayConfig> = async config => {
+  const projects = 'projects' in config ? Object.values(config.projects) : [config];
+
+  return projects.flatMap(project => {
+    const artifactDirectory = project.artifactDirectory;
+
+    if (artifactDirectory == null) return [];
```
webpro, line 25: Maybe we can fall back to using the default `__generated__`?

packages/knip/src/plugins/relay/index.ts
```diff
@@ -0,0 +1,60 @@
...
+        : []
+    );
+
+    return inputs.concat(scalars);
+  });
+};
+
+const args = {
+  binaries: ['relay-compiler'],
+  args: (args: string[]) => ['-c', parseArgs(args)._[0]], // first positional argument is config file
```
webpro, line 49: This means we can ignore the other arguments, right? Fine by me.

webpro (commented):
Great stuff! Thanks a bunch. Almost done I think? :)

packages/knip/src/plugins/relay/index.ts
PatrykWalach: Should be fine

packages/knip/src/plugins/relay/index.ts
PatrykWalach: I'll try `./**/__generated__/*`

PatrykWalach: done

packages/knip/src/plugins/relay/index.ts
```diff
@@ -0,0 +1,60 @@
...
+import { toDeferResolve, toEntry } from '../../util/input.js';
+import { isInternal, join } from '../../util/path.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { RelayConfig } from './types.js';
+// https://relay.dev/docs/next/guides/compiler/#configuration
+// https://github.com/facebook/relay/blob/main/compiler/crates/relay-compiler/relay-compiler-config-schema.json
+
+const title = 'Relay';
+
+const enablers = ['vite-plugin-relay', '@swc/plugin-relay', 'babel-plugin-relay'];
```
webpro, line 12:
```suggestion
const enablers = ['react-relay', 'vite-plugin-relay', '@swc/plugin-relay', 'babel-plugin-relay'];
```

> Some people use artifacts with fetch client other than `react-relay` so this is not a good enabler.

Why is `react-relay` not a good enabler? If that dependency is listed, the plugin should be enabled, right? In what situation is that not the case?

packages/knip/src/plugins/relay/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const isEnabled: IsPluginEnabled = ({ dependencies, config }) =>
+  hasDependency(dependencies, enablers) || 'relay' in config;
```
webpro, line 15:
```suggestion
const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
```

Plugin typically don't need this, and I think I like this consistency (seeing a few instances in other plugins now I'm planning to remove). Users can still enable the plugin using `relay: true`.

packages/knip/src/plugins/relay/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const resolveEntryPaths: ResolveEntryPaths<RelayConfig> = async config => {
+  const projects = 'projects' in config ? Object.values(config.projects) : [config];
+
+  return projects.flatMap(project => {
+    const artifactDirectory = project.artifactDirectory;
+
+    if (artifactDirectory == null) return [toEntry('**/__generated__/*')];
+
+    const inputs = [toEntry(join(artifactDirectory, '**'))];
```
webpro, line 27: I think `toEntry` should be `toProductionEntry` so they're also included in `--production` mode

packages/knip/src/plugins/relay/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    if (project.requireCustomScalarTypes !== true) {
+      return inputs;
+    }
+
+    const scalars = Object.values(project.customScalarTypes ?? {}).flatMap(customScalarType =>
+      typeof customScalarType === 'object'
+        ? [
+            isInternal(customScalarType.path)
+              ? toEntry(join(artifactDirectory, customScalarType.path))
```
webpro, line 37: I think `toEntry` should be `toProductionEntry` so they're also included in `--production` mode

webpro (commented):
Checked out the relay docs, and it seems `react-relay` would be an essential enabler, no?

packages/knip/src/plugins/relay/index.ts
PatrykWalach: 1. Adding all `*.graphql.ts` files would cause issues if user named his own files `*.graphql.ts`
2. The documentation u are linking leads to `rescript` Lang version of relay, their compiler is adjusted to work with `rescript`. In either case the object might not have a `path` prop, it can be a simple string representing a typescript type `number` | `Record<...,  ...>`. I accounted for that, they are filtered out. In the case of rescript they use a string to access module `ModuleName.type`, this shouldn't cause any issues.
3. Ignoring custom scalars all together is an interesting idea. I had them added to knip ignore in my project and just assumed plugin should handle this for me.

Not adding them as entrypoints would require user to explicitly ignore these which can be useful information.

For me this would just be more annoying and increase iteration time, for example a PR will remove a field - now the type is unused, knip will fail in ci, file has to be added to ignore. I'd add all the scalars to ignore/entrypoint manually anyways, and then there's a chance these will get out of sync. The reason I enabled `requireCustomScalarTypes` is so I have these types ready and I can select/remove fields and don't worry about anything missing.

webpro: 1. Sorry I wasn't clear, what I meant was what we're already doing: `toEntry(join(artifactDirectory, '**')` - it was more of a prerequisite prior to the actual proposal to ignore the scalar paths.
2. Regardless of Rescript, if the `path` prop isn't there then `toDeferResolve(customScalarType.path)` means `toDeferResolve(undefined)` - which might eventually not be an issue, but isn't great either?
3. This might not fit well in your workflow, but reporting unused files etc. is what Knip does. Compare this to dependencies in `package.json` or named exports of a file. The idea is that the unused ones are reported (not to keep them "just in case"). Is there maybe a way this could fit in your workflow better e.g. by naming conventions that you can then `ignore` or something?

PatrykWalach: I like what you are proposing, even with the option enabled the user will be warned about unused files and can make their decision to ignore them or delete. 

I already have my files ignored so it's not a big deal for me.

PatrykWalach: done this as well as other suggestions

webpro:
Thanks a lot for the PR and for bearing with me, Patryk!

PR #1051, opened 2025-04-23
Add dotenvx plugin

robinvdvleuten:
Added a plugin to detect the use of dotenvx in any script (https://dotenvx.com/).

### DIFF
--- a/packages/knip/fixtures/plugins/dotenvx/package.json
+++ b/packages/knip/fixtures/plugins/dotenvx/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "@fixtures/dotenvx",
+  "version": "*",
+  "scripts": {
+    "printenv": "dotenvx get --all",
+    "start": "dotenvx run --strict -- node index.js"
+  }
+}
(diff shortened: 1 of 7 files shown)


webpro:
Thanks for the PR, Robin!

If we look at this test: https://github.com/webpro-nl/knip/blob/1b8acc7909166311a3261459a5ec277db7a9c267/packages/knip/test/util/get-inputs-from-scripts.test.ts#L244

It was passed (before this PR), because of this array: https://github.com/webpro-nl/knip/blob/1b8acc7909166311a3261459a5ec277db7a9c267/packages/knip/src/binaries/fallback.ts#L10

E.g. this script:

```
dotenvx run --strict -- node entry.js
```

Would add `entry.js` to the analysis as entry file.

So now I wonder what was the reason for you to add this plugin, what use case did we not cover? This might very well be reason to add the plugin, but I can't see it yet :)

robinvdvleuten:
@webpro I think I created this plugin because of my misinterpretation of Knip's output. We have a "dev" script in our package.json using both `dotenvx` and `tsx`. When Knip is run in default mode, no error is given. But when production mode is used, `tsx` is mentioned as unused binary. That made me think that Knip wasn't correctly resolving the dotenvx script. But it is, otherwise tsx was already mentioned in the default mode as unused binary.

PR #1053, opened 2025-04-23
Mark tanstack-router entries as production

Sheraff:
Following discord discussion: https://discord.com/channels/1143209786612125696/1143209787140620310/threads/1364557464711856148

The new `tanstack-router` plugin automatically adds new entry files. As far as I can tell, all those files should be considered "production entries" (none of them are test, config, storybook, ...)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
+  "npm.packageManager": "bun",
   "[javascript]": {
     "editor.defaultFormatter": "biomejs.biome"
   },
--- a/packages/knip/test/plugins/tanstack-router.test.ts
+++ b/packages/knip/test/plugins/tanstack-router.test.ts
@@ -11,6 +11,7 @@ test('Find dependencies with the tanstack-router plugin', async () => {
   const { issues, counters } = await main({
     ...baseArguments,
     cwd,
+    isProduction: true,
   });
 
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
+  "npm.packageManager": "bun",
```
Sheraff, line 2: This kind of has nothing to do with the rest of the PR, but I used a github codespace to work on the PR, and this is the vscode setting that is used to setup the codespace with the proper package manager.

packages/knip/test/plugins/tanstack-router.test.ts
```diff
@@ -11,14 +11,16 @@ test('Find dependencies with the tanstack-router plugin', async () => {
   const { issues, counters } = await main({
     ...baseArguments,
     cwd,
+    isProduction: true,
   });
 
   assert(issues.files.has(join(cwd, 'routes/-unused.ts')));
```
Sheraff, line 17: Without `isProduction: true`, the test wasn't passing, `issues.files` was an empty `Set`. This makes me really unsure my fix is correct...

webpro:
Thanks for the PR! They should indeed all be production entries. Please duplicate the test with only one having `isProduction: true`. The test passes locally, so maybe there's something off in codespace.

webpro:
Let's close this in favor of deleting the plugin. Thanks for helping me out!

PR #1055, opened 2025-04-26
feat: add Docusaurus plugin

cylewaitforit:
This adds a plugin for [Docusaurus](https://docusaurus.io/). 

Issue: #1056

### Support for
- [X] Plugins, Themes, and Presets
  - [X] Module Shorthand
- [X] Default entry for homepage, docs, blog
- [X] Sidebar files
- [x] Special scope import aliases
  - [X] [`@site`](https://docusaurus.io/docs/next/markdown-features/react#:~:text=The-,%40site,-alias%20points%20to)
  - [x] [`@docusaurus`](https://github.com/facebook/docusaurus/blob/fef6c4c8cb22a45fa4d3dff016618a3d10fdd552/jest.config.mjs#L70)
  - [x]  [Theme aliases](https://docusaurus.io/docs/advanced/client#theme-aliases) 
    - [x] `@theme`
    - [x] `@theme-original`
    - [x] `@theme-init`

### DIFF
--- a/packages/knip/src/plugins/docusaurus/index.ts
+++ b/packages/knip/src/plugins/docusaurus/index.ts
@@ -1,7 +1,7 @@
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
 import { toDeferResolve } from '../../util/input.js';
 import { hasDependency } from '../../util/plugin.js';
-import type { DocusaurusConfig } from './types.js';
+import type { ConfigItem, DocusaurusConfig } from './types.js';
 
 // https://docusaurus.io/docs/configuration
 
@@ -17,10 +17,20 @@ const entry: string[] = [];
 
 const production: string[] = [];
 
-const resolveConfig: ResolveConfig<DocusaurusConfig> = async config => {
-  const themes = config?.themes ?? [];
+const resolveConfigItem = (item: ConfigItem): string | null => {
+  if (!item) return null;
+  if (Array.isArray(item)) return item[0];
+  return item;
+};
 
-  return [themes].flat().map(toDeferResolve);
+const resolveConfig: ResolveConfig<DocusaurusConfig> = async config => {
+  const themes = (config?.themes ?? []).map(resolveConfigItem);
+  const plugins = (config?.plugins ?? []).map(resolveConfigItem);
+  const presets = (config?.presets ?? []).map(resolveConfigItem);
+
+  return [...themes, ...plugins, ...presets]
+    .filter((item): item is string => typeof item === 'string')
+    .map(toDeferResolve);
 };
 
 export default {
--- a/packages/knip/src/plugins/docusaurus/index.ts
+++ b/packages/knip/src/plugins/docusaurus/index.ts
@@ -14,9 +14,7 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
 
 const config: string[] = ['docusaurus.config.{js,ts}'];
 
-const entry: string[] = [];
-
-const production: string[] = [];
+const production: string[] = ['src/pages/index.js', '{blog,docs}/**/*.mdx'];
 
 const resolveConfig: ResolveConfig<DocusaurusConfig> = async config => {
   const themes = (config?.themes ?? []).map(item => resolveConfigItem(item, 'theme'));
@@ -33,7 +31,6 @@ export default {
   enablers,
   isEnabled,
   config,
-  entry,
   production,
   resolveConfig,
 } satisfies Plugin;
--- a/packages/knip/src/plugins/docusaurus/index.ts
+++ b/packages/knip/src/plugins/docusaurus/index.ts
@@ -14,7 +14,13 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
 
 const config: string[] = ['docusaurus.config.{js,ts}'];
 
-const production: string[] = ['src/pages/index.{js,ts,jsx,tsx}', '{blog,docs}/**/*.mdx'];
+const production: string[] = [
+  'src/pages/index.js',
+  'src/pages/index.ts',
+  'src/pages/index.jsx',
+  'src/pages/index.tsx',
+  '{blog,docs}/**/*.mdx',
+];
 
 const resolveConfig: ResolveConfig<DocusaurusConfig> = async config => {
   const themes = (config?.themes ?? []).map(item => resolveConfigItem(item, 'theme'));
--- a/packages/knip/src/plugins/docusaurus/index.ts
+++ b/packages/knip/src/plugins/docusaurus/index.ts
@@ -42,7 +42,7 @@ const resolveConfig: ResolveConfig<DocusaurusConfig> = async (config, options) =
     // Ignore aliases for @docusaurus/theme-classic/lib/theme/ https://docusaurus.io/docs/advanced/client#theme-aliases
     ...(hasClassicTheme ? [toIgnore('(@theme|@theme-init|@theme-original)/*', 'dependencies')] : []),
     // Ignore aliases for @docusaurus/core/lib/client/exports/ https://docusaurus.io/docs/docusaurus-core
-    toIgnore(`@docusaurus/(${[...CORE_CLIENT_API].join('|')})`, 'dependencies'),
+    toIgnore(`@docusaurus/(${CORE_CLIENT_API.join('|')})`, 'dependencies'),
     ...production.map(id => toProductionEntry(id)),
     ...resolveResults.flatMap(result => result.dependencies).map(dep => toDeferResolve(dep)),
     ...resolveResults.flatMap(result => result.entries ?? []).map(entry => toDeferResolveEntry(entry)),
--- a/packages/knip/src/plugins/docusaurus/index.ts
+++ b/packages/knip/src/plugins/docusaurus/index.ts
@@ -1,5 +1,5 @@
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
-import { toDeferResolve, toDeferResolveEntry } from '../../util/input.js';
+import { toAlias, toDeferResolve, toDeferResolveEntry, toProductionEntry } from '../../util/input.js';
 import { hasDependency } from '../../util/plugin.js';
 import { resolveConfigItem } from './helpers.js';
 import type { DocusaurusConfig, ResolveResult } from './types.js';
@@ -26,6 +26,7 @@ const resolveConfig: ResolveConfig<DocusaurusConfig> = async config => {
   );
 
   return [
+    toAlias('@site/*', './*'),
     ...resolveResults.flatMap(result => result.dependencies).map(dep => toDeferResolve(dep)),
     ...resolveResults.flatMap(result => result.entries ?? []).map(entry => toDeferResolveEntry(entry)),
   ];
--- a/packages/knip/src/plugins/docusaurus/index.ts
+++ b/packages/knip/src/plugins/docusaurus/index.ts
@@ -23,7 +23,7 @@ const config = ['docusaurus.config.{js,ts}'];
 
 const production = ['src/pages/index.{js,ts,jsx,tsx}', '{blog,docs}/**/*.mdx'];
 
-const resolveConfig: ResolveConfig<DocusaurusConfig> = async config => {
+const resolveConfig: ResolveConfig<DocusaurusConfig> = async (config, options) => {
   const themes = (config?.themes ?? []).map(item => resolveConfigItem(item, 'theme'));
   const plugins = (config?.plugins ?? []).map(item => resolveConfigItem(item, 'plugin'));
   const presets = (config?.presets ?? []).map(item => resolveConfigItem(item, 'preset'));
@@ -32,10 +32,14 @@ const resolveConfig: ResolveConfig<DocusaurusConfig> = async config => {
     (result): result is ResolveResult => result !== null
   );
 
+  const hasClassicTheme =
+    options.manifest.dependencies?.['@docusaurus/theme-classic'] ||
+    options.manifest.dependencies?.['@docusaurus/preset-classic'];
+
   return [
     toAlias('@site/*', './*'),
     toDependency('@docusaurus/module-type-aliases', { optional: true }),
-    toIgnore('@theme/(Heading|Layout)', 'dependencies'),
+    ...(hasClassicTheme ? [toIgnore('@theme/(Heading|Layout)', 'dependencies')] : []),
     toIgnore('@docusaurus/(Link|useDocusaurusContext)', 'dependencies'),
     ...production.map(id => toProductionEntry(id)),
     ...resolveResults.flatMap(result => result.dependencies).map(dep => toDeferResolve(dep)),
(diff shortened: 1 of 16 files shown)


packages/knip/src/plugins/docusaurus/types.ts
```diff
@@ -0,0 +1,5 @@
+export type DocusaurusConfig = {
+  title: string;
+  url: string;
+  themes?: string[];
```
webpro, line 4: There seem to be more types than we currently support. I'm looking at https://docusaurus.io/docs/configuration#theme-plugin-and-preset-configurations which shows there's `themes`, `plugins`, `presets`, and their types are not just strings `string[]`. Plugins don't have to be 100% complete initially to be (very) useful, but at least the implementation should not fail e.g. if any of the `themes` items we're iterating over is not a string. Would be great to also represent that in the fixture(s). I'm leaving it up to you how much you'd like to cover, but for any Docusauris config option we support we should support it as good as possible (otherwise for existing users Knip might surprise and throw exceptions).

webpro (commented):
Thanks for the PR! It's a great start, but I do have a remark.

packages/knip/src/plugins/docusaurus/types.ts
cylewaitforit: I updated the fixture to add a preset and a couple plugins(one defined and one undefined).  

I guess I'm not sure what the exact right place to stop for the initial version of the plugin.

For example does it need to support [module shorthands](https://docusaurus.io/docs/using-plugins#module-shorthands) out of the gate?  It's not something that I am personally using but others might be and I'm sure there are other examples of that in the docs and code. Or is it better to deliver minimal and see what is brought up that other users need?

webpro: Thanks for the updated fixtures and fixes, that's exactly what I meant 👍

For any config options that we do support, it should at least not result in false positives. So what we could do here is not support module shorthands, but also make sure not to have false positives reported (otherwise there could be surprises/breaking changes for existing users). You could add examples to the fixture and then make sure to ignore those in the plugin. Leaving it up to you for now: support or ignore module shorthands. If you don't fancy implementing that, I'd be happy to, just let me know (I can add commits to your branch).

cylewaitforit: @webpro I took a swing at supporting module shorthand and added to the fixtures to cover it. 

I don't love the way it is checking against the set of first party modules for the ones that should resolve with the `@docusaurus` scope. So if you have any ideas on how to handle that better I'd be open to other options.

webpro: Thanks, much appreciated. Tbh I think it's readable and I'd be happy to merge it. Often I put such functions in a separate `helpers.ts` so the plugin itself is even more digestible.

If you didn't already, one could dive into the Docusaurus source code, or we could ask one of their maintainers, but afaic really not yet necessary in this PR.

packages/knip/src/plugins/docusaurus/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+const enablers = ['@docusaurus/core'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['docusaurus.config.{js,ts}'];
+
+const entry: string[] = [];
+
+const production: string[] = [];
```
webpro, line 18: Are there no default entry or production patterns? Or only the [global default ones](https://knip.dev/explanations/entry-files#default-entry-file-patterns)?

Otherwise, empty arrays can be omitted.

webpro (approved):

packages/knip/src/plugins/docusaurus/types.ts
cylewaitforit: The helpers file did make it easier to read so I've moved the functions there.

cylewaitforit:
> LGTM! Just to be 100% sure, I think this one's finished, right?

@webpro Good question, I used `create-docusaurus@latest` to [spin up a repo](https://github.com/cylewaitforit/knip-docusaurus) when working on the default entry files to make sure the fixtures added would hit what most docusaurus sites would have. 

Pulling on that thread lead to a few other pieces that I'm looking into.

Running knip on the repo with the default entry changes currently returns

```sh
Unused files (2)
sidebars.js                               
src/components/HomepageFeatures/index.js
Unused devDependencies (1)
@docusaurus/module-type-aliases  package.json:26:6
Unlisted dependencies (5)
@docusaurus/Link                       src/pages/index.js
@docusaurus/useDocusaurusContext       src/pages/index.js
@theme/Layout                          src/pages/index.js
@site/src/components/HomepageFeatures  src/pages/index.js
@theme/Heading                         src/pages/index.js
```

From what I can find `@docusaurus/module-type-aliases` is actually unused. ✂️🎉


The sidebars.js is in the config for the ['@docusaurus/plugin-content-docs'](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#ex-config) as the sidebarPath. 

- Could just set a default for `./sidebars.{js,ts}` in the entries (I think this is probably not prefered as anyone using a different path would need to override the knip config)
- Could get the sidebarPath when resolving the config and add it. (This is the path I think I'll go down.)

As shown in the unlisted dependencies, docusaurus is using certain scopes( `@docusaurus`, `@theme`, and `@site`) for some things that aren't packages. Finding these in their docs hasn't been the most straightforward so it may require me spelunking the code more.

- `@docusaurus` appears to be used for some components in `@docusuarus/core`. https://docusaurus.io/docs/docusaurus-core#link
  -  Can't just blanket show anything with the `@docusaurus` scope as used because that would lead to other packages in the `@docusaurus` scope as used when they might be unused. Maybe to start this again is matching against a set of what comes from `@docusaurus/core`.
-  `@site` from the outset seems like it is used for site components that are defined in the project. So it might be the most straightforward of the three.
   - Figuring out this one would resolve the other unused file `src/components/HomepageFeatures/index.js`
- `@theme` appears to be used for components in the default theme but I think could also be overridden by customized project components if the are "swizzled". https://docusaurus.io/docs/swizzling

I'll move the PR back to draft and add a checklist to the top description while working through these. Certainly let me know if not delivering on all of those doesn't seem like the deal breaker I currently think it is.

packages/knip/src/plugins/docusaurus/index.ts
cylewaitforit: Added default entry for [homepage, blog, and docs.](https://github.com/webpro-nl/knip/pull/1055/commits/1f2bf0a4dbe63caa6f069d0e668caeefea3841e3)

webpro:
Thanks a bunch for running this against the docusaurus starter/app.

The custom scoped package specifiers seem indeed blocking, otherwise we'd get the false positives you've listed which are kinda annoying to have to add to `ignoreDependencies` (even though users can do `ignoreDependencies: [/@(docusaurus|site|theme)\/.*/]`)

There's no generic way to handle this in Knip, other than adding those to the global `IGNORED_DEPENDENCIES`, which would be inaccurate because, as you mention, some _are_ regular dependencies.

An idea I had was to go down the AST route (a bit like the [SST plugin is doing](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/sst/resolveFromAST.ts#L9)). But realized we'd have to deal with such imports in MDX files well, which gets ugly quickly.

So at this point I'm thinking the only way to do this is to (1) figure out the internal/custom/swizzled "dependencies" we need to ignore/skip (which is already cumbersome enough it seems) and then (2) have a way for the plugin to tell Knip/core about this list (which I could add).

Maybe (2) could be done in core by adding a new property to existing inputs/functions like `toDeferResolve` and `toDependency` or a new input such as `toIgnoreDependency`, that later down the road can be handed over to the `DependencyDeputy` to be ignored (essentially the same result as users manually doing `ignoreDependencies`).

But before I start implementing (2) we should be confident we can make (1) happen. So... WDYT? :)

cylewaitforit:
@webpro I think that would work for the `@docusaurus` alias and likely for `@theme`(and related `@theme-internal` and `@theme-init` that I found today). Those default component lists should be fairly static and adding them to the Ignored Dependencies initially seems right. 

I could be wrong, but it feels like `@site` would need to be handled differently since it is an [alias for the website's directory. ](https://docusaurus.io/docs/next/markdown-features/react#:~:text=The-,%40site,-alias%20points%20to). Knip needs to know that '@site/src/components/HomepageFeatures' is the same as 'src/components/HomepageFeatures/index.js' for the latter to not be reported as an unused file.

webpro:
> @webpro I think that would work for the `@docusaurus` alias and likely for `@theme`(and related `@theme-internal` and `@theme-init` that I found today). Those default component lists should be fairly static and adding them to the Ignored Dependencies initially seems right.

Alright, that's great.

> I could be wrong, but it feels like `@site` would need to be handled differently since it is an [alias for the website's directory. ](https://docusaurus.io/docs/next/markdown-features/react#:~:text=The-,%40site,-alias%20points%20to). Knip needs to know that '@site/src/components/HomepageFeatures' is the same as 'src/components/HomepageFeatures/index.js' for the latter to not be reported as an unused file.

Sounds like something that belongs locally in `tsconfig.json#compilerOptions.paths`.

If it isn't we could consider and try supporting that from plugins as well. That might help solve other cases as well:

- https://knip.dev/features/compilers#svelte
- https://knip.dev/guides/handling-issues#external-aliased-imports

Maybe something like a new input type so we can do e.g. `toCompilerOptionsPath()` (just thinking out loud here).

webpro:
Not sure if that may in turn require the `--include-libs` flag on user's end (https://knip.dev/guides/handling-issues#external-libraries)

webpro:
Took the liberty to add `toAlias` and return `production` entries, one step close to our goal :)

Tried a few things, but still can't think of anything better than adding the option to return `toIgnore()` patterns from the plugin.

cylewaitforit:
> Took the liberty to add toAlias and return production entries, one step close to our goal :)

That's great! The site scope changes seem to be working solid. 

I had seen the production entry change in the knip changelogs and figured that was why the tests in this branch started throwing but I hadn't had a chance to dig. That looks much simpler to understand for plugins going forward. I'll take a look at adding back the small [ts,jsx,tsx to production entry](https://github.com/webpro-nl/knip/pull/1055/commits/3d4f1537a99b0176fe745da51c8b93478d6c703b) change now that things are green again.

webpro:
> > Took the liberty to add toAlias and return production entries, one step close to our goal :)
> 
> That's great! The site scope changes seem to be working solid.

Awesome.

> I had seen the production entry change in the knip changelogs and figured that was why the tests in this branch started throwing but I hadn't had a chance to dig. That looks much simpler to understand for plugins going forward.

Added a purple box about this: https://knip.dev/guides/writing-a-plugin#7-entry-and-production so hopefully at least it will be a bit more clear to the next developer.

> I'll take a look at adding back the small [ts,jsx,tsx to production entry](https://github.com/webpro-nl/knip/pull/1055/commits/3d4f1537a99b0176fe745da51c8b93478d6c703b) change now that things are green again.

Thanks, we will get there!

webpro:
I've added the `toIgnore` input type and using this in the plugin seems to work well.

Rebased this PR on top of that (which in turn was (re)based on an updated `main` already).

The version of knip that features `toIgnore` can also be installed separately using e.g.:

```sh
https://pkg.pr.new/knip@fd6ec64
```

There's currently one caveat to `toIgnore` though: if we ignore e.g. `@theme/(Heading|Layout)` and nothing of the unlisted dependencies matches with that (i.e. the user does not use any of those) then a configuration hint about the unused option will appear.

Happy to hear your thoughts about it. Feel free to hack on this further 😇

cylewaitforit:
> There's currently one caveat to `toIgnore` though: if we ignore e.g. `@theme/(Heading|Layout)` and nothing of the unlisted dependencies matches with that (i.e. the user does not use any of those) then a configuration hint about the unused option will appear.

Does it make a difference to the configuration hint if the `toIgnore` were to be paired with a `toAlias`? 

e.g.
```
toAlias('@theme/Heading', '@docusaurus/theme-classic/lib/theme/Heading'),
toIgnore('@theme/Heading)', 'dependencies'),
```

cylewaitforit:
Or maybe the @theme alias should only be enabled if `@docusaurus/theme-classic`(or '@docusaurus/preset-classic`) was an enabler. That kind of feels like it starts getting into the Astro/Starlight two plugins realm which I don't know if it feels right for this.

webpro:
At first sight, feels better to keep logic like this within the plugin.

In `resolveConfig`, we do have access to `options.manifest` (the `package.json` object), so we could e.g. check if something is listed in `dependencies` or `devDependencies` (or if there's something specific in `scripts`, etc).

cylewaitforit:
@webpro I think this plugin might be ready to see sunlight. Besides the `toIgnore` changes merging into main I can't think of any blockers.

packages/knip/src/plugins/docusaurus/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    options.manifest.dependencies?.['@docusaurus/preset-classic'];
+
+  return [
+    toAlias('@site/*', './*'),
+    toDependency('@docusaurus/module-type-aliases', { optional: true }),
+    // Ignore aliases for @docusaurus/theme-classic/lib/theme/ https://docusaurus.io/docs/advanced/client#theme-aliases
+    ...(hasClassicTheme ? [toIgnore('(@theme|@theme-init|@theme-original)/*', 'dependencies')] : []),
+    // Ignore aliases for @docusaurus/core/lib/client/exports/ https://docusaurus.io/docs/docusaurus-core
+    toIgnore(
+      '@docusaurus/(BrowserOnly|ComponentCreator|constants|ExecutionEnvironment|Head|Interpolate|isInternalUrl|Link|Noop|renderRoutes|router|Translate|useBaseUrl|useBrokenLinks|useDocusaurusContext|useGlobalData|useIsBrowser|useIsomorphicLayoutEffect|useRouteContext)',
```
webpro, line 46: Maybe we could create an array somewhere inside this `resolveConfig` scope to make this a but more readable/maintainable, such as `["BrowserOnly", ...].join('|')`) so that it won't be such a long line.

Or maybe based on an array similar to `FIRST_PARTY_MODULES`

webpro (approved):
Just one nit, but overall this is pretty great. Glad we ended up with something that has improvements to core and feels manageable even though the underlying tool is heavy on virtual modules (a concept that might require a better solution but we can keep exploring).

packages/knip/src/plugins/docusaurus/index.ts
cylewaitforit: Commit to add it here: https://github.com/webpro-nl/knip/pull/1055/commits/2e48a79fff7515a7e73055a8989ef291e8f13194

webpro:
Ran the plugin against two repos:

- https://github.com/facebook/react-native-website
- https://github.com/algolia/docsearch

This turned out to require a few adaptions, and support for `configureWebpack` turned out to require a refactoring as `findWebpackDependenciesFromConfig` is an `async` function.

Hope you don't mind! WDYT, does it still work well on your own project(s)?

cylewaitforit:
> Hope you don't mind! WDYT, does it still work well on your own project(s)?

Don't mind at all! Glad it will work in even more uses. It still works just as well with the projects I've tested it on.

webpro:
Thank you so much, @cylewaitforit! Great job 🚀

webpro:
:rocket: _This pull request is included in v5.56.0. See [Release 5.56.0](https://github.com/webpro-nl/knip/releases/tag/5.56.0) for release notes._

PR #1060, opened 2025-04-29
Mark commitlint/cli as used dependency when config exists

azat-io:
When using Knip with a `.commitlintrc` or `.commitlintrc.json` configuration file, it incorrectly reports `@commitlint/cli` as an unused dependency, despite the presence of a valid configuration file.

The issue occurs because Knip establishes a connection between configuration files and plugin names rather than directly with the package names listed in enablers.

For most plugins like `stylelint` and `prettier`, this works fine because:

1. The plugin name matches the actual NPM package name
2. The configuration file gets properly associated with the package

However, for commitlint there's a mismatch:

1. The plugin name is `commitlint`
2. The actual package name is `@commitlint/cli`

When Knip finds a `.commitlintrc` file, it correctly identifies it as a configuration for the `commitlint` plugin but fails to mark `@commitlint/cli` as used because it cannot establish the relationship between the plugin name and the scoped package name.

### DIFF
--- a/packages/knip/src/plugins/commitlint/index.ts
+++ b/packages/knip/src/plugins/commitlint/index.ts
@@ -32,7 +32,9 @@ const resolveConfig: ResolveConfig<CommitLintConfig> = async config => {
         ? [parserPreset.path ?? parserPreset]
         : []
     : [];
-  return [...extendsConfigs, ...plugins, ...formatter, ...parserPresetPaths].map(id => toDependency(id));
+  return ['@commitlint/cli', ...extendsConfigs, ...plugins, ...formatter, ...parserPresetPaths].map(id =>
+    toDependency(id)
+  );
 };
 
 export default {
--- a/packages/knip/test/plugins/commitlint.test.ts
+++ b/packages/knip/test/plugins/commitlint.test.ts
@@ -27,7 +27,7 @@ test('Find dependencies with the Commitizen plugin', async () => {
 
   assert.deepEqual(counters, {
     ...baseCounters,
-    devDependencies: 1,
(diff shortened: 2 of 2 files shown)


webpro:
Thanks for the PR, Azat!

This exact issue I remember writing about here: https://knip.dev/guides/handling-issues#missing-binaries

So [@commitlint/cli](https://www.npmjs.com/package/@commitlint/cli) is a dependency of [commitlint](https://www.npmjs.com/package/commitlint). The confusing part is that they both have a `commitlint` binary. Knip's process of enabling plugins is slightly different than you describe. The enabler in Knip's commitlint plugin is not a config file, but the `@commitlint/cli` dependency:

https://github.com/webpro-nl/knip/blob/d879bb77c69ab07c2929e20dd9c219bb46cf1818/packages/knip/src/plugins/commitlint/index.ts#L13

Knip reads the `package.json` file of dependencies and if the `commitlint` binary is used it will mark the `@commitlint/cli` dependency as referenced.

So if there's only a `.commitlintrc` file and the `@commitlint/cli` dependency is listed, but a reference to the `commitlint` binary is not found by Knip, then the dependency is indeed reported as unused. As there's only configuration, not usage.

The issue with adding `toDependency('@commitlint/cli')` is that it no longer needs to be referenced anywhere, introducing a "blind spot". I'd rather try and find out how we can fix the underlying issue. So my guess is that Knip might have trouble finding the `commitlint` binary? Any chance you can share the reproduction?

webpro:
Please either share a reproduction of the issue, otherwise I my guess is that we can close this one?

PR #1067, opened 2025-05-02
Support custom srcDir in Astro plugin

azat-io:
The Astro plugin currently uses hardcoded paths that start with `src/`, but Astro's configuration allows changing the `srcDir` setting (default is `'src'`). This causes incorrect path resolution and "Unresolved imports" errors when users configure a custom source directory in their `astro.config.js/ts`.

Example:
https://github.com/azat-io/azat-io/blob/main/astro.config.ts#L130

This PR enhances the Astro plugin to dynamically read the Astro configuration and respect the `srcDir` setting.

### DIFF
--- a/packages/knip/src/plugins/astro/index.ts
+++ b/packages/knip/src/plugins/astro/index.ts
@@ -1,6 +1,7 @@
-import type { IsPluginEnabled, Plugin, Resolve, ResolveEntryPaths } from '../../types/config.js';
-import { toDependency, toEntry } from '../../util/input.js';
+import type { IsPluginEnabled, Plugin, Resolve, ResolveFromAST } from '../../types/config.js';
+import { toDependency, toEntry, toProductionEntry } from '../../util/input.js';
 import { hasDependency } from '../../util/plugin.js';
+import { getSrcDir } from './resolveFromAST.js';
 
 // https://docs.astro.build/en/reference/configuration-reference/
 
@@ -12,17 +13,19 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
 
 export const config = ['astro.config.{js,cjs,mjs,ts,mts}'];
 
-const resolveEntryPaths: ResolveEntryPaths = async config => {
-  const srcDir = config?.srcDir ?? './src';
+const resolveFromAST: ResolveFromAST = sourceFile => {
+  const srcDir = getSrcDir(sourceFile);
 
   return [
-    `${srcDir}/content/config.ts`,
-    `${srcDir}/content.config.ts`,
-    `${srcDir}/content/**/*.mdx`,
-    `${srcDir}/pages/**/*.{astro,mdx,js,ts}`,
-    `${srcDir}/middleware.{js,ts}`,
-    `${srcDir}/actions/index.{js,ts}`,
-  ].map(path => toEntry(path));
+    ...[`${srcDir}/content/config.ts`, `${srcDir}/content.config.ts`].map(path => toEntry(path)),
+
+    ...[
+      `${srcDir}/pages/**/*.{astro,mdx,js,ts}`,
+      `${srcDir}/content/**/*.mdx`,
+      `${srcDir}/middleware.{js,ts}`,
+      `${srcDir}/actions/index.{js,ts}`,
+    ].map(path => toProductionEntry(path)),
+  ];
 };
 
 const resolve: Resolve = options => {
@@ -45,6 +48,6 @@ export default {
   enablers,
   isEnabled,
   config,
-  resolveEntryPaths,
+  resolveFromAST,
   resolve,
 } satisfies Plugin;
--- a/packages/knip/src/plugins/astro/index.ts
+++ b/packages/knip/src/plugins/astro/index.ts
@@ -1,6 +1,6 @@
 import type { IsPluginEnabled, Plugin, Resolve, ResolveConfig, ResolveEntryPaths } from '../../types/config.js';
 import { toDependency, toEntry, toProductionEntry } from '../../util/input.js';
-import { hasDependency, load } from '../../util/plugin.js';
+import { hasDependency } from '../../util/plugin.js';
 
 // https://docs.astro.build/en/reference/configuration-reference/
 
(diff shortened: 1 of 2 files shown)


azat-io:
Hmm, there seems to be issues with importing `jsonc` files inside `@astrojs/starlight`.

packages/knip/src/plugins/astro/index.ts
```diff
@@ -12,14 +12,26 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
...
-  'src/content/**/*.mdx',
-  'src/middleware.{js,ts}',
-  'src/actions/index.{js,ts}',
-];
+  return [`${srcDir}/content/config.ts`, `${srcDir}/content.config.ts`].map(path => toEntry(path));
+};
+
+const resolveConfig: ResolveConfig = async (_, options) => {
+  const { configFilePath } = options;
+  const astroConfig = await load(configFilePath);
```
webpro, line 25: The `resolveConfig` function is executed with `_` being the `astroConfig` so there would be no need to also load that separately.

However, the  `resolveEntryPaths` is meant to handle (production) entry paths. Same issue there :)

Please see https://knip.dev/guides/writing-a-plugin#6-resolveconfig and https://knip.dev/guides/writing-a-plugin#7-resolveentrypaths

webpro (commented):
Thanks for the PR! Happy to merge, but please see my remark.

packages/knip/src/plugins/astro/index.ts
```diff
@@ -12,14 +12,22 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
...
-const production = [
-  'src/pages/**/*.{astro,mdx,js,ts}',
-  'src/content/**/*.mdx',
-  'src/middleware.{js,ts}',
-  'src/actions/index.{js,ts}',
-];
+  return [`${srcDir}/content/config.ts`, `${srcDir}/content.config.ts`].map(path => toEntry(path));
+};
+
+const resolveConfig: ResolveConfig = async config => {
```
webpro, line 21: I think we need only `resolveEntryPaths` for entry paths, that functions can also return production entry paths.

webpro (commented):
Would be good to add a test + fixtures folder (e.g. `plugins/astro2`) so we can test better and work on a fix.

packages/knip/src/plugins/astro/index.ts
azat-io: This doesn't seem to be working:

```
Unused files (9)
packages/docs/src/util/post.ts                       
packages/docs/src/components/Contributors.astro      
packages/docs/src/components/EmojiBlastButton.astro  
packages/docs/src/components/Projects.astro          
packages/docs/src/components/Sponsors.astro          
packages/docs/src/components/SponsorsChart.astro     
packages/docs/src/components/SponsorsPast.astro      
packages/docs/src/components/Posts.astro             
packages/docs/src/components/Post.astro
```

webpro: Looks like using `resolveConfig` works fine (you can verify by creating a build of Knip and then run `knip-bun`), except that loading `astro.config.ts` doesn't seem feasible without Bun. 

I think it's happening here, but didn't actually try it out: https://github.com/withastro/astro/blob/4393935d9dbdd5fe7aaeeec310e5cc311408c7e2/packages/astro/src/core/config/vite-load.ts#L32-L59

For this reason I'm afraid we need something else than `resolveConfig`. The good thing is, we now also have `resolveFromAST` — which does not actually `load('astro.config.ts')` — but instead read the file and create an AST that we can traverse to find the value of `srcDir`.

We're doing a very similar thing in the Next.js plugin to get `pageExtensions`:

- https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/next/index.ts#L37-L42
- https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/next/resolveFromAST.ts

From `resolveFromAST`, we can return the same inputs (`toEntry`/`toProductionEntry`) as we would with `resolveConfig`.

Is that something you'd be willing to implement? I can understand it's quite the ride already, but might be interesting to walk the AST as well?

packages/knip/src/plugins/astro/index.ts
webpro: FYI, I have refactored plugins so that we only need to implement `resolveConfig` (`resolveEntryPaths` is removed). This is also updated in the docs: https://knip.dev/guides/writing-a-plugin#example-2-config. However, in https://github.com/webpro-nl/knip/pull/1067#discussion_r2072604186 I've suggested to use `resolveFromAST` so eventually it should not matter for this PR.

packages/knip/src/plugins/astro/index.ts
azat-io: It works, thanks!

packages/knip/src/plugins/astro/index.ts
```diff
@@ -41,7 +48,6 @@ export default {
   enablers,
   isEnabled,
   config,
-  entry,
-  production,
```
webpro, line 45: Sorry to bother you once more, but could you please restore and define `entry` and `production` with default values?

That's for the documentation generator to [read default values](https://github.com/webpro-nl/knip/blob/1bccd592a6cd4652c6dd755c276980332113fc8f/packages/docs/scripts/generate-plugin-docs.ts#L61-L63): https://knip.dev/reference/plugins/astro#default-configuration (`entry` and `production` will be merged and shown as `entry` here).

If `resolveConfig` or `resolveAST` is implemented, it will be ignored: https://github.com/webpro-nl/knip/blob/1bccd592a6cd4652c6dd755c276980332113fc8f/packages/knip/src/WorkspaceWorker.ts#L344

In `resolveAST` we can then read `entry` and `production` and replace the default `srcDir` with the return value of `getSrcDir()`.

packages/knip/src/plugins/astro/resolveFromAST.ts
```diff
@@ -0,0 +1,21 @@
...
+import { getPropertyValues } from '../../typescript/ast-helpers.js';
+
+export const getSrcDir = (sourceFile: ts.SourceFile): string => {
+  let srcDir = 'src';
+
+  function visit(node: ts.Node) {
+    if (ts.isObjectLiteralExpression(node)) {
+      const values = getPropertyValues(node, 'srcDir');
+      if (values.size > 0) {
+        srcDir = Array.from(values)[0];
```
webpro, line 11: Nitpick: we can bail out early and `return` once we've found the value.

webpro (commented):
Thanks, this is great. Happy we found a solution!

packages/knip/src/plugins/astro/index.ts
azat-io: Done

packages/knip/src/plugins/astro/resolveFromAST.ts
azat-io: Done

webpro:
Thank you so much, Azat 🙏

webpro:
:rocket: _This pull request is included in v5.55.0. See [Release 5.55.0](https://github.com/webpro-nl/knip/releases/tag/5.55.0) for release notes._

PR #764, opened 2024-08-21
(PostCSS plugin): Add `postcss` to  used dependencies when using PostCSS with Tailwind CSS

taro-28:
Closes https://github.com/webpro-nl/knip/issues/762

PR #1069, opened 2025-05-03
(PostCSS plugin): Add postcss to used dependencies when using `@tailwindcss/postcss` plugin

karmatys8:
Tailwind v4 [docs](https://tailwindcss.com/docs/installation/using-postcss) recommend using the `@tailwindcss/postcss` plugin, but the current code only handles the `tailwindcss` plugin. The following changes add support for both plugins.

This PR is heavily based on #764.


webpro:
Thanks @karmatys8!

Issue #1052, opened 2025-04-23
💡 Support for "prisma"


TheKnarf:
### Suggest an idea for Knip

Prisma (the DB ORM library) allows you to define a seed script for setting up test data.
You can define it in the package.json file as following:

```js
{
//  ...
  "prisma": {
    "seed": "tsx prisma/seed.ts"
  },
//  ...
}
```

Currently this file shows up under "unused files" in knip, together with everything that currently is only imported by that file.

Example from my project:

```
Unused files (54)
eslint-rules/index.js
jobs/export-folder-cleanup.js
prisma/seed.ts
prisma/seedData/accessPermissionStakeholders.ts
prisma/seedData/accessPermissions.ts
prisma/seedData/assignments.ts
prisma/seedData/buyerApproval.ts
prisma/seedData/companies.ts
prisma/seedData/concerns.ts
prisma/seedData/dataRooms.ts
prisma/seedData/events.ts
prisma/seedData/favorites.ts
prisma/seedData/files.ts
prisma/seedData/folders.ts
prisma/seedData/pendingStakeholders.ts
prisma/seedData/questions.ts
prisma/seedData/replies.ts
prisma/seedData/requests.ts
prisma/seedData/roles.ts
prisma/seedData/sellerResolutions.ts
prisma/seedData/stakeholders.ts
prisma/seedData/tasks.ts
prisma/seedData/testUser.ts
prisma/seedData/users.ts
...
```

webpro:
Good idea!

How far will adding `prisma/seed.ts` as an entry file take us? What would then still be missing?

Any chance you'd be up for a pull request?

TheKnarf:
> How far will adding prisma/seed.ts as an entry file take us?

You would have to at least read the `prisma.seed` field from `package.json` and then use that as an entrypoint.

webpro:
Closing this one by #1073 (thanks @nissy-dev!)

PR #1073, opened 2025-05-06
feat: add prisma plugin

nissy-dev:
This PR adds the [prisma](https://www.prisma.io/) plugin with basic settings.

### DIFF
--- a/packages/knip/src/plugins/prisma/index.ts
+++ b/packages/knip/src/plugins/prisma/index.ts
@@ -15,9 +15,8 @@ const config: string[] = ['prisma.config.ts', 'package.json'];
 
 const resolveConfig: ResolveConfig<PrismaConfig> = async (config, options) => {
   // if config is a package.json file
-  if (config.seed) {
-    const inputs = options.getInputsFromScripts(config.seed);
-    return [...inputs.map(input => (isDeferResolveEntry(input) ? toEntry(input.specifier) : input))];
+  if (options.configFileName === 'package.json' && config.seed) {
+    return options.getInputsFromScripts(config.seed);
   }
   return [];
 };
--- a/packages/knip/src/plugins/prisma/index.ts
+++ b/packages/knip/src/plugins/prisma/index.ts
@@ -1,5 +1,7 @@
-import type { IsPluginEnabled, Plugin } from '../../types/config.js';
+import { isDeferResolveEntry, toEntry } from 'src/util/input.js';
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
 import { hasDependency } from '../../util/plugin.js';
+import type { PrismaConfig } from './types.js';
 
 // https://www.prisma.io/docs/orm/prisma-client/setup-and-configuration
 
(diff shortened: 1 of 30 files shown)


webpro (approved):
Thanks @nissy-dev! This is great, I'm happy to merge this.

Looking at #1052, would you be willing to add this to this or a follow-up pull request? No worries if not.

packages/knip/src/plugins/prisma/index.ts
```diff
@@ -0,0 +1,37 @@
...
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['prisma.config.ts', 'package.json'];
+
+const resolveConfig: ResolveConfig<PrismaConfig> = async (config, options) => {
+  // if config is a package.json file
+  if (config.seed) {
+    const inputs = options.getInputsFromScripts(config.seed);
+    return [...inputs.map(input => (isDeferResolveEntry(input) ? toEntry(input.specifier) : input))];
+  }
```
nissy-dev, line 21: I'm not confident about this code. If I remove this logic and just return the input from `getInputsFromScripts` as-is, then `prisma/seed.ts` was recoginized as `deferResolveEntry` and it was counted in test counter.

Personally, I want this file to be treated as an entry, so I'm transforming it this way. However, I don't fully understand deferResolveEntry, so I'm not sure if this code is correct.

nissy-dev:
> Looking at https://github.com/webpro-nl/knip/issues/1052, would you be willing to add this to this or a follow-up pull request? No worries if not.

Thank you for your review! I added a new commit for this.

packages/knip/src/plugins/prisma/index.ts
webpro: Ah, sorry you hit an edge case. Your intentiosns and assumptions were correct, but somewhere in the core of Knip there is a check if `prisma/seed.ts` looks like a package name, and `prisma` also is listed in `package.json`, and that's where it goes awry...

So that has just been fixed in ffa3ca9a4d22b6fe80c7e7f1dee8f38d8d1d2838 so if you rebase from `main` you could just return  `options.getInputsFromScripts` from `resolveConfig` as-is.

webpro (approved):
Thanks! This looks great.

packages/knip/src/plugins/prisma/index.ts
nissy-dev: Thank you for your explanation. I really appreciate the quick fix!

packages/knip/src/plugins/prisma/index.ts
```diff
@@ -0,0 +1,36 @@
+import { isDeferResolveEntry, toEntry } from 'src/util/input.js';
```
webpro, line 1: Please use relative paths (`../`)

webpro:
Thanks a lot, Daiki! Great to have the Prisma plugin 🙏

webpro:
:rocket: _This pull request is included in v5.55.0. See [Release 5.55.0](https://github.com/webpro-nl/knip/releases/tag/5.55.0) for release notes._

PR #1076, opened 2025-05-08
docs: remove rehype-autolink-headings and use Starlight heading anchor links.

cylewaitforit:
Removes `rehype-autolink-headings`. Heading anchor links were [added to Starlight](https://astro.build/blog/starlight-034/#heading-anchor-links) in v0.34.  

| Before  Video|
| ------- |
|      https://github.com/user-attachments/assets/87158780-4f66-444d-adb1-801f9e8de602        |

|   After  Video|
| ------- |
|       https://github.com/user-attachments/assets/bbce02e7-3d82-466e-b958-90bd77cdf05b       |


webpro:
Thanks! Sorry I missed this one before.

The current situation is pretty bad, now I'm seeing two anchors. I think I like the custom one I did before more as it's more subtle. So seems to me we have two options here:

- We disable the Starlight default
- We remove the custom one and style the Starlight default like the current custom one

WDYT?


Current:

Screenshot 2025-05-15 at 6 54 02 AM


Starlight default:

Screenshot 2025-05-15 at 6 52 56 AM

cylewaitforit:
At least on mobile, the Starlight default is easier to use IMO. So I would probably go with option 2 while ensuring to also keep it always visible on smaller screens where hover over isn’t viable.

Going with the Starlight option under the hood also reduces dependencies and should benefit from any future improvements. Since it would be a widely used and battle tested feature, they should be fairly transparent of any breaking changes in the future.

webpro:
You're right. Any chance you could remove the underline if that's acceptable and remove any remaining styles for the current one we're removing?

cylewaitforit:
Cleaned up the css. Let me know if anything still feel off about it.

webpro:

Issue #1072, opened 2025-05-06
🐛Vite conditional Desktop Imports always raise unused files


clickCA:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/clickCA/vite-unused-desktop.git

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

#### Conditional Desktop Imports

This project uses conditional file imports based on the build type environment variable. When running with `build_type=desktop`, Vite will first look for files with a `.desktop` extension before falling back to the standard files.

For example:

- `ExamplePage.tsx` - standard mobile version
- `ExamplePage.desktop.tsx` - desktop-specific version

##### Known Issue with Knip

When using the Knip dependency analyzer with this setup, there's a issue where Knip fails to detect `.desktop.tsx` files as being used in the project. This leads to these files being incorrectly reported as unused.

##### How to Reproduce the Issue

1. Run the application in desktop mode:

   ```
   build_type=desktop pnpm run dev
   ```

   This works correctly because Vite's configuration includes custom resolution priorities:

   ```js
   function getExtensions() {
     const extensions = ['.js', '.ts', '.tsx', '.json'];
     if (isDesktop) {
       return [...extensions.map((ext) => `.desktop${ext}`), ...extensions];
     }
     return extensions;
   }
   ```

2. Run Knip to analyze your project:

   ```
   pnpm dlx knip
   ```

   Knip will incorrectly report all `.desktop.tsx` files as unused because it doesn't understand the conditional import mechanism controlled by environment variables and Vite's custom resolution order.

webpro:
This should be possible, we're doing something similar in other plugins, e.g.:

- https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/metro/index.ts
- https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/next/index.ts

Feel free to open a PR for this, happy to look into that.

clickCA:
Hi Lars. I've made some change please help me review this. Thanks!

webpro:
Closing this by #1080.

PR #1080, opened 2025-05-09
Add vite plugin to resolve file from `resolve.extension`

clickCA:
#1072 from this issues. 
This PR Introduces support for custom file extensions in Vite configurations and adds tests to ensure proper handling of extension-related issues.


packages/knip/src/plugins/vitest/index.ts
```diff
@@ -115,7 +115,12 @@ export const resolveConfig: ResolveConfig<ViteConfigOrFn | VitestWorkspaceConfig
     }
 
     if (cfg.resolve?.alias) addAliases(cfg.resolve.alias);
-
+    if (cfg.resolve?.extensions) {
+      for (const ext of cfg.resolve.extensions) {
+        if (!ext.startsWith('.')) continue;
+        inputs.add(toEntry(join(dir, `**/*${ext}`)));
```
webpro, line 121: This basically adds all extensions to the default ones. Here's an idea to optimize this:

- use the default list: `DEFAULT_EXTENSIONS` from `constants.ts`
- filter `resolve.extensions` so it does not contain default extensions

Also this is a "wide" pattern:, i.e. `**/*${ext}` will look for matches in all sub directories. Note that this `dir` variable is used for test file patterns. For now I suggest to replace `dir` with a literal `"./src"`.

packages/knip/test/plugins/vite2.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    cwd,
+  });
+
+  assert.ok(issues.files.size > 0);
+
+  const fileIssues = Array.from(issues.files).filter(
+    issue => issue && (issue.includes('mock.ts') || issue.includes('mock.desktop.ts'))
+  );
+
+  assert.ok(fileIssues.length > 0, 'Should find issues with mock files');
```
webpro, line 22: Please use assertions like other test files do. Example:

https://github.com/webpro-nl/knip/blob/a28448b284d8d6d32d083f2127e5001c211f0980/packages/knip/test/plugins/next.test.ts#L16-L17

packages/knip/test/plugins/vite2.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+test('Should find 0 issues with proper extensions config', async () => {
+  const { issues, counters } = await main({
+    ...baseArguments,
+    cwd: cwd2,
+  });
+
+  assert.strictEqual(issues.files.size, 0, 'Should find no issues with complete extension config');
+
+  assert.ok(counters.processed > 0, 'Should process at least one file');
```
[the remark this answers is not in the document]
clickCA: Got it. I will adopt these

packages/knip/src/plugins/vitest/index.ts
clickCA: I have a question about filtering out the `DEFAULT_EXTENSIONS`.

If I set my `resolve` configuration to:
```ts
 resolve: {
    extensions: ['.js'],
 }
```

 My intention is for Knip to consider any `**/*.js` file as a potential entry file.

However, I understand from the documentation that Knip uses a default entry configuration like:

```json
"entry": [
  "{index,cli,main}.{js,cjs,mjs,jsx,ts,cts,mts,tsx}",
   "src/{index,cli,main}.{js,cjs,mjs,jsx,ts,cts,mts,tsx}"
 ],
 ```
 
 My local test when filter out from `DEFAULT_EXTENSIONS` will make the `**/*.js` file undetect.
```ts
  const customExtensions = cfg.resolve.extensions.filter(ext => 
        ext.startsWith('.') && !DEFAULT_EXTENSIONS.includes(ext)
      );
      for (const ext of customExtensions) {
        inputs.add(toEntry(`src/**/*${ext}`));
      }
      
```

This leads to my core uncertainty that should we exclude from the `DEFAULT_EXTENSIONS` or I have miss understand something about `toEntry` function that I'm using?

Any clarification or suggestions would be greatly appreciated.

webpro: In general, we want to minimize the number of entry files. This leads to the best results. One of the main reasons for this is that, by default, Knip does not report unused exports from entry files.

The `resolve.extensions` Vite config option is about module resolution, which is a different concept. Ideally, a Knip plugin could tell the core to add extensions to Knip's internal module resolver. But plugins don't have that capability yet, so that's why we're working around this limitation by adding all matching files with additional extensions as entry files.

The `.js` extension is already one of the `DEFAULT_EXTENSIONS` so we should not add all matching files as entry files (because then suddenly all matching files are entry files which isn't optimal).

However, adding for instance `src/**/*.desktop.js` is OK and should "fix" your original issue.

This is also we I'd like to request and remove fixtures that only exercise `resolve.extensions: ['.js']` if we'd filter that extension out anyway.

clickCA: Got it. Let me try to explain what I understand `resolve.extensions` option in Vite tells Vite how to find modules when an import statement doesn't include a file extension (e.g., import './module' could resolve to module.js, module.ts, etc.)
So basically the `.js` fixtures is not a case that should be tested I will remove. To make the change I agree that we should add entry for only files that have specials like `.desktop.ts`

Will fix soon thanks for clarification.

clickCA:
Hi Lars. Please have a look again. I made some changes

packages/knip/test/plugins/vite2.test.ts
```diff
@@ -0,0 +1,26 @@
...
+import { main } from '../../src/index.js';
+import { join, resolve } from '../../src/util/path.js';
+import baseArguments from '../helpers/baseArguments.js';
+import baseCounters from '../helpers/baseCounters.js';
+
+const cwd = resolve('fixtures/plugins/vite2');
+
+test('Should not find issues with extensions when build_type is desktop', async () => {
+  // Set build_type to desktop for this test
+  process.env.build_type = 'desktop';
```
webpro, line 12: We should not make such assumptions in tests. In this case, I guess in the Vite config we can assume `isDesktop` is true and thus omit it. Or would it cause an exception thrown in a real-world scenario?

webpro (commented):
Thanks, Chanakorn! Let's get this sorted.

packages/knip/test/plugins/vite2.test.ts
clickCA: In my case there is no such error. I assume isDesktop = true is fine

packages/knip/test/plugins/vite2.test.ts
```diff
@@ -0,0 +1,26 @@
...
+  const { counters } = await main({
+    ...baseArguments,
+    cwd,
+  });
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    files: 0,
+    devDependencies: 2,
+    unlisted: 0,
```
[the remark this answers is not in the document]
webpro: I didn't mean to alphabetize but to remove `0` values (just to process/scan faster during dev), but it really doesn't matter much,

webpro:
One last thing: could you please run `bun run format`?

webpro:
:rocket: _This pull request is included in v5.58.0. See [Release 5.58.0](https://github.com/webpro-nl/knip/releases/tag/5.58.0) for release notes._

PR #1086, opened 2025-05-15
Add Changelogen and Changelogithub plugins

azat-io:
Added the Changelogen plugin.
https://github.com/unjs/changelogen

This is a popular changelog generation tool.

Also added support for loading c12 configs:
https://github.com/unjs/c12

I am currently having a problem when using Changelogen configs:

```
> eslint-rule-benchmark@0.2.0 test:usage /Users/azat/Developer/eslint-rule-benchmark
> knip

Unused files (1)
changelog.config.ts 
```

### DIFF
--- a/packages/docs/src/content/docs/reference/plugins.md
+++ b/packages/docs/src/content/docs/reference/plugins.md
@@ -1,5 +1,5 @@
 ---
-title: Plugins (104)
+title: Plugins (107)
 tableOfContents: false
 ---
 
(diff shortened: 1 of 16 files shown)


packages/knip/src/plugins/changelogen/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const title = 'Changelogen';
+
+const enablers = ['changelogen', 'changelogithub'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const resolveConfig = () => {
+  return [];
+};
```
webpro, line 15: We can probably omit `resolveConfig` if we're not using it.

packages/knip/src/plugins/changelogen/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const config = ['package.json', ...toC12config('changelog')];
```
webpro, line 17: We should probably use `entry` instead if we don't actually need the configuration object. And in that case omit `package.json`.

packages/knip/test/plugins/changelogen.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+test('Find dependencies with the Changelogen plugin', async () => {
+  const { counters } = await main({
+    ...baseArguments,
+    cwd,
+  });
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    devDependencies: 1,
```
webpro, line 18: How is `changelog` used? I guess it can be referenced in a `package.json` script or in a GitHub Action workflow file? I think it would be good to reflect that in the fixture so we verify in the test the `changelog` dependency is not unused. We would also need to add a minimal `changelog` package to `node_modules` folder there.

webpro (commented):
Thanks! Happy to merge this but I feel it's not finished yet.

packages/docs/src/content/docs/reference/plugins.md
```diff
@@ -1,216 +1,113 @@
 ---
-title: Plugins (103)
+title: Plugins (104)
 tableOfContents: false
 ---
 
 :::section{.columns.min200}
+- [Angular](/reference/plugins/angular "Angular")
```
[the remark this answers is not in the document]
azat-io: Done

packages/knip/src/plugins/changelogen/index.ts
```diff
@@ -0,0 +1,23 @@
+import type { IsPluginEnabled, Plugin } from '../../types/config.js';
+import { toC12config } from '../../util/plugin-config.js';
+import { hasDependency } from '../../util/plugin.js';
+
+// https://github.com/unjs/changelogen
+
+const title = 'Changelogen';
+
+const enablers = ['changelogen', 'changelogithub'];
```
webpro, line 9: Why is `changelogithub` here?

Maybe it should be a separate plugin? Looking at https://github.com/antfu/changelogithub#configuration, it has `changelogithub.config.{json,ts,js,mjs,cjs}` entry files that we are missing here.

azat-io: I added a separate Changelogen plugin. It also uses c12 to load the config:
https://github.com/unjs/changelogen/blob/main/src/config.ts#L2

packages/knip/schema.json
```diff
@@ -327,6 +327,10 @@
           "title": "Capacitor plugin configuration (https://knip.dev/reference/plugins/capacitor)",
           "$ref": "#/definitions/plugin"
         },
+        "changelogen": {
+          "title": "changelogen plugin configuration (https://knip.dev/reference/plugins/changelogen)",
+          "$ref": "#/definitions/plugin"
+        },
```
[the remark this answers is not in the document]
azat-io: Done!

webpro:
Thanks a lot, Azat! Great stuff ✨

webpro:

PR #1087, opened 2025-05-16
Add Hardhat plugin

ItsNickBarry:
Hardhat is a CLI tool that relies on the presence of a JS config file.  The `hardhat` package may or may not be imported.

This plugin is meant to do two things:
1. Add `hardhat.config.x` as an entry.
2. Skip warnings if `hardhat` is not imported anywhere.

Is the implementation correct?  I'm not sure about the `resolve` function.

### DIFF
--- a/packages/knip/fixtures/plugins/hardhat/package.json
+++ b/packages/knip/fixtures/plugins/hardhat/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "@fixtures/hardhat",
+  "dependencies": {
+    "hardhat": "*"
+  }
+}
--- a/packages/knip/fixtures/plugins/hardhat/package.json
+++ b/packages/knip/fixtures/plugins/hardhat/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fixtures/hardhat",
-  "dependencies": {
+  "devDependencies": {
     "hardhat": "*"
   }
 }
--- a/packages/knip/src/plugins/hardhat/index.ts
+++ b/packages/knip/src/plugins/hardhat/index.ts
@@ -1,5 +1,5 @@
-import { toDependency } from '../../util/input.js';
 import type { IsPluginEnabled, Plugin, Resolve } from '../../types/config.js';
+import { toDependency } from '../../util/input.js';
 import { hasDependency } from '../../util/plugin.js';
 
 // https://hardhat.org/docs
--- a/packages/knip/src/plugins/hardhat/index.ts
+++ b/packages/knip/src/plugins/hardhat/index.ts
@@ -14,7 +14,7 @@ const entry: string[] = ['hardhat.config.{js,cjs,mjs,ts}'];
 
 const resolve: Resolve = async () => {
   return [toDependency('hardhat')];
-}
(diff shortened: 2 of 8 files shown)


packages/knip/fixtures/plugins/hardhat/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@fixtures/hardhat",
+  "dependencies": {
+    "hardhat": "*"
```
webpro, line 4: Looks like it's typically used as dev dep? Seeing `--save-dev` in their docs.

packages/knip/src/plugins/hardhat/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const enablers = ['hardhat'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const entry: string[] = ['hardhat.config.{js,cjs,mjs,ts}'];
+
+const resolve: Resolve = async () => {
+  return [toDependency('hardhat')];
+};
```
webpro, line 17: > The `hardhat` package may or may not be imported.

I understand the reasoning here, but if it's indeed not imported anywhere, then perhaps it's used in some other way? E.g. in `package.json#scripts` or somewhere else?

webpro (commented):
Thanks for the plugin! LGTM, but I do have a remark or two.

packages/knip/fixtures/plugins/hardhat/package.json
ItsNickBarry: Yes

packages/knip/src/plugins/hardhat/index.ts
ItsNickBarry: Some people add scripts, but there's also a `hardhat-shorthand` package which is installed globally and adds an `hh` shortcut to replace `npx hardhat`, `yarn run hardhat`, etc.  It parses the Hardhat config and supports tab completion for builtin and user-defined commands, making it categorically different from scripts.

packages/knip/fixtures/plugins/hardhat/package.json
ItsNickBarry: Changed.  Does that make a difference for production mode?

webpro: Indeed. In general just better to have the fixtures as close to reality as possible.

packages/knip/src/plugins/hardhat/index.ts
webpro: Thanks, bit of a bummer but let's roll with it :)

packages/knip/src/plugins/hardhat/index.ts
```diff
@@ -0,0 +1,25 @@
+import { toDependency } from 'src/util/input.js';
```
webpro, line 1: Please use relative imports

ItsNickBarry: fixed

PR #1089, opened 2025-05-16
Fixes for Mocha plugin

apeloquin-agilysys:
The `mocha` plugin has the following issues:

* Relies on presence of an explicit *require* or *import* of `mocha` in source code, but Mocha does not.
* Does not recognize dependencies (e.g. `ts-node/register` or `dotenv/config`) in the `requires` configuration field.

The `mocha` test has the following issues:

* `isEnabled` returns *false* since the fixture *package.json* dependencies do not include `mocha`; so it does not actually test the plugin
* fixture config has `extension` , which plugin does not support, and does not have `spec` which the plugin does support

This PR addresses:

* `isEnabled` will return true if default Mocha config file is present.
* `resolveConfig` adds `mocha` dependency
* `resolveConfig` supports both entries and dependencies from the `require` field
* test fixture is updated to:
  * use `spec` instead of `extension`
  * include the `mocha`, `@types/mocha`, and `ts-node` dependencies in _package.json_ 
  * include both file and dependency values for `require`

### DIFF
--- a/packages/knip/fixtures/plugins/mocha/.mocharc.json
+++ b/packages/knip/fixtures/plugins/mocha/.mocharc.json
@@ -1,4 +1,4 @@
 {
-  "extension": ["ts"],
-  "require": "ts-node/register"
+  "spec": ["**/*.test.ts"],
+  "require": ["ts-node/register", "hooks.ts"]
 }
--- a/packages/knip/src/plugins/mocha/index.ts
+++ b/packages/knip/src/plugins/mocha/index.ts
@@ -1,5 +1,7 @@
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
-import { toEntry } from '../../util/input.js';
+import { _glob } from '../../util/glob.js';
+import { type Input, toDependency, toEntry } from '../../util/input.js';
+import { extname } from '../../util/path.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { MochaConfig } from './types.js';
 
@@ -9,7 +11,9 @@ const title = 'Mocha';
 
 const enablers = ['mocha'];
 
-const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+const isEnabled: IsPluginEnabled = async ({ cwd, dependencies }) => {
+  return hasDependency(dependencies, enablers) || (await _glob({ cwd, patterns: config })).length > 0;
+};
 
 const config = ['.mocharc.{js,cjs,json,jsonc,yml,yaml}', 'package.json'];
 
@@ -18,7 +22,12 @@ const entry = ['**/test/*.{js,cjs,mjs}'];
 const resolveConfig: ResolveConfig<MochaConfig> = localConfig => {
   const entryPatterns = localConfig.spec ? [localConfig.spec].flat() : entry;
   const require = localConfig.require ? [localConfig.require].flat() : [];
-  return [...entryPatterns, ...require].map(id => toEntry(id));
+
+  const inputs: Input[] = [];
+  inputs.push(toDependency('mocha'));
+  inputs.push(...entryPatterns.map(id => toEntry(id)));
+  inputs.push(...require.map(id => (extname(id) ? toEntry(id) : toDependency(id))));
+  return inputs;
 };
 
 const args = {
--- a/packages/knip/fixtures/plugins/mocha/.mocharc.yml
+++ b/packages/knip/fixtures/plugins/mocha/.mocharc.yml
@@ -1,3 +1,5 @@
-extension:
-  - ts
-require: ts-node/register
+spec:
+  - "**/*.test.ts"
+require:
+  - ts-node/register
+  - hooks.ts
--- a/packages/knip/fixtures/plugins/mocha/example.test.ts
+++ b/packages/knip/fixtures/plugins/mocha/example.test.ts
@@ -1,0 +1,1 @@
+describe('Tests', () => {});
--- a/packages/knip/fixtures/plugins/mocha/hooks.ts
+++ b/packages/knip/fixtures/plugins/mocha/hooks.ts
@@ -1,0 +1,3 @@
+export const mochaHooks = {
+  beforeAll: () => {},
+};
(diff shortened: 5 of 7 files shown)


packages/knip/src/plugins/mocha/index.ts
```diff
@@ -9,7 +11,9 @@ const title = 'Mocha';
 
 const enablers = ['mocha'];
 
-const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+const isEnabled: IsPluginEnabled = async ({ cwd, dependencies }) => {
+  return hasDependency(dependencies, enablers) || (await _glob({ cwd, patterns: config })).length > 0;
```
webpro, line 15: Nearly all plugins work by only looking for the dependency. Why should we treat Mocha differently? It should be in `devDependencies`, no? If there's only e.g. `.mocharc.js` file then we `mocha` should end up as being unused so we can remove it?

packages/knip/src/plugins/mocha/index.ts
```diff
@@ -18,7 +22,12 @@ const entry = ['**/test/*.{js,cjs,mjs}'];
 const resolveConfig: ResolveConfig<MochaConfig> = localConfig => {
   const entryPatterns = localConfig.spec ? [localConfig.spec].flat() : entry;
   const require = localConfig.require ? [localConfig.require].flat() : [];
-  return [...entryPatterns, ...require].map(id => toEntry(id));
+
+  const inputs: Input[] = [];
+  inputs.push(toDependency('mocha'));
```
webpro, line 27: Same is with the `isEnabled` comment, I don't think we should add this here.

packages/knip/src/plugins/mocha/index.ts
```diff
@@ -18,7 +22,12 @@ const entry = ['**/test/*.{js,cjs,mjs}'];
...
+  inputs.push(...entryPatterns.map(id => toEntry(id)));
+  inputs.push(...require.map(id => (extname(id) ? toEntry(id) : toDependency(id))));
```
webpro, line 29: This is good, but we have `toDeferResolve` to resolve it properly (also see https://knip.dev/guides/writing-a-plugin#todeferresolve)

packages/knip/fixtures/plugins/mocha/.mocharc.json
```diff
@@ -1,4 +1,4 @@
 {
-  "extension": ["ts"],
```
webpro, line 2: There is https://mochajs.org/#-extension-ext or is this only supported as command-line argument?

webpro (commented):
Thanks for the PR, Aaron! Much appreciated. Just not sure I agree with the `isEnabled` changes. Any chance you could take a look?

packages/knip/src/plugins/mocha/index.ts
apeloquin-agilysys: If this is counter to Knip's MO, so be it.  Here's where I came into it:
* We have (multiple) mono-repo projects where phantom dependencies have become a frequent problem.
* Almost all of these workspaces within the mono-repo have mocha tests.
* Very few of these projects had an actual `mocha` dependency, but of course all "work" due to dependency hoisting.
* The hope is that `knip` could identify these situations for us, i.e.: "you have tests but no mocha"
If that doesn't sway your stance any, I'll back it out.  I'm open to any `knip`-based or other suggestions for alternatives.

packages/knip/src/plugins/mocha/index.ts
apeloquin-agilysys: Thanks, not sure how I missed this.

packages/knip/fixtures/plugins/mocha/.mocharc.json
apeloquin-agilysys: I removed it from the test because the plugin doesn't support it and `spec` with a glob overrules it.  Figured the test should be matched to the plugin implementation.

packages/knip/src/plugins/mocha/index.ts
webpro: The idea is that if there are tests in workspaces that `import 'mocha'` (or `mocha` is referenced in `package.json#scripts`) then `mocha` should be in either the same or an ancestor (usually the root) workspace.

Here, Knip doesn't check whether `mocha` is available in `node_modules`.

Ideally we could use something like [--strict](https://knip.dev/features/production-mode#strict-mode) in dev mode, but we're not there yet.

Not sure if it's an option, but maybe you could work around this by temporarily removing `mocha` from the root `package.json` (no need to actually uninstall) and see if that helps?

Either way, unfortunately we're in "dev mode" zone here and things aren't as strict as you'd like but eventually we'll get there. For now I'd prefer to keep things consistent so that we can work towards such goals more efficiently.

webpro: > > If there's only e.g. .mocharc.js file then we mocha should end up as being unused so we can remove it?
> 
> Mocha config files can be non-js/ts extensions, so without the plugin enabled it wouldn't show up as an unused file, correct?

That's correct, by default `.json` files aren't part of the analysis.

apeloquin-agilysys: The other way this affects us, is that with `mocha` referenced in `package.json#scripts` and NOT via `import 'mocha'` (which isn't required), `@types/mocha` shows up as an unused dependency.  Is there something that could be done here, i.e. tie the `@types` resolution to a dependency referenced only in `package.json#scripts`?

webpro: Sorry, not sure how to achieve what you need here. I'd say adding `@types/mocha` to `ignoreDependencies` isn't unreasonable here? This can be done per workspace (and workspace configs can be defined with wildcards like `packages/*`).

apeloquin-agilysys: That is the workaround (`ignoreDependencies`) we are currently using.  These issues were "resolved" with the approach in this PR, but I also understand wanting behavioral consistency across the board.  I'll back those changes out.

packages/knip/src/plugins/mocha/index.ts
apeloquin-agilysys: Reverted per above.

webpro:
Thanks, Aaron! Much appreciated. Good discussion to better understand how Knip's being used, feel free to open other issues/RFCs/PRs as you see fit.

webpro:

PR #1090, opened 2025-05-16
detect test files if `tsx --test` script is present

ItsNickBarry:
Similar to how the node plugin detects test files if `node --test` is present.

Documented here: https://tsx.is/node-enhancement#test-runner

Two issues:
1. I may be using `toBinary` incorrectly.  The tests are failing.
2. I'm not sure why `isEnabled` wasn't being used for the tsx plugin before.


ItsNickBarry:
The binary issue seems to be fixed.  Is my solution correct?

packages/knip/fixtures/plugins/tsx/node_modules/tsx/package.json
```diff
@@ -0,0 +1,4 @@
+{
+    "name": "tsx",
+    "bin": ""
```
webpro, line 3: What I typically do is add an empty `index.js` file with `"bin": "./index.js"`. 

Knip looks here for the `tsx` executable referenced in `package.json#scripts` and understands the `tsx` dependency is used.

webpro (commented):
Thanks Nick! Happy to merge, but I do have a remark.

packages/knip/src/plugins/tsx/index.ts
```diff
@@ -1,9 +1,38 @@
...
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['package.json'];
+
+const packageJsonPath = (id: PackageJson) => id;
+
+const resolveConfig: ResolveConfig<PackageJson> = localConfig => {
+  const scripts = localConfig.scripts;
+
+  const entries = [toDependency('tsx')];
```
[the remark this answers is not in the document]
ItsNickBarry: fixed

packages/knip/fixtures/plugins/tsx/node_modules/tsx/package.json
ItsNickBarry: fixed

packages/knip/src/plugins/tsx/index.ts
```diff
@@ -1,9 +1,38 @@
-import type { Plugin } from '../../types/config.js';
+import type { PackageJson } from 'src/types/package-json.js';
+import { toEntry } from 'src/util/input.js';
+import { hasDependency } from 'src/util/plugin.js';
```
webpro, line 3: Please use relative imports

ItsNickBarry: fixed

webpro:
Nice one! 🙏

Issue #1078, opened 2025-05-09
🐛 Nx options.commands are not parsed correctly


BreakBB:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

.

### Reproduction access

- [ ] I've made sure the reproduction is publicly accessible

### Description of the issue

I can not run `knip` on a project, because our `nx` config causes the following error:

> node:internal/validators:162
>    throw new ERR_INVALID_ARG_TYPE(name, 'string', value);
>          ^
> TypeError [ERR_INVALID_ARG_TYPE]: The "str" argument must be of type string. Received an instance of Object

After looking at the code of the nx plugin I am quite sure I found the issue here:

https://github.com/webpro-nl/knip/blob/e52b129776f9e8dcf7ecc677378ccece95630739/packages/knip/src/plugins/nx/index.ts#L62

The `options.commands` array can either be an array of `string` or an array of `object` (see [the nx docs](https://nx.dev/nx-api/nx/executors/run-commands#commands)). But the follow up code expects `scripts` to be an array of `string`.

That means the following `project.json` causes the issue above in an nx context, while it is valid:

```json
{
    "name": "website",
    "$schema": "../node_modules/nx/schemas/project-schema.json",
    "sourceRoot": "website",
    "targets": {
        "serve" : {
            "executor": "nx:run-commands",
            "commands": [
                {
                    "command": "nx serve website"
                }
            ]
        }
    }
}
```

Sorry for not adding a reproduction, just wanted to report this right away so I don't forget about it. If you need one, let me know.

webpro:
Thanks for the report! Feel free to submit a pull request, happy to review.

jajarzombek:
@webpro @BreakBB I've stumbled upon the same issue. any news on the thread? :)

jotaRenan:
I'm also running into this

PR #1093, opened 2025-05-20
Add support for NX rich command structure

jotaRenan:
Fixes https://github.com/webpro-nl/knip/issues/1078. I'm not certain about what changes should be introduced to the tests file 🤔.

### DIFF
--- a/packages/knip/fixtures/plugins/nx/libs/b/project.json
+++ b/packages/knip/fixtures/plugins/nx/libs/b/project.json
@@ -27,7 +27,10 @@
       "executor": "./tools/executors/tsc:tsc"
     },
     "lint": {
-      "command": "biome lint ."
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": [{ "command": "biome lint ." }]
+      }
     }
   }
 }
--- a/packages/knip/src/plugins/nx/index.ts
+++ b/packages/knip/src/plugins/nx/index.ts
@@ -59,7 +59,10 @@ const resolveConfig: ResolveConfig<NxProjectConfiguration | NxConfigRoot> = asyn
     .flatMap(target => {
       if (target.command) return [target.command];
       if (target.options?.command) return [target.options.command];
-      if (target.options?.commands) return target.options.commands;
+      if (target.options?.commands)
+        return target.options.commands.map(commandConfig =>
+          typeof commandConfig === 'string' ? commandConfig : commandConfig.command
+        );
       return [];
     });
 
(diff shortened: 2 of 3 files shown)


webpro:
If you add that, it should not break. Before this PR, it would break. So there's coverage already if you add what you suggest 👍

Now if you want to go the extra mile you could add something like the `biome` executable here:

https://github.com/webpro-nl/knip/blob/a45a1d6253f189bad9af53c6e612ba7a8c4f9611/packages/knip/fixtures/plugins/nx/libs/b/project.json#L30

Which ends up in the test assertions like so:

https://github.com/webpro-nl/knip/blob/a45a1d6253f189bad9af53c6e612ba7a8c4f9611/packages/knip/test/plugins/nx.test.ts#L29

You'd also need to increment the `binaries` here:

https://github.com/webpro-nl/knip/blob/a45a1d6253f189bad9af53c6e612ba7a8c4f9611/packages/knip/test/plugins/nx.test.ts#L33

You can run the tests locally (also see https://github.com/webpro-nl/knip/blob/main/.github/DEVELOPMENT.md#fixtures--tests), I like pressing F5 in VS Code when I have e.g. the `nx.test.`ts` file open!

jotaRenan:
@webpro sorry for not being clear. The test does not break if I add the snippet and run the test on a branch that doesn't contain my commit.

webpro:
Then we need to ask ourselves what we are fixing here? :)

webpro:
Ah, seeing that stacktrace I understand the issue. My expectation was that we were patching an uncovered object structure in the Nx plugin. Which you did, but also the issue was that there's a bug in [this error handler](https://github.com/webpro-nl/knip/blob/a45a1d6253f189bad9af53c6e612ba7a8c4f9611/packages/knip/src/binaries/bash-parser.ts#L121) which chokes on `script` not being a string.

Let's skip that assertion you were asking about; yet maybe you could still add the one I've mentioned?

jotaRenan:
@webpro I actually got it now! In this [comment](https://github.com/webpro-nl/knip/pull/1093#issuecomment-2895348216) I mentioned I added the snippet to a project.json file but it was the one in the `apps` path, not the `libs` one used by the test assertions. 

I pushed [ae64221](https://github.com/webpro-nl/knip/pull/1093/commits/ae642214f84fe030dd8b4f61dfd0b883018b962b) and ensured that test indeed breaks in `main` and no longer breaks with [7bcb26b](https://github.com/webpro-nl/knip/pull/1093/commits/7bcb26b510bb9fc7b4089b75e0a79c32a5cdfc6c). I believe that suffices for the PR, right? Thanks for the assistance.

webpro:
Thanks João, much appreciated. Let's roll with it 🛼

webpro:

Issue #1084, opened 2025-05-13
🐛 Astro.js layouts imported inside MDX files are seen as unused files


williamgentil:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://stackblitz.com/edit/github-kj3kesrk?file=package.json,src%2Fpages%2Fservices%2Fservice-1.mdx,src%2Fpages%2Findex.astro

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

Hi, 
Everytime I'm using MDX files with Astro.js, the imported layout inside the MDX file is seen as an "unsued file" by knip (see reproduction url). 

![Image](https://github.com/user-attachments/assets/4e1445b1-9443-4dba-8e4a-93f81ee295ca)

Am I doing something wrong somewhere?  
Thanks for your help.

webpro:
Knip has a few simple built-in "compilers": https://knip.dev/features/compilers

Relevant example here:

```mdx
---
layout: "../../layouts/Layout.astro"
---

import Button from '../../components/Button.astro';

```

The `import` should work fine, but there's no support for anything in the fenced code block yet.

So what you would need here is either:

- Make Knip smarter by not [removing the fenced code block](https://github.com/webpro-nl/knip/blob/3643374749225c5373e38fecca76c993d11684c7/packages/knip/src/compilers/mdx.ts#L20) but extracting out relevant values instead (like we do with the `import` matcher). Maybe you would be up for a pull request and fix this up?
- Use an actual compiler in your project (also see docs linked to above)

cylewaitforit:
I could be wrong, but if I’m following the code correctly, I don’t believe the front matter is being removed as a fenced code block. Since it is surrounded by `---` and not ```.

yudao:
Hello 👋 
@webpro I think just changing the regex could fix this issue, what do you think?

https://github.com/webpro-nl/knip/blob/3643374749225c5373e38fecca76c993d11684c7/packages/knip/src/compilers/compilers.ts#L3

```typescript
[...]

PR #1102, opened 2025-05-25
feat(astro-mdx): add compiler for @astrojs/mdx to treat layout value from mdx frontmatter as import

cylewaitforit:
Treats the layout in the mdx frontmatter in an Astro project as an import.

Related Issue: #1084

### DIFF
--- a/packages/knip/src/compilers/compilers.ts
+++ b/packages/knip/src/compilers/compilers.ts
@@ -28,3 +28,16 @@ export const scriptBodies: SyncCompilerFn = (text: string) => {
   }
   return scripts.join(';\n');
 };
+
+// Extract paths as imports from frontmatter for given keys (e.g., 'layout')
+const frontmatterMatcher = /---[\s\S]*?---/g;
+export const importsWithinFrontmatter = (text: string, keys: string[] = []) => {
+  const imports = [...text.matchAll(frontmatterMatcher)].flatMap(([frontmatter]) => {
+    return keys.flatMap(key => {
+      const valueMatcher = new RegExp(`${key}:\\s*["']([^"']+)["']`, 'i');
+      const match = frontmatter.match(valueMatcher);
+      return match?.[1] ? [`import ${key} from "${match[1]}";`] : [];
+    });
+  });
+  return imports.join('\n');
+};
--- a/packages/knip/src/compilers/compilers.ts
+++ b/packages/knip/src/compilers/compilers.ts
@@ -30,14 +30,15 @@ export const scriptBodies: SyncCompilerFn = (text: string) => {
 };
 
 // Extract paths as imports from frontmatter for given keys (e.g., 'layout')
-const frontmatterMatcher = /---[\s\S]*?---/g;
+const frontmatterMatcher = /---[\s\S]*?---/;
 export const importsWithinFrontmatter = (text: string, keys: string[] = []) => {
-  const imports = [...text.matchAll(frontmatterMatcher)].flatMap(([frontmatter]) => {
-    return keys.flatMap(key => {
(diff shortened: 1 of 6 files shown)


packages/knip/src/compilers/compilers.ts
```diff
@@ -28,3 +28,16 @@ export const scriptBodies: SyncCompilerFn = (text: string) => {
   }
   return scripts.join(';\n');
 };
+
+// Extract paths as imports from frontmatter for given keys (e.g., 'layout')
+const frontmatterMatcher = /---[\s\S]*?---/g;
```
webpro, line 33: This is a greedy match, while there can only be a single frontmatter I think?

packages/knip/src/compilers/mdx.ts
```diff
@@ -17,6 +17,18 @@ const mdxDependencies = [
...
 
-const compiler = (text: string) => [...text.replace(fencedCodeBlockMatcher, '').matchAll(importMatcher)].join('\n');
+const compiler = (text: string) => {
+  // Get imports from the MDX content ignoring imports within fenced code blocks
+  const imports = [...text.replace(fencedCodeBlockMatcher, '').matchAll(importMatcher)].map(match => match[0]);
+
+  let frontmatterImports = '';
+
+  // If this is an Astro project, also treat the layout path in the frontmatter as an import
+  if (mdxDependencies.includes('astro')) {
```
webpro, line 27: The generic `mdxDependencies` array contains enablers for the mdx compiler (it's not specific to the project).

packages/knip/src/compilers/mdx.ts
```diff
@@ -17,6 +17,18 @@ const mdxDependencies = [
...
+    frontmatterImports = importsWithinFrontmatter(text, ['layout']);
```
webpro, line 28: Feels a bit odd with array/string concatenation. How about working with an array of strings only, just until the end? Empty strings aren't an issue then either.

Or even something like `return [...text.replace(), ...importsWithinFrontmatter()]`

webpro (commented):
Thanks for the PR. Seeing this, I'm a bit on the fence when it comes to the matter of adding more complexity to the already "hacky" compilers. This is a "cheap" solution without dependencies, but what's next? Also, can `layout` be some other user-defined key in the frontmatter without issues?

Just wondering, wouldn't something like https://knip.dev/features/compilers#mdx but with `@astrojs/mdx` be feasible? If we can support this `layout` somewhat nicely & safely without users needing to config anything would still be nice, though.

packages/knip/src/compilers/compilers.ts
cylewaitforit: I believe there would only be one. Updated the frontmatter matcher to not be a greedy match.

packages/knip/src/compilers/mdx.ts
cylewaitforit: That makes sense. Seems like one option would be to treat layout as a frontmatter for all mdx but that seemed wasteful. Instead I moved AstroMDX as its own compiler and swapped that in the getIncludedCompilers if there was a dependency match.

packages/knip/src/compilers/mdx.ts
cylewaitforit: Agree, I think working with array of strings until the end is easier to follow.

cylewaitforit:
> Thanks for the PR. Seeing this, I'm a bit on the fence when it comes to the matter of adding more complexity to the already "hacky" compilers. 

@webpro  I iterated more on this based on the feedback.  I think it still feels mostly "hacky".  I will feel no real loss if this doesn't feel like the right direction to solve this. At least if it's wrong it gives a better picture of what right might end up looking like( or not looking like).

PR #1105, opened 2025-05-26
Add `entry` and `project` configuration hints

### LINKED ISSUE #331: Show a warning when a path in the "knip.json" file does not match any results
For example, if the entry point was `foo.ts`, and the user typoed it to be `fooo.ts`, then Knip should immediately show a warning. Failing fast is important, as it potentially save the end-user a bunch of troubleshooting time.

### LINKED ISSUE #618: Issue a warning or fail if a specific file given as an entry point does not exist
Simple `knip.json`:

```
{
  "entry": [
    "src/main.tsx!",
    "does-not-exist.ts",
  ]
}
```

`does-not-exist.ts` is a non-existent file.  When you run `npm run knip`, there is no warning or error even though `does-not-exist.ts` does not exist.

It would be great if Knip had a check that issued an error or warning if any file specified as an entry point did not exist (and maybe also for regex if the regex had no match).

For large projects that change over time, I would prefer to get an error that could be caught in CI so that I could update the `knip.config` when I make major changes that affect the entry points.

### LINKED ISSUE #941: 💡 Report missing entries (file defined in `entry` but missing)
## Repro

Config:

```json
{
    "$schema": "https://unpkg.com/knip@5/schema.json",
    "entry": [
        "b.js"
    ]
}
```

Repo contents:

- a.js

Command:

```sh
$ npx knip@5.43.6
```

## Current output

```
$ npx knip@5.43.6
Unused files (1)
a.js
```

This is a config error on my part, but knip did not alert me.

## Ideal output

```sh
$ npx knip@5.43.6

Unused files (1)
a.js

Entries not found (1)
b.js
```

## Ideal output part 2

Globs could also be reported if they don't match anything, both as `entry` and `project`.

```json
{
    "$schema": "https://unpkg.com/knip@5/schema.json",
    "entry": [
        "*.ts"
    ]
}
```

```sh
$ npx knip@5.43.6

Unused files (1)
a.js

Glob not matching any files (1)
*.ts
```

### LINKED ISSUE #1025: 💡 Detect unnecessary entry points
### Suggest an idea for Knip

Similar to how Knip will currently give a warning log if you specify an ignored dependency/etc.:

```plaintext
$ npx knip
Configuration hints
Unused item in ignoreDependencies: typescript
✂️  Excellent, Knip found no issues.
```

https://github.com/JoshuaKGoldberg/create-typescript-app/issues/2147#issuecomment-2791603713 got me thinking, it would be nice to be informed if an `entry` point doesn't do anything. Both if a Knip config starts off unnecessarily verbose, or if a plugin is added/improved over time to make a previously necessary `entry` string unnecessary.

Proposal: for each string element in `entry`, could Knip check if it doesn't change anything, and report a complaint if so?

```plaintext
$ npx knip
Configuration hints
Unused glob in entry: src/content-script.ts (detected by the esbuild plugin)
✂️  Excellent, Knip found no issues.
```

webpro:
This pull request adds configuration hints for `entry` and `project` items:

- Redundant `entry` or `project` items already added (e.g. by plugins)
- Unused `entry` or `project` items without matches in the globbed files
- Any `entry` or `project` patterns array that results in zero files
- In monorepos, top-level `entry` and `project` that should sit next to the other workspaces (at `workspaces["."]`)

Perhaps best to show some example output:

```
❯ knip
Configuration hints (1)
Redundant item in entry: src/content-script.ts  (hint: remove)
```

```
❯ knip
Configuration hints (3)
Unused item in project (workspace: webview-ui): disappeared/**  (hint: remove, or move to workspace config)
Unused item in top-level entry: [src/extension.ts, …]  (hint: move to workspaces["."])
Unused item in top-level project: [src/**/*.ts, …]  (hint: move to workspaces["."])
```

```
❯ knip
Configuration hints (1)
Revisit entry (workspace: packages/integrations/solid): [test/**/*.test.js]  (hint: no files found)
```

Notes:

- Not behind a flag, we're starting out relatively conservatively and "cheap", no extra detailed output (yet?)
- Negative patterns are ignored (performance penalty, because we'd have to glob without the pattern as well)
- Workspaces that match a wildcard config (e.g. `"packages/*": {}`) are ignored
- Configuration hints are warnings by default, flags `--no-config-hints` and `--treat-config-hints-as-errors` available.

Would be great if you could give it a spin and maybe 👍 or 👎 or some feedback :)

Fixes #331
Fixes #618
Fixes #941
Fixes #1025 

/cc @Zamiell @ptmkenny @fregante @JoshuaKGoldberg @lishaduck @AndreaPontrandolfo @narthur

### DIFF
--- a/packages/knip/fixtures/configuration-hints/knip.json
+++ b/packages/knip/fixtures/configuration-hints/knip.json
@@ -1,0 +1,6 @@
+{
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+  "entry": ["src/entry.js"],
+  "project": ["src/**"],
+  "workspaces": {}
+}
--- a/packages/knip/fixtures/configuration-hints/lib/package.json
+++ b/packages/knip/fixtures/configuration-hints/lib/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "@fixtures/configuration-hints__lib"
+}
--- a/packages/knip/fixtures/configuration-hints/package.json
+++ b/packages/knip/fixtures/configuration-hints/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "@fixtures/configuration-hints",
+  "workspaces": ["lib"]
+}
--- a/packages/knip/fixtures/configuration-hints2/knip.json
+++ b/packages/knip/fixtures/configuration-hints2/knip.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+  "entry": ["src/entry.js", "lib/index.js"],
+  "project": ["src/**"],
+  "workspaces": {
+    ".": {
+      "entry": ["src/entry.js", "lib/index.js"],
+      "project": ["src/**", "lib/**"]
+    }
+  }
+}
--- a/packages/knip/fixtures/configuration-hints2/lib/package.json
+++ b/packages/knip/fixtures/configuration-hints2/lib/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "@fixtures/configuration-hints__lib"
+}
--- a/packages/knip/fixtures/configuration-hints2/package.json
+++ b/packages/knip/fixtures/configuration-hints2/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "@fixtures/configuration-hints",
+  "workspaces": ["lib"]
+}
--- a/packages/knip/src/ConfigurationChief.ts
+++ b/packages/knip/src/ConfigurationChief.ts
@@ -1,6 +1,7 @@
 // biome-ignore lint/nursery/noRestrictedImports: ignore
 import path from 'node:path';
 import picomatch from 'picomatch';
+import type { z } from 'zod';
 import { partitionCompilers } from './compilers/index.js';
 import type { SyncCompilers } from './compilers/types.js';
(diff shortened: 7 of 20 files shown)


fregante:
Tested and it seems to work after I add redundant entries. I can also confirm that there are no false positives or other issues in my existing long config.

> * Workspaces that match a wildcard config (e.g. `"packages/*": {}`) are ignored

Probably a good behavior, even if it means that errors here can't be caught.

webpro:
We can do `knip --performance-fn worker.getConfigurationHints` and track perf of just this method

narthur:
Just tried it in a couple of my projects, everything looks good, and it found some unused stuff in one of my knip config files. Great stuff!

webpro:
Thanks everyone! LFG 🚀

Hopefully this will help both...

- users new to knip and nudge them in the right direction in large repos where tons of unused files in output can be overwhelming:

Screenshot 2025-06-04 at 10 17 14 AM


- and existing users and notify about unused `entry` and `project` etc. as Knip itself improves:


Screenshot 2025-06-04 at 10 19 10 AM

webpro:
:rocket: _This pull request is included in v5.60.0. See [Release 5.60.0](https://github.com/webpro-nl/knip/releases/tag/5.60.0) for release notes._

fregante:
> * Workspaces that match a wildcard config (e.g. `"packages/*": {}`) are ignored

Does this cover only `/*` or should it work for `/*-e2e`? Because I'm now getting some warnings on my `@org/*-e2e` config object.

PR #1106, opened 2025-05-27
Fix binary resolution

bartschriever:
This PR fixes binary resolution issues.


webpro:
What issues exactly? Any chance you could describe the issue in a little bit more detail and maybe give an example of the issue we're looking to resolve?

PR #1109, opened 2025-05-28
feat: `tsdown` plugin

TheAlexLichter:
This plugin adds a plugin for [`tsdown`](https://github.com/rolldown/tsdown/).

I've reused most code from the [existing tsup plugin](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/tsup/index.ts).


TheAlexLichter:
Whoops, didn't see the scaffolding script 🙈

Ran it now!

webpro:
Good stuff! 🚀

(for next time: https://knip.dev/guides/writing-a-plugin)

PR #1112, opened 2025-06-02
Fix Eleventy plugin error with certain configuration file formats

### LINKED ISSUE #1110: 🐛
### Prerequisites

### Reproduction url

https://gitlab.com/aarongoldenthal/aarongoldenthal

### Reproduction access

### Description of the issue

In some cases, running knip on an eleventy project fails with the following error (this is the Alpine Linux output, but occurs on Win 11 as well)

```sh
$ knip
file:///usr/local/lib/node_modules/knip/dist/plugins/eleventy/index.js:19
    const templateFormats = localConfig.templateFormats || defaultEleventyConfig.templateFormats;
                                        ^
TypeError: Cannot read properties of undefined (reading 'templateFormats')
    at Object.resolveConfig (file:///usr/local/lib/node_modules/knip/dist/plugins/eleventy/index.js:19:41)
    at async runPlugin (file:///usr/local/lib/node_modules/knip/dist/WorkspaceWorker.js:276:40)
    at async WorkspaceWorker.runPlugins (file:///usr/local/lib/node_modules/knip/dist/WorkspaceWorker.js:319:13)
    at async build (file:///usr/local/lib/node_modules/knip/dist/graph/build.js:94:35)
    at async main (file:///usr/local/lib/node_modules/knip/dist/index.js:[43](https://gitlab.com/aarongoldenthal/aarongoldenthal/-/jobs/10184307712#L43):88)
    at async run (file:///usr/local/lib/node_modules/knip/dist/cli.js:26:111)
    at async file:///usr/local/lib/node_modules/knip/dist/cli.js:111:1
Node.js v22.16.0
```

The issue is in line https://github.com/webpro-nl/knip/blob/8aa2f13d4f81a7dcd2cd8c3c2a1e7d7c6102b9cb/packages/knip/src/plugins/eleventy/index.ts#L31 of the plugin. The error occurs when `localConfig` is undefined, and this line is the one that doesn't handle that.

The following change resolved the issue.

```diff
-  const templateFormats = localConfig.templateFormats || defaultEleventyConfig.templateFormats;
+  const templateFormats = localConfig?.templateFormats || defaultEleventyConfig.templateFormats;
```

aarongoldenthal:
This PR resolves a bug where the Eleventy plugin throws an error when the configuration file uses a named `config` export in addition to default export  (the first option in the [config file formats](https://www.11ty.dev/docs/config-shapes/)), and updates the associated tests and documentation.

Closes #1110.

- Fixed issue where Eleventy plugin throws an error if the named `config` export is used, and therefore the default export function does not return a value.
- Added a note to the Eleventy plugin on supported config file formats (following the ESLint plugin as an example).
- Updated test and fixture `eleventy3` to include a `data` directory that does not use the default value, which allowed case 3/4 to have the same fixture but with different config formats (and different results).
- Added a test and fixture `eleventy4`, which is a duplicate of `eleventy3` but uses the unsupported `config` export to confirm the error is no longer thrown.

@webpro I did not see an obvious way to support the named `config` export in addition to the default export without a something more significant like changing the plugin to use `resolveFromAST` (or having the plugin load manually load the `config` export, which seemed inconsistent with the general plugin methodology), but let me know if I'm missing something.

### DIFF
--- a/packages/knip/fixtures/plugins/eleventy3/.eleventy.js
+++ b/packages/knip/fixtures/plugins/eleventy3/.eleventy.js
@@ -48,7 +48,7 @@ module.exports = function (eleventyConfig) {
       input: 'src',
       includes: '_includes',
       layouts: '_includes/layouts',
-      data: '_data',
+      data: '_siteData',
       output: '_site',
     },
   };
--- a/packages/knip/fixtures/plugins/eleventy3/src/_siteData/global.cjs
+++ b/packages/knip/fixtures/plugins/eleventy3/src/_siteData/global.cjs
@@ -1,0 +1,5 @@
+const data = {
+  title: 'Eleventy',
+};
+
+module.exports = data;
--- a/packages/knip/fixtures/plugins/eleventy4/.eleventy.js
+++ b/packages/knip/fixtures/plugins/eleventy4/.eleventy.js
@@ -49,7 +49,7 @@ module.exports.config = {
     input: 'src',
     includes: '_includes',
     layouts: '_includes/layouts',
-    data: '_data',
+    data: '_siteData',
     output: '_site',
   },
 };
--- a/packages/knip/fixtures/plugins/eleventy4/.eleventy.js
+++ b/packages/knip/fixtures/plugins/eleventy4/.eleventy.js
@@ -1,0 +1,55 @@
+// biome-ignore lint/suspicious/noRedundantUseStrict:  fixture festa
+'use strict';
+
+const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
+const syntaxHighlight = require('@11ty/eleventy-plugin-syntaxhighlight');
+const path = require('node:path');
+const highlighter = require('./src/_plugins/syntax-highlighter');
+path;
+highlighter;
+
+// biome-ignore lint/complexity/useArrowFunction:  fixture festa
+module.exports = async function (eleventyConfig) {
+  eleventyConfig.addGlobalData('site_name', 'example');
+  eleventyConfig.addDataExtension('yml', () => {});
+  eleventyConfig.addFilter('limitTo', () => {});
+  eleventyConfig.addFilter('sortByPageOrder', () => {});
+  eleventyConfig.addPlugin(eleventyNavigationPlugin);
+  eleventyConfig.addPlugin(syntaxHighlight, {
+    alwaysWrapLineHighlights: true,
+    templateFormats: ['liquid', 'njk'],
+  });
+  eleventyConfig.setLibrary('md', () => {});
+  eleventyConfig.addNunjucksShortcode('link', () => {});
+  eleventyConfig.addShortcode('related_rules', () => {});
+  eleventyConfig.addWatchTarget('./src/assets/');
+  eleventyConfig.addPassthroughCopy({
+    'src/_includes/abc.js': '/assets/abc.js',
+  });
+  eleventyConfig.addPassthroughCopy({
+    './src/static': '/',
+  });
+  eleventyConfig.addPassthroughCopy('./src/assets/');
+  eleventyConfig.addPassthroughCopy('src/static');
+  eleventyConfig.addCollection('docs', () => {});
+  eleventyConfig.addCollection('library', () => {});
+  eleventyConfig.ignores.add('src/static/sitemap.njk');
+
+  eleventyConfig.addPassthroughCopy({
+    './node_modules/@org/lib/dist/build.js': '/assets/lib.js',
+  });
+};
+
+module.exports.config = {
+  passthroughFileCopy: true,
+  pathPrefix: '/docs/head/',
+  markdownTemplateEngine: 'njk',
+  htmlTemplateEngine: 'njk',
+  dir: {
+    input: 'src',
+    includes: '_includes',
+    layouts: '_includes/layouts',
+    data: '_data',
+    output: '_site',
+  },
+};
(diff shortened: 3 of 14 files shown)


webpro:
Thanks for the pull request, Aaron!

Could we please minimize the changes to the fixtures directory. There was a lot of files already, my bad, but maybe we could reduce it a bit and keep just the essentials?

The other idea I'm having is to actually support the option that tooling could consume a specified named export if it's not the default export. It's quite the exception so we don't have this yet, but also it'd be rather "cheap" to support: 

https://github.com/webpro-nl/knip/blob/18842738a4acc06e11654a853c5247f0e8399d27/packages/knip/src/WorkspaceWorker.ts#L378

Here we could say "hey if the default export (`localConfig`) is not set and the plugin has some option set we could also read that named export instead". First idea/example:

```ts
const namedExport = 'config'; // add to default export of eleventy plugin
```

Is this something you'd like to implement? If so, great, that could go into this PR (I'm happy to assist further). Otherwise I'm good to pick it up after this PR has been merged. Either way, we can remove the `note` from the plugin as I'm convinced we don't need it :)

aarongoldenthal:
> Could we please minimize the changes to the fixtures directory. There was a lot of files already, my bad, but maybe we could reduce it a bit and keep just the essentials?

I updated fixture4 so that it only tests the two exports. I also removed the changes to fixture3 since the `data` directory is tested in fixture2.

> 
> The other idea I'm having is to actually support the option that tooling could consume a specified named export if it's not the default export. It's quite the exception so we don't have this yet, but also it'd be rather "cheap" to support:
> 
> https://github.com/webpro-nl/knip/blob/18842738a4acc06e11654a853c5247f0e8399d27/packages/knip/src/WorkspaceWorker.ts#L378
> 
> Here we could say "hey if the default export (`localConfig`) is not set and the plugin has some option set we could also read that named export instead". First idea/example:
> 
> ```ts
> const namedExport = 'config'; // add to default export of eleventy plugin
> ```
> 
> Is this something you'd like to implement? If so, great, that could go into this PR (I'm happy to assist further). Otherwise I'm good to pick it up after this PR has been merged. Either way, we can remove the `note` from the plugin as I'm convinced we don't need it :)

Note removed :)

I love the idea, especially since it's the default format that Eleventy advertises (I believe it was a v3 change). Just to be clear, it does still use the default export (which would populate `dummyUserConfig` with actual data), but the `config` export has what would otherwise be returned from the default export (for example, the `dir` object). I think similar logic would still work though - load the default export, and load a named export based on a plugin option. Unfortunately I think that's more than I can take on at this point.

webpro:
Thanks!

To be 100% on the same page, regarding the `default` or named `config` export. My understanding is that we need only one (never both). Here it says "instead": https://www.11ty.dev/docs/config-shapes/#optional-return-object, and here the docs say "or": https://www.11ty.dev/docs/config/#configuration-2. Would you agree that we can just do "use `config` if it exists, otherwise the default one"? I'm asking, because the fixture has both: https://github.com/webpro-nl/knip/blob/2ba25c2e123456613ab2a43421e8c8e881e09d7d/packages/knip/fixtures/plugins/eleventy4/.eleventy.js

aarongoldenthal:
> Thanks!
> 
> To be 100% on the same page, regarding the `default` or named `config` export. My understanding is that we need only one (never both). Here it says "instead": https://www.11ty.dev/docs/config-shapes/#optional-return-object, and here the docs say "or": https://www.11ty.dev/docs/config/#configuration-2. Would you agree that we can just do "use `config` if it exists, otherwise the default one"? I'm asking, because the fixture has both: https://github.com/webpro-nl/knip/blob/2ba25c2e123456613ab2a43421e8c8e881e09d7d/packages/knip/fixtures/plugins/eleventy4/.eleventy.js

I was double checking the docs and found a new and interesting feature that may complicate things.

- My understanding has been that if the default export was a function, it was passed an `eleventyConfig` class that configures plugins, shortcodes, filters, etc. In knip the `dummyUserConfig` is passed and is then populated with any applicable data. In the eleventy4 case, it sets the pass though directory to `./src/assets/`, which adds that path to production entrypoints, and then `./src/assets/index.js` is not identified as an error.
- The eleventy config file also provides a config object with static data, mostly configuring directories and file formats used by eleventy. This can be provided as the default export, can be returned from the default export function, or provided through the named `config` export (per [the docs](https://www.11ty.dev/docs/config-shapes/). This would provide the `dir.input` and `dir.data` values, and then add the files in `src/_siteData/**/*.js...` as entrypoints.

I think that's all still true, and the eleventy4 case covers the default and named exports (there's another example in the [Eleventy base blog template](https://github.com/11ty/eleventy-base-blog/blob/main/eleventy.config.js) with the two exports).

The complication I found in the docs is that Eleventy v3 added functions to the `eleventyConfig` class as alternatives to the `config` properties (for example `eleventyConfig.setInputDirectory()` instead of `dir.input` (per [the docs](https://www.11ty.dev/docs/config/#input-directory)). Those new functions aren't included at https://github.com/webpro-nl/knip/blob/bd18e06deafd432dd9b9f583312f048944876296/packages/knip/src/plugins/eleventy/helpers.ts#L4, so if used they throw an error. When I found it I was checking to see how it was returned from the default export function. I didn't have a chance yet to see how many of options like that there are.

webpro:
Eventually plugins try their best to gather all the configurations they can, but don't necessarily replicate all of the tool's behaviors. In the Knip configuration we can adjust as necessary, e.g. if the `data: '_data'` option isn't recognized properly users can configure it like so in `knip.json`:

```json
{
  "eleventy": {
    "entry": ["src/**/*.ts", "_siteData/**/*.ts"]
  }
}
```

I'm going to merge this and then add a few dummy `setInputDirectory` config methods to further reduce potential trouble.

Feel free to tweak things further as you see fit. Thanks!

webpro:
:rocket: _This pull request is included in v5.60.1. See [Release 5.60.1](https://github.com/webpro-nl/knip/releases/tag/5.60.1) for release notes._

Issue #901, opened 2025-01-04
🐛 False Unused devDependencies report for biome extensions


ianzone:
### Prerequisites

- [X] I'm using the latest version
- [X] I've read the relevant [documentation](https://knip.dev)
- [X] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [X] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [X] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/ianzone/react-storybook

### Reproduction access

- [X] I've made sure the reproduction is publicly accessible

### Description of the issue

knip reports
```bash
Unused devDependencies (2)
@storybook/blocks  package.json
biome-extends      package.json
```
While `biome-extends` is consumed by `biome.json`.

webpro:
Thanks for the report, Yiheng! There's no Biome plugin yet, so this makes sense.

Would you like to add a plugin and submit a pull request?

ianzone:
Ok, I'll write it.

insulineru:
@ianzone if you don't get a chance to implement it, please post it here :)

ianzone:
> @ianzone if you don't get a chance to implement it, please post it here *)

Hi, I'm getting busy this month. Many thanks if you're willing to take it

tidefield:
@ianzone I'm trying to reproduce your report. With your repo, shouldn't report be

```
Unused devDependencies (2)
@storybook/blocks  package.json
@biomejs/biome      package.json
```
instead of `biome-extends` ?

tidefield:
@ianzone I think I was able to find the culprit commit https://github.com/ianzone/react-storybook/commit/cb3450a581a3a042ec8a24d92592d66809835ac2#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bb. My plugin implementation will support this use case.

Issue #1071, opened 2025-05-05
💡 Add plugin support for biome


daivinhtran:
### Suggest an idea for Knip

https://knip.dev/reference/plugins does not have support for [biome](http://biomejs.dev/) yet. I'd be happy to contribute one.

jdfm:
Seems like this issue may be a duplicate of https://github.com/webpro-nl/knip/issues/901, though, this one corresponds more closely to what needs to be added for #901 to be resolved.

daivinhtran:
Yeah I haven't found time to work on this. I'll get started this weekend!

joealden:
@webpro @daivinhtran FYI this should be be closed (https://github.com/webpro-nl/knip/pull/1113)

webpro:
Absolutely, thank you!

PR #1113, opened 2025-06-02
feat: `biome` plugin

daivinhtran:
This PR adds a plugin for biome.

Since the biome is just tooling on top of existing Typescript projects, we don't need to implement for entry and project files because it will be redundant with at least the [Typescript plugin](https://knip.dev/reference/plugins/typescript).

Fixes https://github.com/webpro-nl/knip/issues/901 https://github.com/webpro-nl/knip/issues/1071


packages/knip/test/plugins/biome.test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+test('Find dependencies with the biome plugin', async () => {
+  const { issues } = await main({
+    ...baseArguments,
+    cwd,
+  });
+
+  assert(issues.devDependencies['package.json']['@org/unused-config'])
+  assert(issues.unresolved['biome.json']['./shared/non-exist-base.json'])
+  assert(issues.unlisted['biome.json']['@org/unlisted-configs'])
```
webpro: The idea of plugin tests is to have no issues left and `counters` with only `total` and `processed` (see other plugin tests for inspiration please).

packages/knip/src/plugins/biome/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+
+const resolveExtends = (extendsArray: string[], options: PluginOptions): Input[] => {
+  const inputs = [] as Input[];
+  for (let item of extendsArray) { 
+    // https://biomejs.dev/guides/configure-biome/#share-a-configuration-file
+    if (item.endsWith(".json")) {
+      inputs.push(toDeferResolve(item));
+    } else {
+      if (require.resolve(item, { paths: [options.cwd] })) {
```
webpro: Ideally, plugins aren't concerned with doing any resolving like this themselves

packages/knip/src/plugins/biome/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const config: string[] = ['biome.json'];
+
+
+const resolveExtends = (extendsArray: string[], options: PluginOptions): Input[] => {
+  const inputs = [] as Input[];
+  for (let item of extendsArray) { 
+    // https://biomejs.dev/guides/configure-biome/#share-a-configuration-file
+    if (item.endsWith(".json")) {
+      inputs.push(toDeferResolve(item));
```
webpro: Maybe we mean to use `toConfig()` here if they're files we also like to go through the same `resolveConfig()` function?

packages/knip/src/plugins/biome/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+  return inputs
+}
+
+
+const resolveConfig: ResolveConfig<BiomeConfig> = (config, options) => {
+  return [
+    // If we're using biome plugin, biome should be ignored in the report
+    toIgnore("@biomejs/biome", "dependencies"),
```
webpro: Isn't the `biome` executable expected in e.g. some `package.json#scripts`? That would add a reference to the `bin` of the `@biomejs/biome` package. For instance, Knip itself has this too: https://github.com/thoroc/knip/blob/22739cee31e0845d8e73aeb26c77a0e75a7e1e71/package.json#L23-L25

packages/knip/src/plugins/biome/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const title = 'biome';
+
+const enablers = ['@biomejs/biome'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => 
+  {
+    const enabled = hasDependency(dependencies, enablers);
+    return enabled;
+  }
```
webpro: Formatting feels a bit off, please make sure to run `bun run format`

packages/knip/src/index.ts
```diff
@@ -172,7 +172,7 @@ export const main = async (unresolvedConfiguration: CommandLineOptions) => {
   }
 
   if (!isWatch) streamer.clear();
-
+
```
webpro: Please undo unrelated change

packages/knip/fixtures/plugins/biome/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "noFallthroughCasesInSwitch": true,
+    "noUncheckedIndexedAccess": true,
+    "noImplicitOverride": true,
+
+    // Some stricter flags (disabled by default)
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
+    "noPropertyAccessFromIndexSignature": false
+  }
+}
```
webpro: Let's try to keep fixtures as small and minimal as possible. Do we need this file? If so, could it be smaller?

packages/knip/fixtures/plugins/biome/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@plugins/biome",
+  "module": "index.ts",
+  "type": "module",
+  "private": true,
+  "devDependencies": {
+    "@biomejs/biome": "^1.9.4",
+    "@types/bun": "latest",
+    "knip": "^5.59.1",
```
webpro: Do we need all these dependencies here? Let's try to keep fixtures as small and focused as possible.

webpro (commented):
Thanks, Vinh! Almost there, let's make it happen.

packages/knip/test/plugins/biome.test.ts
tidefield: If we expect no issues, how do we test unused and unresolved dependencies which is the spirit of `knip` in the first place?

packages/knip/src/plugins/biome/index.ts
tidefield: Hmm what is your recommendation to support module-based `extends` in biome config?

https://biomejs.dev/guides/configure-biome/#share-a-configuration-file

packages/knip/test/plugins/biome.test.ts
tidefield: @webpro I'm not sure if I'm misunderstanding something. I've skimed through other tests and see lots of the tests assert on `issues` which is the same as this test. Can you help me understand this a bit more?

packages/knip/src/plugins/biome/index.ts
tidefield: yeah that's true. let me fix this

packages/knip/src/plugins/biome/index.ts
tidefield: Yeah I must have misread it somewhere and ran `npm run format` from the root folder instead. I just ran `bun run format` now.

packages/knip/src/plugins/biome/index.ts
tidefield: The idea here was that I wanted to report the case when the config extends non-existent `json`. If I use `toConfig()`, the report won't list that.

However, I feel like that error will be detected by `biome` already. So I switched to using `toConfig`.

packages/knip/fixtures/plugins/biome/tsconfig.json
tidefield: Yeah we don't need this file. I removed it.

packages/knip/fixtures/plugins/biome/package.json
tidefield: Yeah. I removed the redundant deps.

tidefield:
Thanks @webpro for the review. I left two questions in the threads and will work on addressing https://github.com/webpro-nl/knip/pull/1113#discussion_r2120833476 as soon as I have time.

packages/knip/test/plugins/biome.test.ts
webpro: Most of the plugin tests are pretty much "end to end" tests in that the fixture folder is consumed in its entirety including `package.json` that should contain the dependencies referenced in the tool configuration, and also executables. E.g. `node_modules/@biomejs/biome/package.json` contains `bin` field for `biome` binary that should actually be referenced and result in zero issues left. This full-circle approach results in well-defined and maintainable tests.

Testing that source code or configuration results in unused or unresolved can happen in additional/other tests. There is definitely room for improvement, but this is the direction I like to go.

packages/knip/src/plugins/biome/index.ts
webpro: We can use `toConfig()` for this (https://knip.dev/guides/writing-a-plugin#toconfig, while I realize the docs say "by a different plugin" this could also be another config file but the same plugin)

webpro: The TypeScript plugin does basically the same thing: https://github.com/webpro-nl/knip/blob/18842738a4acc06e11654a853c5247f0e8399d27/packages/knip/src/plugins/typescript/index.ts#L22-L26

packages/knip/src/plugins/biome/index.ts
webpro: Currently formatting is still off. Please run `bun run format` from the root so that the formatter uses the root `biome.json`. Sorry there's something I need to fix up here I guess.

packages/knip/test/plugins/biome.test.ts
tidefield: Ahh this is a really cool way of looking at testing. It was counter-intuitive to me at first but then it finally clicked. I updated the test. PTAL.

packages/knip/src/plugins/biome/index.ts
tidefield: Done

packages/knip/src/plugins/biome/index.ts
tidefield: Should be good now. I think it was my fault.

tidefield:
@webpro I addressed all your comments and offered a bonus PR https://github.com/webpro-nl/knip/pull/1123 😊

webpro:
Can you please check the changeset carefully, there's a lot of unrelated changes.

tidefield:
> Can you please check the changeset carefully, there's a lot of unrelated changes.

Ooops. I must have missed something when syncing with the latest `main` branch. I just reverted them.

packages/knip/src/plugins/biome/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const resolveConfig: ResolveConfig<BiomeConfig> = (config, options) => {
+  return [...resolveExtends(config.extends || [], options)];
+};
+
+const resolve: Resolve = options => {
+  const { manifest } = options;
+  const inputs: Input[] = [];
+  if (Object.values(manifest.scripts || {}).some(script => script.includes('biome'))) {
+    inputs.push(toDependency('@biomejs/biome'));
```
webpro: With https://github.com/webpro-nl/knip/pull/1113#discussion_r2120042883 we should also be able to remove this. Which essentially means we can remove the whole `resolve` function.

tidefield: Done. PTAL

packages/knip/src/plugins/biome/index.ts
```diff
@@ -0,0 +1,36 @@
+import type { IsPluginEnabled, Plugin, PluginOptions, ResolveConfig } from '../../types/config.js';
+import { type Input, toConfig } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { BiomeConfig } from './types.js';
+
+const title = 'biome';
```
webpro:
```suggestion
const title = 'Biome';
```

packages/knip/src/plugins/biome/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const enablers = ['@biomejs/biome', 'biome'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => {
+  const enabled = hasDependency(dependencies, enablers);
+  return enabled;
+};
```
webpro:
```suggestion
const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
```

webpro (approved):
Just two nitpicks to wrap it up!

webpro:
:rocket: _This pull request is included in v5.61.0. See [Release 5.61.0](https://github.com/webpro-nl/knip/releases/tag/5.61.0) for release notes._

PR #1116, opened 2025-06-03
Hardhat plugin updates

ItsNickBarry:
#1087 adds basic support for Hardhat, which is suitable for developing Hardhat plugins.  However, it lacks support for most situations faced by consumers of Hardhat.

* [x] detect imports from `dependencesToCompile` in `hardhat.config.*`
* [ ] detect `*.sol` imports from `*.sol` files
* [ ] apply remappings to `*.sol` imports
* [ ] detect test files if a Hardhat test runner plugin is installed
* [ ] check formatting

### DIFF
--- a/packages/knip/src/plugins/hardhat/index.ts
+++ b/packages/knip/src/plugins/hardhat/index.ts
@@ -17,8 +17,6 @@ const enablers = ["hardhat"];
 const isEnabled: IsPluginEnabled = ({ dependencies }) =>
   hasDependency(dependencies, enablers);
 
-const entry: string[] = ["hardhat.config.{js,cjs,mjs,ts}"];
-
 const config = ["hardhat.config.{js,cjs,mjs,ts}"];
 
 const resolve: Resolve = async () => {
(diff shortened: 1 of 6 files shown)


packages/knip/src/plugins/hardhat/index.ts
```diff
@@ -1,25 +1,58 @@
...
+  hasDependency(dependencies, enablers);
 
-const entry: string[] = ['hardhat.config.{js,cjs,mjs,ts}'];
+const config = ["hardhat.config.{js,cjs,mjs,ts}"];
 
 const resolve: Resolve = async () => {
-  return [toDependency('hardhat')];
+  const inputs = [toDependency("hardhat")];
+
+  // TODO: only add test files if a Hardhat test reporter is installed
```
ItsNickBarry, line 25: Is it possible to check for the presence of other packages here, or does that require a separate plugin?

webpro:
What's the status? Is there still interest in wrapping this up?

ItsNickBarry:
No.

PR #1121, opened 2025-06-05
Fix Node test runner detection

### LINKED ISSUE #1120: 💡 Entry files when using `node --test`
### Suggest an idea for Knip

When using Node's built-in test runner (`node --test`), [test files](https://nodejs.org/api/test.html#running-tests-from-the-command-line) should be treated as entry files.

I have absolutely no idea how you would enable that automatically though, because there's no evidence of the user using that runner in the codebase (like a dev dependency).

MidnightDesign:
- Fixes a bug where previously the `"script"` keys were checked for `node --test` instead of the values
- Update glob patterns to match TypeScript files

Fixes #1120

### DIFF
--- a/packages/knip/src/plugins/node/index.ts
+++ b/packages/knip/src/plugins/node/index.ts
@@ -15,8 +15,14 @@ const resolveConfig: ResolveConfig<PackageJson> = localConfig => {
 
   const entries = [toProductionEntry('server.js')];
 
-  if (scripts && Object.keys(scripts).some(script => /(?<=^|\s)node\s(.*)--test/.test(scripts[script]))) {
-    const patterns = ['**/*{.,-,_}test.?(c|m)js', '**/test-*.?(c|m)js', '**/test.?(c|m)js', '**/test/**/*.?(c|m)js'];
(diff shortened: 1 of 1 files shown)


webpro:
Thanks Rudolph, happy to merge this. I'll sort out the test failures.

webpro:
There actually wasn't an issue with `keys` vs `values` - but perhaps you needed the `.ts` extensions?

MidnightDesign:
> There actually wasn't an issue with `keys` vs `values` - but perhaps you needed the `.ts` extensions?

Ooooh you're right. It was just an awkward way to get at the values. : )

Then yes, it was definitely the `.ts` extensions. 

Thanks!

MidnightDesign:
Feel free to rename the PR so it fits its new purpose.

webpro:
Good point. Hopefully it wasn't my finest work indeed 😬

webpro:
:rocket: _This pull request is included in v5.60.2. See [Release 5.60.2](https://github.com/webpro-nl/knip/releases/tag/5.60.2) for release notes._

PR #1125, opened 2025-06-09
Restructure biome configurations

daivinhtran:
Currently, we allow `format` script inside `packages/knip` but the default config in `knip` package is not on par with the root `biome.json`. This can lead to unrelated diff as in https://github.com/webpro-nl/knip/pull/1113/commits/ba0c5ca23c1ebb4dffb5a531c638598bf0eb980d.

As a plugin writer, I needed to run `bun test test/plugins/biome.test.ts` inside `packages/knip`. Unfortunately, I had to run `bun run format` at the root folder to avoid the unintended diff. It mandates an extra step in the developer workflow.

This PR is to improve the ergonomic when running `bun run format` inside `packages/knip`. Moving forward, we can define `packages/knip`-specific biome config inside `package/knip/biome.json`.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,13 +1,8 @@
 {
   "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
+  "extends": ["packages/knip/biome.json"],
   "files": {
-    "ignore": [
-      "**/dist",
-      "**/tmp",
-      "packages/knip/package.json",
-      "packages/knip/vendor/bash-parser/index.js",
-      "packages/docs/.astro"
-    ]
+    "ignore": ["**/dist", "**/tmp", "packages/docs/.astro"]
   },
   "organizeImports": {
     "enabled": true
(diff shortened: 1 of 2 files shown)


biome.json
```diff
@@ -48,10 +43,7 @@
     "enabled": true,
     "lineWidth": 120,
     "indentStyle": "space",
-    "formatWithErrors": true,
```
tidefield, line 51: I removed the ignore and reran the formatter. It didn't result to any fixes so I think we're good to remove it.

tidefield: Oh because of https://github.com/webpro-nl/knip/commit/b346903187cebf19fc0516d2ea46bd8dec7d4dfe, it now fails :D

webpro:
Thanks Vinh! Very happy with this one 🙏

webpro:
:rocket: _This pull request is included in v5.61.0. See [Release 5.61.0](https://github.com/webpro-nl/knip/releases/tag/5.61.0) for release notes._

PR #1132, opened 2025-06-11
Add plugin for Hey API

narthur:
This PR adds a plugin for Hey API.

https://heyapi.dev/


narthur:
To pass my tests, since I'm  using `toDeferResolve`, I committed my fixture's node_modules folder, but this feels weird. Is this the correct way to handle this? Should I instead hand-craft the files in my node_modules folder to keep things minimal?

webpro:
Yeah please see other plugin fixtures

narthur:
@webpro Should be better now!

packages/knip/src/plugins/hey-api/index.ts
```diff
@@ -0,0 +1,32 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import { type Input, toDeferResolve, toEntry } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { PluginConfig } from './types.js';
+
+// https://heyapi.dev/openapi-ts/get-started
+
+const title = 'hey-api';
```
webpro:
```suggestion
const title = 'Hey API';
```

packages/knip/src/plugins/hey-api/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const enablers = [/^@hey-api\//];
```
webpro:
```suggestion
const enablers = ['@hey-api/openapi-ts'];
```

packages/knip/src/plugins/hey-api/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['openapi-ts.config.@(ts|cjs|mjs)'];
```
webpro:
```suggestion
const config: string[] = ['openapi-ts.config.@(js|ts|cjs|mjs)'];
```

packages/knip/src/plugins/hey-api/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const resolveConfig: ResolveConfig<PluginConfig> = async (config): Promise<Input[]> => {
+  const plugins = (config.plugins ?? []).map(plugin => {
+    const pluginName = typeof plugin === 'string' ? plugin : plugin.name;
+    return toDeferResolve(pluginName);
```
webpro: If the plugins are always dependencies (i.e. not local files) we can use `toDependency` (it's faster as there's no need to resolve anything on the fs).

packages/knip/src/plugins/hey-api/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+  const outputPath = typeof config.output === 'string' ? config.output : config.output.path;
+  const ignored = [`./${outputPath}/**/*`].map(toEntry);
+  return [...plugins, ...ignored];
```
webpro: If the intention is to add entry file(s) we can improve a bit:

```suggestion
  const entries = outputPath ? [toEntry(`./${outputPath}/**`)] : [];
  return [...plugins, ...entries];
```

packages/knip/src/plugins/hey-api/index.ts
narthur: Done!

narthur:
@webpro I'm not sure how to debug this CI error, and I don't see that I have the option to rerun the failed checks.

```
Run npm install typescript@5.0.4 @types/node@18 --ignore-scripts
npm error code E[4](https://github.com/webpro-nl/knip/actions/runs/15637574744/job/44056620965?pr=1132#step:4:5)00
npm error 400 Bad Request - GET https://registry.npmjs.org/@release-it%2fbumper - Bad request
npm error A complete log of this run can be found in: /Users/runner/.npm/_logs/202[5](https://github.com/webpro-nl/knip/actions/runs/15637574744/job/44056620965?pr=1132#step:4:6)-06-13T14_57_16_228Z-debug-0.log
Error: Process completed with exit code 1.
```

https://github.com/webpro-nl/knip/actions/runs/15637574744/job/44056620965?pr=1132

packages/knip/src/plugins/hey-api/index.ts
```diff
@@ -0,0 +1,42 @@
...
+const config: string[] = ['openapi-ts.config.@(js|ts|cjs|mjs)'];
+
+const resolveConfig: ResolveConfig<PluginConfig> = async (config): Promise<Input[]> => {
+  const plugins = (config.plugins ?? []).reduce<Input[]>((acc, plugin) => {
+    if (typeof plugin === 'string') {
+      acc.push(toDependency(plugin));
+      return acc;
+    }
+
+    const dependencies = plugin._dependencies ?? [];
```
webpro: Is there documentation or code available regarding this `_dependencies` property?

webpro (commented):
Is there a public repo available where we can exercise this plugin to make sure loading `config` files works fine and we're not missing something obvious?

packages/knip/src/plugins/hey-api/index.ts
narthur: Hmm. When I look up types in my local machine, I find this:

```ts
interface Meta<Config extends BaseConfig> {
  /**
   * Dependency plugins will be always processed, regardless of whether user
   * explicitly defines them in their `plugins` config.
   */
  _dependencies?: ReadonlyArray<AnyPluginName>;
  // ...
}
```

But when I look in their repo I find [this](https://github.com/hey-api/openapi-ts/blob/fcb77c9661707e259721ad1fc7a44b75cc027fba/packages/openapi-ts/src/plugins/types.d.ts#L62):

```ts
interface Meta<Config extends BaseConfig> {
  /**
   * Dependency plugins will be always processed, regardless of whether user
   * explicitly defines them in their `plugins` config.
   */
  dependencies?: ReadonlyArray<AnyPluginName>;
  // ...
}
```

When I look at [their docs](https://heyapi.dev/openapi-ts/plugins/custom) it shows `dependencies` being set in the custom plugin's config, but it also says this:

> Reserved fields are prefixed with an underscore and are not exposed to the user.

So as best I can tell, it's specified as `dependencies` in the custom plugin's configuration, and then is changed to `_dependencies` when the plugin is loaded in.

That said, since this is all a bit vague, and their docs says the custom plugin API is currently unstable, maybe I should remove the custom plugin support from this PR for now?

webpro: > maybe I should remove the custom plugin support from this PR for now?

Not sure. Is there a public repo available where we can exercise this plugin to make sure loading config files works fine and we're not missing something obvious?

narthur:
> Is there a public repo available where we can exercise this plugin to make sure loading `config` files works fine and we're not missing something obvious?

Just created one:

https://github.com/narthur/hey-api-demo

I included a custom plugin in that repo. Its types forced me to use `_dependencies`, not `dependencies`, so not sure what to make of that--maybe the docs are out of date?

I also discovered that in some places in their docs they show including default plugins when adding your own plugins:

https://heyapi.dev/openapi-ts/plugins/tanstack-query#installation

Which means I need to figure out how to make the knip plugin handle that without flagging those default dependencies as unlisted.

narthur:
Trying to think this through... feels like a bit of a catch-22. If the user interpolates default plugins and I allow them through as usual, they may be caught as unlisted. If I instead filter out default plugins from the user's provided plugins, then they may end up flagged as unused if the user also includes them in their direct dependencies.

I'm leaning toward just letting them through and allowing them to be caught as unlisted.

- The user is explicitly providing that list of default plugins, so it seems arguably correct to call them unlisted if the user does this and doesn't include them in direct dependencies.
- Handling it this way is the simplest way when it comes to implementation.

webpro:
We could do something like this for the default plugins:

```ts
  const defaultPlugins = ['@hey-api/typescript', '@hey-api/sdk'];
  const plugins = (config.plugins ?? []).reduce<Input[]>((acc, plugin) => {
    if (typeof plugin === 'string') {
      if (!defaultPlugins.includes(plugin)) acc.push(toDependency(plugin));
      return acc;
    }
```

From a Knip perspective, are `_dependencies` transitive dependencies? I.e. we can ignore those?

narthur:
@webpro Seems like it's working well now, both by test and demo. Maybe ready for another review?

webpro:
There's quite a few projects using it seems: https://github.com/hey-api/openapi-ts/network/dependents, we don't we check some of those out and smoke-test a bit? My guess is that this might shed some light to some of the doubts we're having?

narthur:
Hmm, interesting. I wonder if there's a way I can filter that list for projects that use Hey API.

narthur:
Looks like this is the search to do that:

https://github.com/search?q=knip+%40hey-api%2Fopenapi-ts+path%3A**%2Fpackage.json&type=code

webpro:
The page I linked to is the list of projects using Hey API which this PR adds a plugin for. If any of those projects is already using Knip that's cool, but what we're mostly interested in is the configuration any project using Hey API does **not** need with this Knip plugin added. If a project is using Hey API and Knip, and assuming the Knip configuration was good, then what we want to see is the Knip configuration they can remove due to the new plugin.

To this end, I'll test drive a few projects in the coming days.

narthur:
@webpro Gotcha, ok. I've reverted my additions to the automated tests since it sounds like that wasn't the correct way to do this.

narthur:
For what it's worth, I've installed this version of Knip in the project that initially prompted me to make this PR, and it's working well so far.

webpro:
Is there still interest in getting this merged? Please see my previous comments, we'll need some real-world insights to see how Knip behaves in projects using Hey API with this plugin added.

narthur:
I'd still like to get this merged, but I'm no longer actively working on using hey api anywhere, so I guess I'm probably even less help on this than I was when I started the PR. 🫤

webpro:
No worries! I'm going to close this, but you or anyone else is free to submit a pull request whenever the time is right.

Issue #1136, opened 2025-06-14
💡 Support global `gitignore` / `core.excludesFile`


alecmev:
### Suggest an idea for Knip

As discussed here: https://github.com/webpro-nl/knip/issues/531#issuecomment-2969907337 From [gitignore spec](https://git-scm.com/docs/gitignore#_description):

> - Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by `core.excludesFile` in the user’s `~/.gitconfig`. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.
>
> ... Higher-level Git tools, such as **git status** and **git add**, use patterns from the sources specified above.

cylewaitforit:
This seems like it would cause inconsistencies between runs. If knip looked at files outside the repo it might become more flakey since what one machine has in its global ignores wouldn’t match another user’s or CI.

Those global ignores seem like they are great to use for ensuring when you create new repos you don’t inadvertently commit those files.

I would say if knip finds these files as issues when you are running it in a repo it is a great indicator that you should be adding it to a repo level gitignore so others pulling down the repo benefit from sharing that configuration.

webpro:
Thanks for the pull request, Alec. Gave it another thought and while following `git check-ignore` might seem like a good idea, that's only on local/user level. It isn't from a project perspective (as @cylewaitforit points out): it can lead to inconsistent behavior across machines/CI.

#531 had that issue as well, and has recently been fixed.

Double-checked that other similar tooling like ESLint or Prettier also doesn't look at the global gitignores.

That's why I'm going to close this one, and close #1137.

So where we currently are:

- From `cwd`, respect `.gitignore` files in same and descendant directories
- From `cwd`, respect `.gitignore` files in ancestor directories up to and including Git root dir
- From `cwd`, respect `.git/info/exclude'` file — this could be improved because file not found if cwd !== Git root dir, but it's bit of an edge case

alecmev:
@cylewaitforit 

> I would say if knip finds these files as issues when you are running it in a repo it is a great indicator that you should be adding it to a repo level gitignore so others pulling down the repo benefit from sharing that configuration.

This was addressed in the top post of the original issue: https://github.com/webpro-nl/knip/issues/531#issue-2161472406

[...]

PR #1137, opened 2025-06-14
Support global gitignore

alecmev:
Finally actually contributing something 😛 Fixes https://github.com/webpro-nl/knip/issues/1136

Submitting as a draft because I'm not sure how'd you like to go about testing this, and don't want to invest time into the wrong direction. I assume mocking `getGitConfigValue` and `process.env` would be your preference?

### DIFF
--- a/packages/knip/src/util/glob-core.ts
+++ b/packages/knip/src/util/glob-core.ts
@@ -8,6 +8,7 @@ import { timerify } from './Performance.js';
 import { compact } from './array.js';
 import { debugLogObject } from './debug.js';
 import { isDirectory, isFile } from './fs.js';
+import { getGitConfigValue } from './git.js';
 import { parseAndConvertGitignorePatterns } from './parse-and-convert-gitignores.js';
 import { dirname, join, relative, toPosix } from './path.js';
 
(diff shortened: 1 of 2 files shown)


packages/knip/src/util/glob-core.ts
```diff
@@ -130,6 +131,17 @@ export const findAndParseGitignores = async (cwd: string) => {
 
   if (isFile('.git/info/exclude')) addFile('.git/info/exclude', cwd);
 
+  const globalExcludesFile = getGitConfigValue('core.excludesFile', undefined, '--global');
+  if (globalExcludesFile !== undefined && isFile(globalExcludesFile)) {
+    addFile(globalExcludesFile, cwd);
+  } else if (typeof process.env.XDG_CONFIG_HOME === 'string' && process.env.XDG_CONFIG_HOME.trim() !== '') {
+    const xdgIgnore = `${process.env.XDG_CONFIG_HOME}/git/ignore`;
+    if (isFile(xdgIgnore)) addFile(xdgIgnore, cwd);
```
alecmev, line 139: Following the logic of the docs here:

> Its default value is `$XDG_CONFIG_HOME/git/ignore`. If `$XDG_CONFIG_HOME` is either not set or empty, `$HOME/.config/git/ignore` is used instead.

Which I read to mean that `isFile(xdgIgnore) === false` means "give up", not "try `$HOME`".

webpro:
Closed by https://github.com/webpro-nl/knip/issues/1136#issuecomment-2975584488

PR #1144, opened 2025-06-20
docs: Tsdoc for knip config

cylewaitforit:

### DIFF
--- a/packages/docs/src/content/docs/reference/configuration.md
+++ b/packages/docs/src/content/docs/reference/configuration.md
@@ -100,7 +100,7 @@ TypeScript semantics:
 - Path values are an array of relative paths.
 - Paths without an `*` are exact matches.
 
-### Plugins
+## Plugins
 
 There are a few options to modify the behavior of a plugin:
 
--- a/packages/docs/src/content/docs/reference/configuration.md
+++ b/packages/docs/src/content/docs/reference/configuration.md
@@ -182,6 +182,16 @@ notation below is valid and will report only exports tagged `@lintignore` or
 
 Also see [JSDoc & TSDoc Tags][8].
 
+### `treatConfigHintsAsErrors`
+
+Exit with non-zero code (1) if there are any configuration hints.
+
+```json title="knip.json"
+{
+  "treatConfigHintsAsErrors": true
+}
+```
+
 ## Ignore Issues
 
 ### `ignore`
--- a/packages/knip/src/schema/configuration.ts
+++ b/packages/knip/src/schema/configuration.ts
@@ -106,14 +106,71 @@ const baseWorkspaceConfigurationSchema = z.object({
   includeEntryExports: z.boolean().optional(),
 });
 
-const projectSchema = z
-  .object({
-    entry: globSchema.optional(),
-    project: globSchema.optional(),
-    paths: pathsSchema.optional(),
-    workspaces: z.record(z.string(), baseWorkspaceConfigurationSchema.merge(pluginsSchema.partial())).optional(),
-  })
-  .merge(pluginsSchema.partial());
+const projectSchema = z.object({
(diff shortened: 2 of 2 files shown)


webpro:
Thanks for the PR! My first comments:

- tbh ideally we'd find a way to have a single source of truth (whichever that is), otherwise it's bound to get out of sync
- `syncCompilers` and `asyncCompilers` are internal and don't need to be documented: `compilers` is meant to be the public facing configuration item (they're split internally automatically)

cylewaitforit:
Thanks @webpro !

>     * `syncCompilers` and `asyncCompilers` are internal and don't need to be documented: `compilers` is meant to be the public facing configuration item (they're split internally automatically)

That makes sense and sounds great. I will start by updating the tsdoc for those and remove them from the docs page.

>     * tbh ideally we'd find a way to have a single source of truth (whichever that is), otherwise it's bound to get out of sync

Totally, I felt the same way when I started tackling this. I had put doc generation in the "Likely Out of Scope" section of the PR description simply because the related issues only requested the IDE hints. I didn't want to overhaul anything too much on the docs side if that wasn't desired. If it is a blocker to solving the IDE hints, like it sounds it might be, I'll move this back to draft and work further on the doc generation piece of the puzzle.

packages/docs/src/content/docs/reference/configuration.md
cylewaitforit, line 1: @webpro I can break these changes to the `configuration.md` into a separate PR if it is more helpful for them to land sooner or be reviewed independent of the tsdoc.

webpro: Feels like it makes sense to break it up, but I'm happy to review whatever is more convenient for you :)

cylewaitforit: Moved the changes to #1152

packages/docs/.config/typedoc.json
cylewaitforit, line 1: @webpro I added typedoc to this PR to generate the configuration docs from the TSdoc comments. This is the first time I've ever tried or used typedoc so it would not hurt my feelings at all if this feels like the wrong direction. Your honest opinion (https://youtu.be/2q0lsBgOTJs?si=3Cl_RYlOeK_kqHci) is appreciate.

webpro: _My honest opinion.._ 😃 is that I **love automation** and I really appreciate the pull request to actually see and feel TSDoc, this feels like losing too much control over the output. The few minor issues I have with it (which tbh is basically most output that's not like the original) do add up and I don't think we'll be able to get it to a satisfactory level without "fighting" the tool too much. As it stands, I'd rather see the project managing things cleanly in two locations over introducing TSDoc plus configuration.

cylewaitforit: I think I feel the same way.  Autodocs generation from JSDoc/TSDoc comments via something like typedoc is still something I’d personally like to explore further in some project. I wasn’t soured by the experience. I especially feel positive about using typedoc to generate markdown.(By default, without `typedoc-plugin-markdown`, it creates its own standalone html site. I think Starlight is nicer.)

Maybe starting something greenfield or as part of a new feature it could be easier to accept some of the shortcomings or architect the api to be easier to document with typedoc. Trying to shoehorn it in here probably left too many rough edges. Especially since I am not a typedoc wizard.

I’ll close this PR in it’s current state so that it exists for posterity and try in the next few days to take a round two at just adding the JSDoc/TSDoc comments per the original issue’s request.

webpro: It's definitely an interesting tool. I guess introducing it in an existing project that has manually written docs might be quite the challenge when it comes to preserving end-user UX. But the idea is sound, co-location etc.

PR #1161, opened 2025-07-02
docs: only add TSDoc for configuration

### LINKED ISSUE #892: 💡 Add description to `KnipConfig ` types
### Suggest an idea for Knip

I would like to use `KnipConfig ` types in a configuration file, so the IDE could hint me more accurately.

```js
/** @type { import('knip'). KnipConfig } */
const config = { ... };
export default config;
```

But without at least description, it is not that helpful.

It would be nice if you add description for every type in `KnipConfig`:
https://github.com/webpro-nl/knip/blob/9590b9238d558098430f9535f95c1f40d437c2ec/packages/knip/src/ConfigurationValidator.ts#L47-L62

It's probably worth to add `@default` and `@see`. See:

- https://github.com/vitejs/vite/blob/9290d85b5d2ad64991bd296157cb3bcb959c341d/packages/vite/src/node/config.ts#L296
- https://github.com/eslint/eslint/blob/5db226f4da9ad7d53a4505a90290b68d4036c082/lib/types/index.d.ts#L1254
- https://github.com/prettier/prettier/blob/d9d0999aac70c04b06b35cbcb2c804b4a9c8d8a8/src/index.d.ts#L308

cylewaitforit:
Adds [tsdoc](https://tsdoc.org/) for showing schema documentation in IDE.

## To do

- [x] Add TSDoc to rootConfigurationSchema
- [x] Add appropriate @default tags

## Likely out of scope

- Add TSdoc for all config in the Plugins
  - Defined in a different file and need to understand better first how the plugin generation process might be affected.

Closes #892

### DIFF
--- a/packages/knip/src/schema/configuration.ts
+++ b/packages/knip/src/schema/configuration.ts
@@ -45,28 +45,332 @@ const ignoreExportsUsedInFileSchema = z.union([
 ]);
 
 const rootConfigurationSchema = z.object({
+  /**
+   * A `$schema` field is a URL that you put at the top of your JSON file. This
+   * allows you to get red squiggly lines inside of your IDE when you make a typo or
+   * provide an otherwise invalid configuration option.
+   *
+   * @default undefined
+   *
+   * @example
+   * In JSON, use the provided JSON schema:
+   *
+   * ```json title="knip.json"
+   * {
+   *   "$schema": "https://unpkg.com/knip@5/schema.json"
+   * }
+   * ```
+   *
+   * @example
+   * In JSONC, use the provided JSONC schema:
+   * ```jsonc title="knip.jsonc"
+   * {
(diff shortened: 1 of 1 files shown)


cylewaitforit:
@webpro This should be ready for a critical eye. 

It's much scaled back from #1144, just adding the TSDoc/JSDoc style comments for use in IDE's for now. None of the refactoring or doc generation.

cylewaitforit:
@webpro Is this worth moving forward on or do you think it would be best to address #892 in some other way?

webpro:
Ah yeah sorry for not getting back to this one. Appreciate you've put the work in this one and it's nice for users, but not great for maintainers. Still, I'm going to merge which will hopefully be a motivation to come up with some (custom) solution/script.

Thank you very much! Much appreciated.

webpro:

PR #1163, opened 2025-07-02
Handle SvelteKit service worker integration

gwennlbh:
Probably needs some more code [in case the service worker path is changed in the sveltekit config](https://svelte.dev/docs/kit/configuration#files:~:text=app%20(see%20Routing)-,serviceWorker,-?:%20string;) (to change the entry), or [automatic sw registration is disabled](https://svelte.dev/docs/kit/configuration#serviceWorker) (to remove the entry)

Relevant config pages: https://svelte.dev/docs/kit/service-workers (general info) and https://svelte.dev/docs/kit/$service-worker (for the `$service-worker` alias)

### DIFF
--- a/packages/knip/src/plugins/svelte/index.ts
+++ b/packages/knip/src/plugins/svelte/index.ts
@@ -18,11 +18,13 @@ const production = [
   'src/routes/**/+{page,server,page.server,error,layout,layout.server}{,@*}.{js,ts,svelte}',
   'src/hooks.{server,client}.{js,ts}',
   'src/params/*.{js,ts}',
+  'src/service-worker{.js,.ts,/index.{js,ts}}',
 ];
 
 const resolve: Resolve = options => {
-  const alias = toAlias('$app/*', [join(options.cwd, 'node_modules/@sveltejs/kit/src/runtime/app/*')]);
-  return [alias];
+  const app = toAlias('$app/*', [join(options.cwd, 'node_modules/@sveltejs/kit/src/runtime/app/*')]);
+  const sw = toAlias('$service-worker', [join(options.cwd, 'node_modules/@sveltejs/kit/types/index.d.ts')]);
+  return [app, sw];
 };
 
 export default {


webpro:
Thanks Gwenn!

If the service worker location depends on the local sveltekit config, then `resolveConfig` can be added to the plugin, and then conditionally have it return the entry (this config is the first argument of `resolveConfig(config, options) {}`.

Would also be great if you could add a fixture or two to exercise the updated plugin a bit in a new or existing test.

gwennlbh:
How should I go about getting the value of the service worker path from the local config? From what I see in the Vite plugin, you use static analysis instead of running, in that case, svelte.config.[jt]s ? I guess this would prevent dynamic setups from working, which is fine imo

Considering that there are multiple ways to define the config (a const declaration then a export default config statement, as I saw in the fixture; a direct export default statement, ...) I wonder if you have utils already in place to help with that, so that I don't reinvent some wheels?

webpro:
For now (not behind a computer), I can say that Knip actually loads the config file and the function receives the exported resolved object. Hopefully that's mentioned in the docs somewhere ("writing a plugin"). There is resolveAST though, if necessary (but rarely the case).

gwennlbh:
ok so, i think I got it. sorry for not checking out the contributing.md, i was searching for it at /, didn't see it under .github/ '^^

I still get the unused file error when running ` bun test tests/plugins/svelte.test.ts` , I don't really know why. console.log doesn't seem to work either, so I can't see what's wrong in my resolveConfig logic '^^

EDIT: If I hardcode src/sw.ts in `const production = [` it still shows an error... what am I doing wrong?

webpro:
Please see the following docs for more details & tips:

- https://github.com/webpro-nl/knip/blob/main/.github/DEVELOPMENT.md
- https://knip.dev/guides/writing-a-plugin

webpro:
Is there still interest in wrapping this up? Happy to help out.

webpro:
Closing this with c0ed40a

PR #1165, opened 2025-07-07
Update reporters to have the correct name for the default reporter

gavinhenderson:
In the docs currently it says the default reporter is `symbol`. This is not true, if you do `--report symbol` you get an error. 

After looking at the code it turns out that the reporter is actually called `symbols`. I have updated the docs so if you are reading the docs you will see the correct reporter name.

Thanks for this brilliant project!


webpro:
Thank you very much, Gavin!

webpro:

Issue #806, opened 2024-10-08
💡 Add `eslint-import-resolver-<resolver-name>` support


ronbraha:
### Suggest an idea for this project

While onboarding Knip on one of the business repositories, I noticed that while the Knip ESLint plugin has broad support for automatically picking up dependencies that ESLint uses (especially with the new flat config), it still does not recognize import resolvers.
Is that something you considered adding support for?

Current day, I have to add `eslint-import-resolver-webpack` to `ignoreDependencies` manually.

guillaumebrunerie:
Same problem here, it worked before with a `.eslintrc` but then stopped working when I upgraded my ESLint config to a flat config.

Can be reproduced with the following `eslint.config.js` in a new project:
```
npm create vite@latest knip-import-resolver-test -- --template vanilla-ts`
cd knip-import-resolver-test
npm install eslint @eslint/js typescript-eslint eslint-plugin-import eslint-import-resolver-typescript knip
```
```js
// eslint.config.js
import importPlugin from 'eslint-plugin-import';
import js from '@eslint/js';
import tseslint from "typescript-eslint";

export default [
	js.configs.recommended,
	...tseslint.configs.recommended,
	importPlugin.flatConfigs.recommended,
	{
		files: ['**/*.{ts,tsx}'],
		settings: {
			"import/parsers": {
				"@typescript-eslint/parser": [".ts", ".tsx"],
			},
			"import/resolver": {
				typescript: true,
				node: true,
			},
		},
		rules: {
			/* [...] */
		},
	},
];
```
```
> npx knip
Unused dependencies (1)
eslint-import-resolver-typescript  package.json
```

With the corresponding `.eslintrc` instead, the `eslint-import-resolver-typescript` dependency is correctly detected:
```js
// .eslintrc
{
	"parser": "@typescript-eslint/parser",

	"settings": {
		"import/parsers": {
			"@typescript-eslint/parser": [".ts", ".tsx"]
		},
		"import/resolver": {
			"typescript": true,
			"node": true,
		},
	},

	"plugins": [
		"import",
		"@typescript-eslint",
	],

	"extends": [
		"eslint:recommended",
		"plugin:import/recommended",
		"plugin:import/typescript",
		"plugin:@typescript-eslint/recommended",
	],
}
```

webpro:
Thanks for using Knip everyone, but please create an actual issue reproduction. Or rather, open a pull request.

[...]

PR #1173, opened 2025-07-12
Mention `eslint-import-resolver-typescript` in docs

fregante:
`eslint-plugin-import` and `eslint-plugin-import-x` don't need this config, they pick up the resolver automatically, hence the confusion by many in:

- https://github.com/webpro-nl/knip/issues/806

---

By the way, the "Edit page" link in https://knip.dev/reference/plugins/eslint#generated-from-source is a 404. Maybe it should be removed or pointed to the same link as [_"eslint plugin source code"_](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/eslint/index.ts))

### DIFF
--- a/packages/knip/src/plugins/eslint/index.ts
+++ b/packages/knip/src/plugins/eslint/index.ts
@@ -59,9 +59,7 @@ consider using an extended glob pattern like this:
 
 ### \`eslint-import-resolver-typescript\`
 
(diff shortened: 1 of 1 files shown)


webpro:
Thanks! I think it would be even better to just not bother users anymore about such dependencies at all. Recently all ESLint flat config have become optional. Hopefully the attached commit settles it. We can add more as the need arises. Added a fixture (eslint5) based on https://github.com/un-ts/eslint-plugin-import-x?tab=readme-ov-file#typescript-example.

fregante:
Thanks! That should fix it

PR #1174, opened 2025-07-12
Add option to ignore class member implementations

Desuuuu:
### Motivation

At the moment, abstract class members that have an implementation are never detected as unused because the implementation itself is returned in the references for the abstract class member.

```ts
export abstract class AbstractClass {
  public abstract used(): void;
  public abstract unused(): void;
}

export class ExtendedClass extends AbstractClass {
  public used(): void { }
  public unused(): void { }
}
```

```ts
import { AbstractClass, ExtendedClass } from './definitions';

const instance: AbstractClass = new ExtendedClass();

instance.used();
```

In this example, neither `unused` method would be detected as unused.

### Changes

This PR adds the `ignoreClassMemberImplementations` option as well as the `--ignore-class-member-implementations` CLI flag.

When enabled, implementations are ignored when looking for references of class members, allowing for the removal of unused class members which are never directly referenced.

In the previous example, only `AbstractClass.unused` would be detected as unused at first. If it is removed, then `ExtendedClass.unused` would be detected as unused.


webpro:
Thanks for the PR! I think this should be default behavior, and thus we wouldn't even need the flag. As this would basically eliminate the need for the current/old behavior, WDYT?

Desuuuu:
This being the default behavior does make sense to me, I just wasn't sure whether the current one was valuable to someone else.

Let me know if you'd like me to make the change!

webpro:
I do appreciate the initial considerate take! But yes, would be great if you could make it the default behavior instead.

Desuuuu:
Done, got rid of the CLI flag and option!

packages/knip/src/ProjectPrincipal.ts
```diff
@@ -260,14 +261,30 @@ export class ProjectPrincipal {
...
       this.findReferences = timerify(languageService.findReferences);
+      this.getImplementationAtPosition = timerify(languageService.getImplementationAtPosition);
     }
 
     return members.filter(member => {
       if (member.jsDocTags.has(PUBLIC_TAG)) return false;
-      const referencedSymbols = this.findReferences?.(filePath, member.pos) ?? [];
+      const implementations =
+        this.getImplementationAtPosition?.(filePath, member.pos)?.filter(
+          impl => impl.fileName !== filePath || impl.textSpan.start !== member.pos
```
webpro, line 274: Being in the same file should not be a differentiator, right? Especially the abstract class and the class implementing it might be in the same file.

packages/knip/fixtures/class-members/index.ts
```diff
@@ -27,3 +26,5 @@ instance.bUsedExternal;
 instance.cUsedExternal;
 instance.dUsedExternal();
 MyClass.eUsedExternal;
+instance2.implemented;
+instance3.implemented;
```
webpro, line 30: ✅ If I comment out both references to `implemented` here, then both are flagged as unused as I would expect.

❌ Likewise, unmodified I would expect that `implemented` of both `AbstractClass` and `AbstractClassGen` would be referenced/used. But the `implemented` member of `AbstractClass` is flagged as unused (made explicit in the tests). Is this an issue with the implementation or with my understanding of the intent here?

packages/knip/src/ProjectPrincipal.ts
Desuuuu: `getImplementationAtPosition` can return the location of the symbol at the requested position itself in some cases, which this is excluding.

For example, on line 42 of `packages/knip/fixtures/class-members/members.ts`, getting the implementations of `usedGetter` returns itself, and that would exclude anything referencing that symbol.

packages/knip/fixtures/class-members/index.ts
Desuuuu: The key difference here is that `instance2` has type `ExtendedClass` (the implementing class) while `instance3` has type `AbstractClassGen` (the abstract class).

In the first case, this results in the method being flagged as unused in the abstract class since it is never directly referenced, it would not be flagged in the implementing class though.

packages/knip/src/ProjectPrincipal.ts
webpro: Ah little brain cloud, this is same as `!(file && pos)`

packages/knip/fixtures/class-members/index.ts
webpro: Gotcha! I think we when playing with it I had the `instance3: AbstractClassGen` override removed. Thanks for the explanation, all seems to work as intended indeed.

webpro:
Thanks @Desuuuu! Expecting this to be released shortly.

webpro:

Issue #1154, opened 2025-06-24
🐛 Biome plugin fails for Biome 2.0


asomethings:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://codesandbox.io/p/devbox/fdc577

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

Knip tries to parse `biome.json` configuration file's `extends` key as array which causes an error due to  [monorepo support.](https://biomejs.dev/blog/biome-v2/#monorepo-support)

In reproduction example below error will thrown

```
Analyzing workspace (packages/server)…
file:///project/workspace/node_modules/.pnpm/knip@5.61.2_@types+node@20.19.1_typescript@5.7.2/node_modules/knip/dist/plugins/biome/index.js:8
    return extendsArray.map(specifier => toConfig('biome', specifier, { containingFilePath: options.configFilePath }));
                        ^

TypeError: extendsArray.map is not a function
    at resolveExtends (file:///project/workspace/node_modules/.pnpm/knip@5.61.2_@types+node@20.19.1_typescript@5.7.2/node_modules/knip/dist/plugins/biome/index.js:8:25)
    at Object.resolveConfig (file:///project/workspace/node_modules/.pnpm/knip@5.61.2_@types+node@20.19.1_typescript@5.7.2/node_modules/knip/dist/plugins/biome/index.js:11:16)
    at runPlugin (file:///project/workspace/node_modules/.pnpm/knip@5.61.2_@types+node@20.19.1_typescript@5.7.2/node_modules/knip/dist/WorkspaceWorker.js:282:53)
    at async WorkspaceWorker.runPlugins (file:///project/workspace/node_modules/.pnpm/knip@5.61.2_@types+node@20.19.1_typescript@5.7.2/node_modules/knip/dist/WorkspaceWorker.js:325:13)
    at async build (file:///project/workspace/node_modules/.pnpm/knip@5.61.2_@types+node@20.19.1_typescript@5.7.2/node_modules/knip/dist/graph/build.js:93:35)
[...]

PR #1177, opened 2025-07-14
Support new biome config extends definition 

Thomascogez:
Since biome 2.0.0, they introduce support for workspace support and a new format for extends that can now be either an array of config to extends (same behavior than previous major version) or a reference to the workspace with a new "micro-syntax" (explained here https://biomejs.dev/guides/big-projects/#monorepo) (eg: `{ "extends: "//" }` will mean extends root workspace config)

## Summary of changes
- Updated the biome plugin `resolveConfig` in order to handle new extends format (either single or array)
- Resolve `"extends": "//"` -> `"extends": "[workspaceroot]/biome.(json|jsonc}"`

I'm still exploring project, I have added a new basic fixtutre `biome-workspace` in order to test this behavior, I would love to have some guidance here in order to add (if needed) more relevant test cases.

And i'm not really sure where and how should i resolve project root reference, since it's really tight to biome, I tried to keep things inside biome plugin file

### DIFF
--- a/packages/knip/src/plugins/biome/index.ts
+++ b/packages/knip/src/plugins/biome/index.ts
@@ -1,9 +1,8 @@
 import type { IsPluginEnabled, Plugin, PluginOptions, ResolveConfig } from '../../types/config.js';
 import { arrayify } from '../../util/array.js';
 import { type Input, toConfig } from '../../util/input.js';
-import { extname, join } from '../../util/path.js';
+import { join } from '../../util/path.js';
 import { hasDependency } from '../../util/plugin.js';
-import { _createSyncResolver } from '../../util/resolve.js';
 import type { BiomeConfig } from './types.js';
 
 const title = 'Biome';
@@ -13,23 +12,15 @@ const enablers = ['@biomejs/biome'];
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
 const config: string[] = ['biome.json', 'biome.jsonc'];
-const configExtensions = config.map(configFileName => extname(configFileName));
-
-const resolveRootConfigReference = (specifier: string, baseDir: string) => {
-  const resolver = _createSyncResolver(configExtensions);
-  const resolvedPath = resolver(specifier, baseDir);
-
-  return resolvedPath;
-};
 
 const isRootConfigReference = (specifier: string) => specifier === '//';
 
 const resolveExtends = (extendsArray: string[], options: PluginOptions): Input[] => {
   return extendsArray.map(specifier => {
     if (isRootConfigReference(specifier)) {
-      const resolvedSpecifier = resolveRootConfigReference(join(options.rootCwd, 'biome'), options.rootCwd);
-      return toConfig('biome', resolvedSpecifier ?? specifier, { containingFilePath: options.configFilePath });
+      return toConfig('biome', join(options.rootCwd, 'biome'), { containingFilePath: options.configFilePath });
     }
+
     return toConfig('biome', specifier, { containingFilePath: options.configFilePath });
   });
 };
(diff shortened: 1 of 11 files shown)


Thomascogez:
## Add support for biome(2.x) new `config.extends` format
> Related issue https://github.com/webpro-nl/knip/issues/1154

Since biome 2.0.0, they introduce support for workspace support and a new format for extends that can now be either an array of config to extends (same behavior than previous major version) or a reference to the workspace with a new "micro-syntax" (explained here https://biomejs.dev/guides/big-projects/#monorepo) (eg: `{ "extends: "//" }` will mean extends root workspace config)

## Summary of changes
- Updated the biome plugin `resolveConfig` in order to handle new extends format (either single or array)
- Resolve `"extends": "//"` -> `"extends": "[workspaceroot]/biome.(json|jsonc}"`

I'm still exploring project, I have added a new basic fixtutre `biome-workspace` in order to test this behavior, I would love to have some guidance here in order to add (if needed) more relevant test cases

alexprudhomme (approved):

packages/knip/src/plugins/biome/index.ts
```diff
@@ -10,13 +13,29 @@ const enablers = ['@biomejs/biome'];
...
+  return resolvedPath;
+};
+
+const isRootConfigReference = (specifier: string) => specifier === '//';
 
 const resolveExtends = (extendsArray: string[], options: PluginOptions): Input[] => {
-  return extendsArray.map(specifier => toConfig('biome', specifier, { containingFilePath: options.configFilePath }));
+  return extendsArray.map(specifier => {
+    if (isRootConfigReference(specifier)) {
+      const resolvedSpecifier = resolveRootConfigReference(join(options.rootCwd, 'biome'), options.rootCwd);
```
webpro: In general, plugins shouldn't need to do much resolving. Would something like this work:

```ts
toConfig("biome", { dir: options.rootCwd })
```

Hoping it gets resolved here: https://github.com/webpro-nl/knip/blob/fix/eslint-flat-eslint-import-resolver-typescript/packages/knip/src/util/get-referenced-inputs.ts#L98

Thomascogez: Hey ! Thanks for your review

Yeah that was my first taught thanks for pointing this,

indeed I updated the code to the following:
```ts
toConfig('biome', join(options.rootCwd, 'biome'), { containingFilePath: options.configFilePath });
```
And the output of https://github.com/webpro-nl/knip/blob/535ecd691b8309d5d1bae4a6e4465d79b1fcb6e7/packages/knip/src/util/get-referenced-inputs.ts#L98 Look like this
```txt
"/Users/thomascogez--allix/Developer/knip/packages/knip/fixtures/plugins/biome-workspace/biome.jsonc"
```

Which seem correct ! 

I also updated the default extensions for `createSyncResolver` and added `.jsonc`

Thomascogez: I also ignored this path `fixtures/plugins/biome-workspace/packages/stub` in the biome config since it contain a biome config that mimic 2.x.x format and the current project version is 1.9.x

packages/knip/src/plugins/biome/index.ts
```diff
@@ -11,12 +13,20 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
...
+      return toConfig('biome', join(options.rootCwd, 'biome'), { containingFilePath: options.configFilePath });
+    }
+
+    return toConfig('biome', specifier, { containingFilePath: options.configFilePath });
+  });
 };
 
 const resolveConfig: ResolveConfig<BiomeConfig> = (config, options) => {
-  return [...resolveExtends(config.extends || [], options)];
+  return [...resolveExtends(arrayify(config.extends || []), options)];
```
webpro: Nit: `arrayify` returns an empty array if value is not a string or array

packages/knip/src/plugins/biome/index.ts
webpro: Thanks, all good. Yes we should ignore the stub config, even if Knip itself goes Biome 2.

webpro (approved):
Just one nit left. Thanks!

webpro:
Let's go 🚀

Thomascogez:
Thanks for your time ! And thanks for Knip !

webpro:

PR #1180, opened 2025-07-16
Correctly detect used/unused namespace exports that begin with double underscores

akallem:
**Bug**
knip fails to detect that an identifier beginning with two underscores is used when accessed as a namespace property.

**Example**
```
// exports.js
export const __foo = 1;

// imports.js
import * as NS from "./exports.js";

console.log(NS.__foo);
```

**Explanation**
In the above example, the export of `__foo` is processed [here](https://github.com/webpro-nl/knip/blob/cbfb9c3b1c6fd4f77b5e5f0987e0f027ee609feb/packages/knip/src/typescript/visitors/exports/exportKeyword.ts). Importantly, the identifier name is obtained using `getText()`.

However, the namespace property access `NS.__foo` is processed [here](https://github.com/webpro-nl/knip/blob/cbfb9c3b1c6fd4f77b5e5f0987e0f027ee609feb/packages/knip/src/typescript/ast-helpers.ts#L184) using `node.name.escapedText`. For identifiers that start with two or more underscores, `escapedText` intentionally adds an _additional_ underscore (see discussion at https://github.com/microsoft/TypeScript/issues/46401). As a result, the export name is incorrectly not found in the analysis of used exports.

**Fix**
Change `ast-helpers.ts` to use `getText()` [here](https://github.com/webpro-nl/knip/compare/main...akallem:knip:fix/id-underscores?expand=1#diff-bedfe15e5c9800561eb8ac5d536ad02084cb1b5c9c1b72f2d8dc84acc6d65e2fR184).

**Tests**
Added a new test `ignore-exports-used-in-file-id-underscores.test.ts` to verify the fix. This new test fails if the proposed fix is reverted.

Also added an additional example and assertion to `exports-special-characters.test.ts`. This passes both before and after the proposed fix, nevertheless it seemed worthwhile to check.

### DIFF
--- a/packages/knip/fixtures/exports-special-characters/exports.ts
+++ b/packages/knip/fixtures/exports-special-characters/exports.ts
@@ -1,6 +1,7 @@
 export const $dollar = '$';
 export const dollar$ = '$$';
 export const _underscore = '_';
+export const __underscores = '__';
 
 export class DollarMembers {
   $member: string;
--- a/packages/knip/fixtures/ignore-exports-used-in-file-id-underscores/imported.ts
+++ b/packages/knip/fixtures/ignore-exports-used-in-file-id-underscores/imported.ts
@@ -1,0 +1,2 @@
+export const __underscoresUnused = 'underscoresUnused';
+export const __underscoresUsed = 'underscoresUsed';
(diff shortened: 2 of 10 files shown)


webpro:
Thanks Andrew, this is great. Much appreciated!

webpro:

Issue #609, opened 2024-04-29
`peerDependency` range conflicts with `typescript` beta releases


voxpelli:
Similar to how the ESLint 9 pre-releases caused problems with peer dependency ranges in ESLint plugins (see https://github.com/eslint-community/eslint-utils/pull/206) the peer dependency range for `typescript` in `knip` causes problem when using with `5.5.0-beta`

This is primarily a problem with how `>=` is handled the same in `peerDependencies` as in `dependencies` in npm, but I wanted to raise it as an issue here as well so that I can reference it from there: https://github.com/npm/rfcs/pull/397

fregante:
https://github.com/webpro-nl/knip/blob/cbfb9c3b1c6fd4f77b5e5f0987e0f027ee609feb/packages/knip/package.json#L78-L81

Not really a bug here, every npm package that uses `peerDependencies` is affected. The only fix is not using `peerDependencies`.

fregante:
![the only way to win is not to play](https://github.com/user-attachments/assets/fde7873c-9ad2-4cf6-b31c-2a4c2d157a95)

voxpelli:
As stated in the linked ESLint PR one can allow pre-releases by doing something like `^8.0.0 || >=9.0.0-0` and as mentioned in the npm RFC: This is something that npm should work on fixing.

> The only fix is not using `peerDependencies`.

It’s not.

fregante:
> This is something that npm should work on fixing

This repo is not npm, hence my comment: "Not really a bug here"

I opened a PR with your suggested fix, easier than tracking this bug for 15 months.

webpro:
Let's close this one, also see https://github.com/webpro-nl/knip/pull/1181#issuecomment-3078769688

PR #1181, opened 2025-07-16
Allow TypeScript betas as a peer dependency of knip

fregante:
Fixes https://github.com/webpro-nl/knip/issues/609 according to @voxpelli. Seems like an easy enough fix, the problem is having to do this every 3 months

```diff
-    "typescript": ">=5.0.4 || >=5.9.0-0"
+    "typescript": ">=5.0.4 || >=6.0.0-0"
```
```diff
-    "typescript": ">=5.0.4 || >=6.0.0-0"
+    "typescript": ">=5.0.4 || >=6.1.0-0"
```

### DIFF
--- a/packages/knip/package.json
+++ b/packages/knip/package.json
@@ -77,7 +77,7 @@
   },
   "peerDependencies": {
     "@types/node": ">=18",
-    "typescript": ">=5.0.4"
+    "typescript": ">=5.0.4 || >=5.9.0-0"
   },
   "devDependencies": {
     "@jest/types": "^29.6.3",


voxpelli:
One could add the `>=6.0.0-0` and `>=6.1.0-0` already to cut it down to less often.

Thinking about it: One could probably make a scheduled GitHub Action that opens a PR whenever it needs to be extended 🤔 I could make one when I have vacation in a couple of weeks if there's interest

webpro:
Why should Knip support TypeScript beta?

Btw there's coverage in CI (for `typescript@next`) to catch issues early.

voxpelli:
> Why should Knip support TypeScript beta?

Because otherwise it blocks people from trying out the TypeScript pre-releases (or forcing them to use eg. `--legacy-peer-deps` or such), see: https://github.com/eslint-community/eslint-utils/issues/183#issuecomment-1987046649

So its not so much about supporting beta as in fixing the issue in npm (https://github.com/npm/rfcs/pull/397) where `>=5.0.4` does not mean:

> as far as we know we will work with all versions, future and current, as long as they are at least 5.0.4

but weirdly explicitly instead means:

> as far as we know we will work with all versions, future and current, as long as they are at least 5.0.4 and **are not a prerelease of any kind – we are explicitly incompatible with all prereleases**

webpro:
Sorry, but no. It's not really an issue with Knip and there seems to be a workaround. Or use a different package manager. I'd rather clean things up around here than introduce this added scope.

voxpelli:
/ start of rant

This is why we desperately need a vendor neutral specification of the semantics of semantic versioning ranges in general as well as one for package.json dependency fields in particular - as currently pretty much everyone are dismissing issues like these as someone else’s problem and there is no proper home to file issues 🤷‍♂️

/ end of rant

webpro:
Truly sorry if it's frustrating. Trying to be a good OSS citizen, but personally I just don't find much joy in trying to support these kinds of things, while the range/variety of concepts and issues in this particular project is already pretty broad.

PR #1188, opened 2025-07-25
[react-router] mark routes and entries as production entries

rossipedia:
The `react-router` plugin incorrectly omits route and entry files when using `--production`.

This should fix that up.

Also, I added a small check against a common case when omitting `entry.server.tsx`, which is that `@react-router/node` and `isbot` are still necessary to be present in `dependencies`, even though they're not referenced explicitly in project code.

So if we detect there's no server entry, then we add those packages to the ignore list if they're present in `package.json`

One additional change as well:

The previous escaping logic was adding a `\` prior to any `$` in the path. This worked fine when there was characters _after_ the `$`, but had problems with the bare `$.tsx` and `$/route.tsx` pattern accepted by `@react-router/fs-routes` for "catch-all" routes (commonly used for 404 handling).

As best I can tell, this is a problem micromatch, but playing with this it seems as though escaping that is actually unnecessary, as removing the `$` in the RegExp allowed all tests to pass (I added both a `routes/$.tsx` and a `routes/$/route.tsx` file to the `react-router` fixture to veriy and updated the expected counters in the test).


rossipedia:
I don't think these test failures are related to my changes, but not 100% sure 🤔

packages/knip/src/plugins/react-router/index.ts
```diff
@@ -49,14 +50,36 @@ const resolveConfig: ResolveConfig<PluginConfig> = async (localConfig, options)
...
+    toProductionEntry(join(appDir, 'entry.{client,server}.{js,jsx,ts,tsx}')),
+    ...routes.map(id => toProductionEntry(id)),
+  ];
+
+  // When using @react-router/serve,  @react-router/node and isbot will also
+  // need to be installed. Since the server entry is optional, knip will flag
+  // these dependencies as unused.
+  // So add them to as ignored if
+  //  - there's no server entry and
+  //  - they're found in the manifest.
```
webpro: Shouldn't `@react-router/node` and `isbot` be peer deps of `@react-router/serve`?

Either way, following the logic, how about: if `hasServerEntry` then add `toProductionDependency('@react-router/node')` and `toProductionDependency('isbot')` to `resolved`?

packages/knip/src/plugins/react-router/index.ts
```diff
@@ -49,14 +50,36 @@ const resolveConfig: ResolveConfig<PluginConfig> = async (localConfig, options)
...
+  const hasServerEntry = fg.sync(join(appDir, 'entry.server.{js,ts,jsx,tsx}')).length !== 0;
```
webpro: Please import `_glob` for globbing

packages/knip/src/plugins/react-router/index.ts
rossipedia: > Shouldn't `@react-router/node` and `isbot` be peer deps of `@react-router/serve`?

So `@react-router/node` is a `dependency` of `@react-router/serve`, and when using React Router without explicit `entry.*.tsx` files then the project itself doesn't need to have it installed.

`isbot` probably needs to be added to `@react-router/serve` as an `optionalDependency`, since it's really only an implicit dependency if you're using the default entry files, which wouldn't be processed by `knip`.

I'll talk to the RR team about getting that added in 👍 

> Either way, following the logic, how about: if `hasServerEntry` then add `toProductionDependency('@react-router/node')` and `toProductionDependency('isbot')` to `resolved`?

SGTM :+1:

webpro:
> I don't think these test failures are related to my changes, but not 100% sure 🤔

Just pushed something to `main` - if you rebase we should be good (except for that sentry one in "Run Knip against external projects ").

packages/knip/src/plugins/react-router/index.ts
rossipedia: Actually, I'm not sure that's going to work. The implicit dependency only shows up when `entry.server.tsx` is _not_ present 🤔. When it is, then we don't need to add any config because `knip` can just walk the entry file that exists.

(Thinking my way through this here...)

Basically, when you don't have a `entry.server.tsx` file, `@react-router/dev` will include [its default entry file](https://github.com/remix-run/react-router/blob/65f431d6b90d1939e44dde6059a8e3b584a4d6a7/packages/react-router-dev/config/defaults/entry.server.node.tsx) as a part of _your_ build, which means your project needs to include the dependencies it needs (in this case: `@react-router/node` and `isbot`).

Implementation code is [here](https://github.com/remix-run/react-router/blob/65f431d6b90d1939e44dde6059a8e3b584a4d6a7/packages/react-router-dev/config/config.ts#L847-L895), for reference.

So I think maybe:

```ts
if (!hasServerEntry) {
  resolved.push(
    toProductionDependency('@react-router/node'), 
    toProductionDependency('isbot'),
  );
}
```

Should do it?

packages/knip/test/plugins/react-router.test.ts
```diff
@@ -25,3 +24,20 @@ test('Find dependencies with the react-router plugin', async () => {
...
+    cwd,
+  });
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    processed: 11,
+    total: 11,
+    // There is a bug with routes that include () on Windows so they will not be found there, revert when
+    // the bug is fixed
+    files: isWindows ? 1 : 0,
```
webpro: Let's keep the new fixture folder clean so we can focus on the issue at hand. Then we don't need to specify any of this Windows-related trouble here.

webpro (commented):
Thanks! Your explanation makes sense and the tests pass so I think we're good to go. Only thing left for me would be to clean up the fixtures folder a bit (no use in duplicating all files from the other react-router fixtures folder).

packages/knip/test/plugins/react-router.test.ts
```diff
@@ -25,3 +24,17 @@ test('Find dependencies with the react-router plugin', async () => {
     files: isWindows ? 1 : 0,
   });
 });
+
+test('Find dependencies with the react-router plugin [with entries]', async () => {
```
rossipedia:
```suggestion
test('Find dependencies with the react-router plugin [with custom server entry]', async () => {
```

webpro:
Thank you 🙏

webpro:

PR #1205, opened 2025-08-07
feat!: upgrade Zod to v4 for compatibility and improved performance

MagicKriss:
BREAKING CHANGE: Upgrade Zod from v3 to v4. While the knip configuration API remains the same, this major dependency upgrade may affect projects that directly depend on Zod types exported by knip or have conflicting Zod versions.

- Update zod from ^3.22.4 to ^4.0.15
- Update zod-validation-error from ^3.0.3 to ^4.0.0
- Fix function schema syntax for Zod v4 compatibility
- Update ignoreExportsUsedInFile schema for proper optional object handling


MagicKriss:
It's not just the performance - I've encountered that when switching to zod v4 on my project, the `knip` breaks because it's using zod as a dependency under the hood and it is expecting zod v3.
So the immediate step was to disable `knip` while it's updated - hence the PR

packages/knip/src/schema/configuration.ts
```diff
@@ -3,8 +3,8 @@ import { globSchema, pluginsSchema } from './plugins.js';
 
 const pathsSchema = z.record(z.string(), z.array(z.string()));
 
-const syncCompilerSchema = z.union([z.function().args(z.string(), z.string()).returns(z.string()), z.literal(true)]);
-const asyncCompilerSchema = z.function().args(z.string(), z.string()).returns(z.promise(z.string()));
+const syncCompilerSchema = z.union([z.custom<(from: string, content: string) => string>((val) => typeof val === 'function'), z.literal(true)]);
```
MagicKriss, line 6: Zod v4 has a breaking change here - `z.functions()` no longer creates a schema, instead it is a function factory. It does not have `.args` nor `.returns`   chaining methods.

https://zod.dev/v4/changelog?id=zfunction

packages/knip/src/schema/configuration.ts
```diff
@@ -31,17 +31,14 @@ const rulesSchema = z.record(issueTypeSchema, z.enum(['error', 'warn', 'off']));
...
-      z.literal('class'),
-      z.literal('enum'),
-      z.literal('function'),
-      z.literal('interface'),
-      z.literal('member'),
-      z.literal('type'),
-    ]),
-    z.boolean()
-  ),
+  z.object({
```
MagicKriss: Zod v4 became more strict with how it treats `record` - missing keys are now treated as validation failures.
Moving this to object with optionals allows "sparse records" to be valid

packages/knip/src/schema/configuration.ts
```diff
@@ -31,17 +31,14 @@ const rulesSchema = z.record(issueTypeSchema, z.enum(['error', 'warn', 'off']));
...
-      z.literal('class'),
-      z.literal('enum'),
-      z.literal('function'),
-      z.literal('interface'),
-      z.literal('member'),
-      z.literal('type'),
-    ]),
-    z.boolean()
-  ),
+  z.strictObject({
```
MagicKriss, line 34: Based on 
https://zod.dev/v4/changelog?id=deprecates-strict-and-passthrough

packages/knip/src/schema/configuration.ts
```diff
@@ -386,14 +383,18 @@ const baseWorkspaceConfigurationSchema = z.object({
   includeEntryExports: z.boolean().optional(),
 });
 
-const workspaceConfigurationSchema = baseWorkspaceConfigurationSchema.merge(pluginsSchema.partial());
+const workspaceConfigurationSchema = z.object({
```
MagicKriss, line 386: Based on
https://zod.dev/v4/changelog?id=deprecates-merge

webpro:
Thanks Krisjanis!

Much appreciated. Any chance you could elaborate a bit more on the breaking changes due to the exported types? In particular, what sort of use case(s)/configurations would break?

MagicKriss:
> Thanks Krisjanis!
> 
> Much appreciated. Any chance you could elaborate a bit more on the breaking changes due to the exported types? In particular, what sort of use case(s)/configurations would break?

Thank you for your considerations @webpro 
Investigating this further, this is the setup that's breaking:
- we have a project with different libs that are using different versiosn zod
- to have all of the libs compatible `resolutions` for zod  version is set in `package.json`
- with the existing resolution of zod the yarn selective dependency resolution did not work for us and versions are clashing

However, since then we've updated this and specified knip/zod resolution to be v3.

MagicKriss:
The PR still should hold if you want to upgrade zod version for the performance benefits and updating the deps.
But it was the package resolution that was giving the headaches here.
So I am ok also with closing this PR, if you don't want to get zod v4

webpro:
> > Much appreciated. Any chance you could elaborate a bit more on the breaking changes due to the exported types? In particular, what sort of use case(s)/configurations would break?
> 
> Thank you for your considerations @webpro Investigating this further, this is the setup that's breaking:
> 
>     * we have a project with different libs that are using different versiosn zod
> 
>     * to have all of the libs compatible `resolutions` for zod  version is set in `package.json`
> 
>     * with the existing resolution of zod the yarn selective dependency resolution did not work for us and versions are clashing
> 
> 
> However, since then we've updated this and specified knip/zod resolution to be v3.

Issues at the dependency level is what you've been looking into here. I meant this more at the type level: what could "break" for consumers of public Knip types such as `KnipConfig`? If there's breakage I'm not sure I'm that interested at this point. But if it's low impact then it wouldn't hurt to upgrade and have a smaller/faster Zod?

webpro:
Closing this one as fixed in [v5.64.1](https://github.com/webpro-nl/knip/releases/tag/5.64.1)

Issue #1193, opened 2025-07-29
🐛 Knip cannot resolve absolute reporter and preprocessor paths


scandar:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/scandar/repro-knip-path

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

I'm trying to pass an absolute path to knip for the `--preprocessor` argument but it doesn't properly resolve it since it treats it as a relative path.

For example this 
```bash
knip --preprocessor /Users/girgesscandar/code/repro-knip-path/src/B/utils/preproc.ts
```

Produces this error
```
Error loading /Users/girgesscandar/code/repro-knip-path/Users/girgesscandar/code/repro-knip-path/src/B/utils/preproc.ts
```

As you can see, the absolute path is appended to another absolute path.

webpro:
Ah yeah. Any chance you'd be up for a pull request? No worries if not!

scandar:
> Ah yeah. Any chance you'd be up for a pull request? No worries if not!

I'll try to whenever I have time, but it's up for grabs if someone else wants to volunteer.

PR #1216, opened 2025-08-12
fix: resolve reporter and preprocessor absolute paths properly

scandar:
Handles `reporter` and `preprocessor` absolute paths properly. Previously, absolute paths were treated as relative. 

Fixes: https://github.com/webpro-nl/knip/issues/1193


packages/knip/src/util/path.ts
```diff
@@ -19,7 +19,7 @@ export const toPosix = (value: string) => value.split(path.sep).join(path.posix.
 export const cwd = directory ? path.posix.resolve(directory) : toPosix(process.cwd());
 
 export const resolve = (...paths: string[]) =>
-  paths.length === 1 ? path.posix.join(cwd, paths[0]) : path.posix.resolve(...paths);
+  paths.length === 1 && !isAbsolute(paths[0]) ? path.posix.join(cwd, paths[0]) : path.posix.resolve(...paths);
```
webpro: Thanks for the fix! Yet I feel this is a bit "deep" into the core at this point. It's a good idea, but patching this here means we should probably also check a bunch of other places where we might not need to use `isAbsolute` anymore.

Is there any chance you could create a patch closer to the issue we're trying to fix here?

scandar: The issue comes from `resolve` ignoring absolute paths, the only other solution I can think of, which I'm not a fan of is to get the wrongly formed path from `path` and then trim the extra bit, but that is very hacky. 

I think the fix should be in `path` because the correct behaviour should not treat absolute paths as relative.

I'm not familiar with the test coverage in the repo. Would such a change create a lot of problems? Can tests catch most of the problems?

webpro: Was more thinking of a local fix such as `isAbsolute(proc) ? proc : resolve(proc)` - could that work maybe?

scandar: Sure, that would work, but that will need to be applied individually to all accepted `path` args like `reporter`, `preprocessor`, etc...

webpro: Well yeah just those two I think? Like I said it's a good idea, but a fix covering the whole codebase is a bit out of scope for this PR I think.

webpro:
Thanks Girges! 🙏

webpro:

PR #1218, opened 2025-08-12
feat(lsp): Add LSP to knip

clord:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,14420 @@
+{
+  "name": "@knip/root",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@knip/root",
+      "version": "0.0.0",
+      "license": "ISC",
+      "workspaces": [
+        "packages/*"
+      ],
+      "devDependencies": {
+        "@biomejs/biome": "1.9.4",
+        "installed-check": "^9.3.0"
+      },
+      "engines": {
+        "node": ">=22.0.0"
+      }
+    },
+    "node_modules/@astro-community/astro-embed-youtube": {
+      "version": "0.5.6",
+      "resolved": "https://registry.npmjs.org/@astro-community/astro-embed-youtube/-/astro-embed-youtube-0.5.6.tgz",
+      "integrity": "sha512-/mRfCl/eTBUz0kmjD1psOy0qoDDBorVp0QumUacjFcIkBullYtbeFQ2ZGZ+3N/tA6cR/OIyzr2QA4dQXlY6USg==",
+      "license": "MIT",
+      "dependencies": {
+        "lite-youtube-embed": "^0.3.3"
+      },
+      "peerDependencies": {
+        "astro": "^2.0.0 || ^3.0.0-beta || ^4.0.0-beta || ^5.0.0-beta"
+      }
+    },
+    "node_modules/@astrojs/check": {
+      "version": "0.9.4",
+      "resolved": "https://registry.npmjs.org/@astrojs/check/-/check-0.9.4.tgz",
+      "integrity": "sha512-IOheHwCtpUfvogHHsvu0AbeRZEnjJg3MopdLddkJE70mULItS/Vh37BHcI00mcOJcH1vhD3odbpvWokpxam7xA==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@astrojs/language-server": "^2.15.0",
+        "chokidar": "^4.0.1",
+        "kleur": "^4.1.5",
+        "yargs": "^17.7.2"
+      },
+      "bin": {
+        "astro-check": "dist/bin.js"
+      },
+      "peerDependencies": {
+        "typescript": "^5.0.0"
+      }
+    },
+    "node_modules/@astrojs/compiler": {
+      "version": "2.12.2",
+      "resolved": "https://registry.npmjs.org/@astrojs/compiler/-/compiler-2.12.2.tgz",
+      "integrity": "sha512-w2zfvhjNCkNMmMMOn5b0J8+OmUaBL1o40ipMvqcG6NRpdC+lKxmTi48DT8Xw0SzJ3AfmeFLB45zXZXtmbsjcgw==",
+      "license": "MIT"
+    },
+    "node_modules/@astrojs/internal-helpers": {
+      "version": "0.7.1",
+      "resolved": "https://registry.npmjs.org/@astrojs/internal-helpers/-/internal-helpers-0.7.1.tgz",
+      "integrity": "sha512-7dwEVigz9vUWDw3nRwLQ/yH/xYovlUA0ZD86xoeKEBmkz9O6iELG1yri67PgAPW6VLL/xInA4t7H0CK6VmtkKQ==",
+      "license": "MIT"
+    },
+    "node_modules/@astrojs/language-server": {
+      "version": "2.15.4",
+      "resolved": "https://registry.npmjs.org/@astrojs/language-server/-/language-server-2.15.4.tgz",
+      "integrity": "sha512-JivzASqTPR2bao9BWsSc/woPHH7OGSGc9aMxXL4U6egVTqBycB3ZHdBJPuOCVtcGLrzdWTosAqVPz1BVoxE0+A==",
+      "dev": true,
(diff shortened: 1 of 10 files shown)


clord:
Most of the checks are working, I've noticed sometimes the member-related diagnostics (enum, class) don't show up but could just be my test harness (update: found the reason, new push will fix). I figured it's good enough to get started.

packages/docs/src/content/docs/features/language-server-protocol.md
```diff
@@ -0,0 +1,266 @@
...
+
+1. Install the [Generic LSP Client](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd) extension
+2. Add to your `.vscode/settings.json`:
+
+```json
+{
+  "clangd.path": "knip",
+  "clangd.arguments": ["--lsp"]
+}
+```
```
webpro, line 44: For me this part doesn't work at all.

packages/docs/src/content/docs/reference/cli.md
```diff
@@ -423,6 +423,29 @@ knip --preprocessor preprocessor-package
...
+
+The LSP server provides:
+- Real-time diagnostics as you code
+- Code actions for quick fixes
+- Automatic re-analysis when files change
+- Support for workspace folders
+
+See [Language Server Protocol][19] for detailed setup instructions.
+
+### `--preprocessor-options [json]`
```
webpro: dedupe please

packages/knip/docs/lsp-mode.md
```diff
@@ -0,0 +1,177 @@
...
+- Use production mode to skip dev dependencies
+
+## Contributing
+
+The LSP implementation is in `packages/knip/src/reporters/lsp.ts`. Key areas for contribution:
+
+- Additional code actions for automatic fixes
+- Performance optimizations for large projects
+- Enhanced file watching capabilities
+- Support for additional LSP features (hover, completion, etc.)
```
webpro: Feels like there's quite some duplicate content.

packages/knip/src/cli.ts
```diff
@@ -66,6 +67,12 @@ const isShowProgress = isNoProgress === false && process.stdout.isTTY && typeof
 const workspace = rawWorkspaceArg ? toPosix(rawWorkspaceArg).replace(/^\.\//, '').replace(/\/$/, '') : undefined;
 
 const run = async () => {
+  if (isLsp) {
+    // Start the LSP server mode
+    await import('./lsp/server.js');
+    // The LSP server will handle its own lifecycle
+    return;
+  }
```
webpro: Should probably go into its own package, and then figure out a way to run things from there (without publicly exposing internals).

packages/knip/test/lsp/server.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  // Ensure critical issues are errors
+  assert.equal(severityMap['unresolved'], 1);
+  
+  // Ensure dependency issues are warnings
+  assert.equal(severityMap['dependencies'], 2);
+  assert.equal(severityMap['devDependencies'], 2);
+  
+  // Ensure code quality issues are hints
+  assert.equal(severityMap['enumMembers'], 4);
+  assert.equal(severityMap['classMembers'], 4);
```
webpro, line 129: Are these placeholders? We're not testing anything really.

packages/knip/package.json
```diff
@@ -68,6 +68,8 @@
     "picomatch": "^4.0.1",
     "smol-toml": "^1.4.1",
     "strip-json-comments": "5.0.2",
+    "vscode-languageserver": "9.0.1",
+    "vscode-languageserver-textdocument": "1.0.12",
```
webpro: Please also move deps to separate package.

package-lock.json
webpro, line 1: Please remove this lockfile, this project uses Bun.

packages/knip/docs/lsp-mode.md
```diff
@@ -0,0 +1,177 @@
...
+
+# Or use the dedicated binary
+knip-lsp
+```
+
+### As a reporter
+
+```bash
+# Output LSP-formatted diagnostics
+knip --reporter lsp
```
webpro: There is no such reporter?

clord: yea, i started out trying to make it a reporter, but that imposes certain restrictions on how it runs (LSP reruns things on change). I thought i removed all refs, will clean this up

clord:
> Thanks for the PR! While I really like the idea, it's kinda hard to assess at this point, as it feels a bit unfinished, and some parts were generated too?

Yea, it's mostly complete. The 'server side' is mostly done, as far as I can tell with tests. The client side however still needs a bit of work. I think it depends on the client to show some of the diagnostics. I need more experience with the different editors, but it's hard to test all the combinations. But even with the 'only certain things work' this is very useful, so i was thinking we could iterate it to eventually support everything we can via lsp. 

Yes, I ran the changes through an llm to try to match the repo expectations. Can you tell it generated the docs? ;) 

> I think it mostly boils down to create an actual LSP rather than trying to hook into an existing on? I can imagine this is to get things started, but I'm not yet experienced in this domain enough to wire things up myself in this PR.

I think Knip is ideal to be an LSP for this problem because it constructs a representation that can be run quickly and fed into the client. The 'blast damage' of adding an lsp is this one file. 

> Is there a way forward here? What are the steps to run this? I tried for a bit, to no avail. Btw I'm using VS Code myself.

Here is a Gist for getting neovim to use it locally. https://gist.github.com/clord/35b8608c028e588c82773921cd73042e

I don't use vscode myself (I'm motivated to get it for nvim and the Zed editor), but according to my quick search, it needs an 'extension' to use lsps. basically, a thin translation layer that speaks 'Knip' on one side and 'editor's primitives' on the other. These are very quick to write. I wrote one for zed (which i can't use until the `--lsp` option is on npm, as far as i can tell) there are 'generic extensions' for lsps but not sure if they work: https://github.com/mattn/efm-langserver

clord:
I'm happy to do some work to get it mergeable. This is very useful to have live in the editor, as you see it as a diagnostics just after you make something unused. There really isn't any other competition for knip if it has the ability to act as an lsp.

packages/docs/src/content/docs/features/language-server-protocol.md
clord: yea, i'll remove, it's generated anyway, but a real knip extension is probably required.

packages/knip/test/lsp/server.test.ts
clord: yea, i am not sure what the best way to test is. i think the right way would be to somehow seed a report (severityMap) with some diagnostics, then invoke the LSP and confirm it also reports the same issues, which is what these are trying to bootstrap. Open to suggestions on an alternative. Maybe there are already test cases with known problems i can just reuse?

packages/knip/src/cli.ts
clord: That's a really good idea. moving it over now. Looks like it should work without having to expose any internals.

clord: i'll just remove the `--lsp` option too, and have that package ship a `knip-lsp` binary.

clord:
Another thing i'd like to get help with is this function:

```ts
  private getIssueSeverity(issueType: string): DiagnosticSeverity {
    switch (issueType) {
      case "unresolved":
      case "unlisted":
      case "binaries":
      case "optionalPeerDependencies":
      case "enumMembers":
      case "classMembers":
        return DiagnosticSeverity.Warning;

      case "exports":
        return DiagnosticSeverity.Error;
        
      case "types":
      case "dependencies":
      case "devDependencies":
      case "nsExports":
      case "nsTypes":
      case "duplicates":
        return DiagnosticSeverity.Information;

      default:
        return DiagnosticSeverity.Information;
    }
  }
```

Right now, I don't have a good intuition about how serious these are. I'm tempted to just make them all warnings, because some editors just suppress Info/Hints. 

Edit: made them all DiagnosticSeverity.Warning for now.

webpro:
I'm going to reject this PR. But I'm grateful for the work and it def inspired me to start building one that's not vibe-coded ;) I'll post the branch here once it's up.

webpro:
New PR opened: https://github.com/webpro-nl/knip/pull/1370

Issue #641, opened 2024-05-16
Support local custom plugin


zhanzizhen:
Is there no other more convenient way except for submitting the code to the Knip repository to create a new plugin (which may be very time-consuming)?
It would be convenient if knip could support local plugins like:
```ts
// knip.config.js
{
  "my-plugin": {
      title:xxx,
      enablers: xxx,
      isEnabled: xxx,
      config:xxx,
      resolveConfig: xxx
  },
}
```

webpro:
Unfortunately, there is not. Sorry to hear building new plugins for Knip is time-consuming for you. Supporting this means it'll be more time-consuming for me, and I currently don't fancy this idea.

FelipePhilippCbh:
An use case: I have a few private libraries which are used across my work and would be great to have a way to add a local plugin to it. Currently the workaround is to either ignore files or add a `@public` comment

PR #1220, opened 2025-08-13
feat: prairielearn plugin

reteps:
Related: #641 

We need a plugin like the one described in this PR to pickup [our project](https://github.com/PrairieLearn/PrairieLearn)'s dependencies correctly. This plugin serves absolutely zero value unless you are developing within our custom framework. It doesn't make sense to me that for niche use-cases like this, I can't make a custom plugin -- it seems straightforward to export the types/utilities used for making a plugin, and put a disclaimer that the API is not stable. This also fixes the problem of "I want to use knip but my custom plugin is for an extremely small project, will it be added?"

I opened this PR as a discussion point for "how else would I do this in knip?", as I want to check these dependencies, and not just ignore them.


packages/knip/fixtures/plugins/prairielearn/question/info.json
```diff
@@ -0,0 +1,19 @@
+{
+    "controller": "pl-rich-text-editor.py",
```
reteps, line 2: PrairieLearn is a framework for creating custom autograded questions. Instructors use this custom format to specify the scripts that should be loaded for their question. Consequently, we have to be extra careful removing dependencies that any instructors are relying on.

webpro:
Hi Peter! Thanks for bringing this up. I totally get your point and I'm sorry if it's frustrating. Please consider this: I enjoy working on Knip, but in the end it's really still just a hobby project I do in my spare time. Also see https://knip.dev/reference/faq#why-doesnt-knip-have

reteps:
I was able to workaround this issue with this setup in `knip.ts`.

```ts
import { readFile } from 'fs/promises';

import { glob } from 'glob';
import type { KnipConfig } from 'knip';

const usedExampleDependencies = [
  ...(await glob('{exampleCourse,testCourse}/**/info.json')),
  ...(await glob('apps/prairielearn/elements/**/info.json')),
];

const questionAndElementConfigFiles = await Promise.all(
  usedExampleDependencies.map(async (path) => {
    const content = await readFile(path, 'utf-8');
    return JSON.parse(content);
  }),
);

const questionAndElementDependencies = questionAndElementConfigFiles.flatMap((infoJson) => [
  ...(infoJson.dependencies?.nodeModulesStyles ?? []),
  ...(infoJson.dependencies?.nodeModulesScripts ?? []),
  ...Object.values(infoJson.dynamicDependencies?.nodeModulesScripts ?? {}),
]);


const packageDependencies = new Set<string>(
  questionAndElementDependencies.map((dep) => {
    const parts = dep.split('/');
    if (parts[0].startsWith('@')) {
      return parts.slice(0, 2).join('/');
    }
    return parts[0];
  }),
);

const config: KnipConfig = {
  // ...
  ignoreDependencies: packageDependencies,
};

export default config;
```

webpro:
Guess we can close this one then.

PR #1231, opened 2025-08-22
feat: GitHub actions reporter

### LINKED ISSUE #1012: 💡 Add `github-actions` reporter!
### Suggest an idea for Knip

By printing out some formatted keywords, GitHub can understand the file and highlight it if knip is run inside CI!

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message

cylewaitforit:
closes #1012 

Builds of the work from @arthurfiorette and @ezrabrooks to add a github actions reporter to knip. 

Replaces `@actions/core` dependency with a simple local logger.

An example repo using the reporter can be found at https://github.com/cylewaitforit/knip-test/pull/1/files.


arthurfiorette:
I think you should enable this reporter to knip's own test suite

packages/knip/src/reporters/githubActions.ts
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  return {
+    info: (message: string) => console.log(message),
+    error: (message: string, options: { file: string; startLine?: number; endLine?: number; startColumn?: number; endColumn?: number }) => formatAnnotation('error', message, options),
+    warning: (message: string, options: { file: string; startLine?: number; endLine?: number; startColumn?: number; endColumn?: number }) => formatAnnotation('warning', message, options),
+  };
+};
+
+const core = createGitHubActionsLogger();
```
webpro: A bit of a nitpick I guess, but we need this only if the reporter's actually being used, so I think this could go into the reporter function.

packages/knip/test/cli/cli-reporter-githubActions.test.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const cwd = resolve('fixtures/module-resolution-non-std');
+
+test('knip --reporter githubActions (files, unlisted & unresolved)', () => {
+  assert.equal(
+    exec('knip --reporter githubActions', { cwd }).stdout,
+    `${cwd}/src/unused.ts
+::error file=${cwd}/src/index.ts::Unlisted dependencies
+::error file=${cwd}/src/index.ts::Unlisted dependencies
+::error file=${cwd}/src/index.ts,line=8,endLine=8,col=23,endColumn=23::Unresolved imports`
```
webpro: Would be great if there's coverage for `warn(ing)` too.

On a related not, would perhaps be good to find and use a fixture that has unused export and dependency for more diverse output (e.g. `fixtures/rules` or `fixtures/workspaces` or `fixtures/plugins/nuxt`, or multiple separate ones)

Btw I should add `pos` info to unlisted dependencies (WIP!)

EzraBrooks (commented):
Thanks so much for picking this up, I have been totally swamped with other work!

I agree with @arthurfiorette that a good application for this would be Knip's own CI here in GitHub Actions!

packages/knip/test/cli/cli-reporter-githubActions.test.ts
cylewaitforit: Added tests against the suggested fixtures.

packages/knip/src/reporters/githubActions.ts
cylewaitforit: Totally fair, moved this now.

cylewaitforit:
> Thanks so much for picking this up, I have been totally swamped with other work!
> 
> I agree with @arthurfiorette that a good application for this would be Knip's own CI here in GitHub Actions!
@EzraBrooks 


@webpro Thoughts on this? Would you like the Knip repo to use this reporter in GHA?

webpro:
Dogfooding might be a good idea! We can add the `--reporter actions` or `--reporter github-actions` somewhere in a workflow. That said, lets not use camelCase on CLI args :)

Any chance we can see it in action somehow somewhere?

packages/knip/test/cli/cli-reporter-githubActions.test.ts
webpro: Thanks!

FYI: 7959a11096b6867a9f09f1f958691f8e3733329a

cylewaitforit:
Here's an example repo using the github actions reporter 
https://github.com/cylewaitforit/knip-test/pull/1/files

packages/knip/test/cli/cli-reporter-githubActions.test.ts
cylewaitforit: This GithubActions reporter should work in a way where when position info is added it will start using it without it needing to be updated. In order for it to use annotations in the files section the annotation has to have some sort of position info so by default I set it to just fall back to the beginning of the file. Without setting position info it will only show the annotation in the summary and not in the files tab.

webpro:
Thanks a bunch, Cyle! Happy to merge. I've tested things and it LGTM. Any chance @EzraBrooks and @arthurfiorette can confirm this matches their expectations in their project(s) as well?

EzraBrooks:
This is a great first cut! It would be awesome to later make it annotate the package.json entries for unused dependencies, but I don't think that's necessary for the MVP.

cylewaitforit:
> This is a great first cut! It would be awesome to later make it annotate the package.json entries for unused dependencies, but I don't think that's necessary for the MVP.
@EzraBrooks 

Great news is that unused dependencies was already [accounted and tested for](https://github.com/webpro-nl/knip/blob/494bb5d4170aeb38bfe7e796850c2253f70ed344/packages/knip/test/cli/cli-reporter-githubActions.test.ts#L13). 

I also now added [an used dependency](https://github.com/cylewaitforit/knip-test/pull/1/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) to my knip playground test repo now as well if you are interested in seeing that "live in living color".

This actions reporter should also be fairly future tolerant. So that when the work to add [pos info in unlisted deps lands](https://github.com/webpro-nl/knip/pull/1231#discussion_r2321085619) it would work without needing changes to this reporter.

webpro:
What more can we ask for, Cyle? 🙏 Great job.

webpro:

PR #1254, opened 2025-09-09
Add `env-cmd` Support

joealden:
[`env-cmd`](https://github.com/toddbluhm/env-cmd) just had a v11 release that enforces `--` usage like `dotenvx` - this PR adds support for it.

If you're wondering why someone would use `env-cmd` over `dotenvx` - `env-cmd` has different behavior (doesn't do a load of logging), different defaults (overriding existing env vars), and supports more file formats (JSON etc.) - which is crucial when you don't have control over the format (for example, Google Cloud Functions require JSON dotenv files).


webpro:
Thank you, Joe 🙏

webpro:

Issue #1262, opened 2025-09-14
🐛 `filePath.includes is not a function` error with Nuxt module functions


wattanx:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://codesandbox.io/p/devbox/boring-poitras-f67j54?workspaceId=ws_AX1NvGrjCYpj1ZXVaSMqdq

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

Nuxt Modules can be configured not only with strings but also with functions.
However, when running `knip` with a function configured, the following error occurs:

```bash
export const isInNodeModules = (filePath) => filePath.includes('node_modules');
                                                      ^

TypeError: filePath.includes is not a function
    at isInNodeModules (file:///project/workspace/node_modules/.pnpm/knip@5.63.1_@types+node@24.0.13_typescript@5.8.3/node_modules/knip/dist/util/path.js:10:55)
    at getPackageNameFromSpecifier (file:///project/workspace/node_modules/.pnpm/knip@5.63.1_@types+node@24.0.13_typescript@5.8.3/node_modules/knip/dist/util/modules.js:19:59)
```

PR #1263, opened 2025-09-14
fix: handle only string modules in dependency resolution

wattanx:
Fixes: https://github.com/webpro-nl/knip/issues/1262

## Description
Since Nuxt modules can be configured not only as strings but also as functions or arrays, we should use toDependency only when the value is a string.
https://nuxt.com/docs/4.x/guide/concepts/modules#add-nuxt-modules

### DIFF
--- a/packages/knip/src/plugins/nuxt/index.ts
+++ b/packages/knip/src/plugins/nuxt/index.ts
@@ -48,6 +48,7 @@ const resolveConfig: ResolveConfig<NuxtConfig> = async localConfig => {
 
   const deps =
     localConfig.modules?.reduce<Input[]>((acc, id) => {
+      if (Array.isArray(id) && typeof id[0] === 'string') acc.push(toDependency(id[0]));
       if (typeof id === 'string') acc.push(toDependency(id));
       return acc;
     }, []) ?? [];


packages/knip/src/plugins/nuxt/index.ts
```diff
@@ -45,7 +46,11 @@ const resolveConfig: ResolveConfig<NuxtConfig> = async localConfig => {
     'server/plugins/**/*.ts',
   ].map(pattern => toProductionEntry(join(srcDir, pattern)));
 
-  const deps = localConfig.modules?.map(id => toDependency(id)) ?? [];
+  const deps =
+    localConfig.modules?.reduce<Input[]>((acc, id) => {
+      if (typeof id === 'string') acc.push(toDependency(id));
```
webpro, line 52: Looks like we could also add the first item of `id` in case it's an array?

wattanx: Thank you for the review!
I tried modifying it so that, in the case of arrays, it also adds the first index.
https://github.com/webpro-nl/knip/pull/1263/commits/e2fc5a19465499d1b6bd8c25077cdc8b99fe1e51

webpro:

Issue #1255, opened 2025-09-10
💡 Nuxt 4 Support


OrbisK:
### Suggest an idea for Knip

It looks like nuxt 4 paths are not yet supported. 

@danielroe I wonder if we could integrate with Knip more effectively by creating a nuxt module/knip plugin that generates the correct Knip configuration. 🤔

danielroe:
Indeed, I think we should have a Nuxt plugin, and this should be relatively straightforward.

webpro:
What are nuxt 4 paths?

If it is what I think it is, Knip plugins can return TS config style path aliases ([example](https://github.com/webpro-nl/knip/blob/21d6b5183ad10b3296cdb9c8f21a8f2d01bb36e9/packages/knip/src/plugins/svelte/index.ts#L24)), though this is still undocumented I see.

Knip has a Nuxt plugin ([source](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/nuxt/index.ts#L64)), it's just not very advanced.

OrbisK:
@webpro I experimented a bit.

I resolved the relevant paths with nuxt/kit: https://github.com/OrbisK/knip/blob/feat/nuxt/packages/knip/src/plugins/nuxt/index.ts (prototype)

However, this approach only resolves the static nuxt config.

OrbisK:
Do you think knip will support extending configs/patterns from a path in the near future?

This would allow us to write a Nuxt module to generate a `.nuxt/knip.json` file containing all entry points and dependencies within the nuxt context (on `nuxt prepare`). It would work in a similar way to `nuxt/eslint`s approach: https://eslint.nuxt.com/packages/module#manual-setup

```json5
// knip.json
{
  extends: [
    ".nuxt/knip.json"
  ]
}
```

This should eliminate false positive and false negative issues with `layers`, local `modules` and custom configuration patterns.

webpro:
Maybe we could get away by adding e.g. `.nuxt/knip.json` to the `config` array? Would require a "hybrid" `resolveConfig` so that it handles both files separately but in a single function (not pretty, but more plugins do this e.g. [example](https://github.com/webpro-nl/knip/blob/21d6b5183ad10b3296cdb9c8f21a8f2d01bb36e9/packages/knip/src/plugins/nx/index.ts#L44)). not sure if feasible in this case though.

Btw why would we need `@nuxt/kit`?

OrbisK:
[...]

PR #1265, opened 2025-09-16
feat(nuxt): use `nuxt` to resolve its own paths

danielroe:
resolves https://github.com/webpro-nl/knip/issues/1255

this rewrites the nuxt plugin to use nuxt itself to resolve its own paths, supporting layers, modules, etc.

#### 🚧 todo

at the moment test is failing.

we need to enable components + imports to register dependencies without being entries - ie. detect unused composables, but if a composable imports from something else, then that should count as that other thing being used

(this may be straightforward)

### DIFF
--- a/packages/knip/.release-it.json
+++ b/packages/knip/.release-it.json
@@ -5,8 +5,8 @@
     "before:init": [
       "pnpm lint",
       "pnpm build",
-      "pnpm knip",
-      "pnpm knip:production",
+      "NODE_PATH= pnpm knip",
+      "NODE_PATH= pnpm knip:production",
       "pnpm test"
     ]
   },
--- a/packages/knip/package.json
+++ b/packages/knip/package.json
@@ -1,6 +1,6 @@
 {
   "name": "knip",
-  "version": "5.85.0",
+  "version": "5.86.0",
   "description": "Find and fix unused dependencies, exports and files in your TypeScript and JavaScript projects",
   "keywords": [
     "analysis",
@@ -77,8 +77,8 @@
     }
   },
   "scripts": {
-    "knip": "node ./dist/cli.js --directory ../..",
-    "knip:production": "node ./dist/cli.js --directory ../.. --production --strict",
+    "knip": "node ./src/cli.ts --directory ../..",
+    "knip:production": "node ./src/cli.ts --directory ../.. --production --strict",
     "lint": "oxlint",
     "format": "oxfmt -c ../../.oxfmtrc.json --ignore-path ../../.prettierignore",
     "test": "node scripts/run-test.ts",
--- a/packages/knip/src/version.ts
+++ b/packages/knip/src/version.ts
@@ -1,1 +1,1 @@
-export const version = '5.85.0';
+export const version = '5.86.0';
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@knip/language-server",
-  "version": "2.0.1",
+  "version": "2.0.2",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/webpro-nl/knip.git",
--- a/packages/mcp-server/package.json
+++ b/packages/mcp-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@knip/mcp",
-  "version": "0.0.19",
+  "version": "0.0.20",
   "description": "Knip MCP Server",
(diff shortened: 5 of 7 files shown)


packages/knip/fixtures/plugins/nuxt/app.tsx
danielroe, line 1: this is a valid nuxt entry, believe it or not, and if present nuxt would actually use this

packages/knip/src/plugins/nuxt/index.ts
```diff
@@ -24,37 +25,140 @@ const production = [
...
+      availableSources.add(resolve(rootDir, config.dir.shared ?? 'shared', `types/*{${extensionGlob}}`));
+    }
+
+    inputs.push(...entryPatterns.map(s => toProductionEntry(s)));
+  }
+
+  // TODO: we need to register these as source files which may not be all used
+  for (const path of availableSources) {
+    inputs.push(toProductionEntry(path, { allowIncludeExports: true }));
+  }
```
danielroe: is there a way to register these paths so any imports _into_ them are respected, but unused exports are detected?

webpro: Yes, if I understand correctly, this is the breakdown:

- default behavior: exports of entry files are not reported
- `skipExportsAnalysis: false` - report unused exports (what you're asking for)
- `allowIncludeExports: true` - report unused exports only with flag `--include-entry-exports`/`includeEntryExports`

darthf1:
Would be awesome to see this finished :)

webpro:
What's the status here? Is there still interest in wrapping this up?

danielroe:
yes! apologies on the long delay.

webpro:
Took a different stab at this in #1517. If anyone could take it for a spin in their projects, that would be great :)

yschroe:
> Took a different stab at this in #1517. If anyone could take it for a spin in their projects, that would be great :)

Thanks! I tried your PR and for me it worked pretty well and I could already remove a bit of dead code from my project. 👍

A few things I noticed:
- [app.config.ts ](https://nuxt.com/docs/4.x/directory-structure/app/app-config) files are marked as unused (in my project this file is inside a nuxt layer folder)
- `index.ts` from an auto-imported module is marked as unused (for me also inside a nuxt-layer)
- Inside this `index.ts` I have registered a routeMiddleware via `addRouteMiddleware`. The middleware file is also marked as unused
- I have a `layers/auth/app/types/page-meta.d.ts` file to augment the definePageMeta type -> this file is also marked as unused
- Nuxt modules that are added only in the `modules` property in `defineNuxtConfig`/`nuxt.config.ts` are marked as unused dependencies (e.g. @nuxt/image)
- `vitest-environment-nuxt` is marked as a unlisted dependency if used in `vitest.config.ts` (via environment: 'nuxt' - see [nuxt test-utils](https://github.com/nuxt/test-utils). nuxt test-utils are listed in my devDependencies)

Not sure how much I normally would need to use a knip ignore list (if this exists) as I have never used knip outside of this one nuxt project.

webpro:
Thanks Yannik! This is great feedback. Any chance you could copy-paste it into #1517?

webpro:
Thanks for the PR, Daniel! It is much appreciated. Using `nuxt/kit` clearly has some advantages, including at least:

- Catches more entries more reliably.
- Being a peer dependency provides versioned updates alongside project.

Either approach needs to `prepare()` and requires most/all of what the plugin is currently doing. There are downsides as well:

- Additional overhead of running `loadNuxt()`.
- The "auto imports" feature adds complexity to Nuxt itself. Essentially, that's the reason we're adding a bunch of complexity to Knip and we should minimize that as good as we can.
- Might set a precedent for other frameworks.

By design, Knip has affordances to take what it needs from configuration files with input helpers like `toEntry`, `toConfig`, `toAlias` and there is `resolveFromAST` to the rescue as well. Tool or framework internals should not be necessary, to keep Knip as lean & mean as possible (tbh Knip isn't that fast to begin with). We currently probably cover >95% of cases in Knip already, and I'm sure we can still do better. Edge cases can usually be covered with a manual `entry` or `ignore*` item.

The vision, direction and resources of this project do not align with taking in more of such complexity, and I would prefer not to include the extra dev/peer dependencies. We could still refactor the pull request and skip the new dependencies. Keep the remaining improvements to help the Knip + Nuxt community minimize their required configuration. What do you think?

webpro:

PR #1275, opened 2025-09-21
fix: enable pnpm plugin on root config & lockfile

nyarthan:
Take into account `pnpm-lock.yaml` & `pnpm-workspace.yaml` for enabling the pnpm plugin.

### DIFF
--- a/packages/knip/src/plugins/pnpm/index.ts
+++ b/packages/knip/src/plugins/pnpm/index.ts
@@ -1,10 +1,15 @@
 import type { IsPluginEnabled, Plugin } from '../../types/config.js';
+import { _firstGlob } from '../../util/glob.js';
 
 // https://pnpm.io/pnpmfile
 
 const title = 'pnpm';
 
-const isEnabled: IsPluginEnabled = ({ manifest }) => Boolean(manifest.packageManager?.startsWith('pnpm@'));
+const isEnabled: IsPluginEnabled = async ({ cwd, manifest }) =>
+  Boolean(
+    manifest.packageManager?.startsWith('pnpm@') ||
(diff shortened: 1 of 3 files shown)


packages/knip/src/plugins/pnpm/index.ts
```diff
@@ -1,10 +1,15 @@
...
 
 // https://pnpm.io/pnpmfile
 
 const title = 'pnpm';
 
-const isEnabled: IsPluginEnabled = ({ manifest }) => Boolean(manifest.packageManager?.startsWith('pnpm@'));
+const isEnabled: IsPluginEnabled = async ({ cwd, manifest }) =>
+  Boolean(
+    manifest.packageManager?.startsWith('pnpm@') ||
+      (await _firstGlob({ cwd, patterns: ['pnpm-lock.yaml', 'pnpm-workspace.yaml'] }))
```
webpro, line 11: I think `pnpm-workspace.yaml` isn't needed here? Could it ever happen that there is a `pnpm-workspace.yaml`, and neither `package.json#pnpm` nor `pnpm-lock.yaml`?

webpro (commented):
Thanks Jannis, one comment

packages/knip/src/plugins/pnpm/index.ts
nyarthan: I suppose it's highly unlikely but there _is_ the [`--no-lockfile` option](https://pnpm.io/cli/install#--no-lockfile) and people might not commit the lockfile to version control

webpro: Alright then! I'm wary of changes here as this runs in each workspace. But I see this plugin does not have `isRootOnly` yet so that'll "compensate" (I'll add it).

webpro:

PR #1277, opened 2025-09-21
Add `Rstest` Plugin

nyarthan:
Adds plugin for [`Rstest`](https://rstest.rs/)

### DIFF
--- a/packages/knip/src/plugins/rstest/index.ts
+++ b/packages/knip/src/plugins/rstest/index.ts
@@ -1,5 +1,5 @@
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
-import { toDeferResolve } from '../../util/input.js';
+import { toDeferResolve, toEntry } from '../../util/input.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { RstestConfig } from './types.js';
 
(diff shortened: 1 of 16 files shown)


packages/knip/src/plugins/rstest/index.ts
```diff
@@ -0,0 +1,42 @@
...
+// https://rstest.rs/config/test/include
+const entry = ['**/*.{test,spec}.?(c|m)[jt]s?(x)'];
+
+const resolveConfig: ResolveConfig<RstestConfig> = async config => {
+  const entries = (config.include ?? entry)
+    .concat(...mocks)
+    .map(toEntry)
+    .concat(...(config.exclude ?? []).map(id => toEntry(`!${id}`)));
+
+  const environments = config.testEnvironment === 'node' ? [] : [config.testEnvironment ?? 'node'];
```
webpro: Should any of the valid `testEnvironment` actually resolve to a (peer) dependency?

Dependencies of rstest should not be added (but peer deps should).

packages/knip/test/plugins/rstest.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const cwd = resolve('fixtures/plugins/rstest');
+
+test('Find dependencies with the rstest plugin', async () => {
+  const options = await createOptions({ cwd });
+  const { counters } = await main(options);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    files: 1,
```
webpro, line 19: Looks like `not-included.ts`(?) is intentionally unused, so that's good. Might be good to have something like `assert(issues.files.has(join(cwd, file))` to double-check it's the one expected.

packages/knip/test/plugins/rstest2.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const cwd = resolve('fixtures/plugins/rstest2');
+
+test('Find dependencies with the rstest plugin (2)', async () => {
+  const options = await createOptions({ cwd });
+  const { counters } = await main(options);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    files: 2,
```
webpro, line 20: ^^ same here

packages/knip/src/plugins/rstest/index.ts
nyarthan: Yup, they are peers
https://github.com/web-infra-dev/rstest/blob/3c196415be578d463d69285a4efeb4305fc6fdf5/packages/core/package.json#L90

packages/knip/test/plugins/rstest.test.ts
nyarthan: Done!

packages/knip/src/plugins/rstest/index.ts
webpro: Right, thanks. Guess I got a bit confused by the `node` one - that fallback can be removed, right?

webpro: It should actually even be explicitly filtered out if `config.testEnvironment` has value `"node"`.

nyarthan: Oh yeah good catch - the fallback is indeed wrong, will fix

webpro:
Awesome, thanks Jannis!

webpro:

PR #1278, opened 2025-09-21
feat: add plugin for `bumpp`

nyarthan:
Adds plugin for [`bumpp`](https://github.com/antfu-collective/bumpp#bumpp)

### DIFF
--- a/packages/knip/schema.json
+++ b/packages/knip/schema.json
@@ -319,6 +319,10 @@
           "title": "biome plugin configuration (https://knip.dev/reference/plugins/biome)",
           "$ref": "#/definitions/plugin"
         },
+        "bumpp": {
+          "title": "bumpp plugin configuration (https://knip.dev/reference/plugins/bumpp)",
+          "$ref": "#/definitions/plugin"
+        },
         "bun": {
           "title": "bun plugin configuration (https://knip.dev/reference/plugins/bun)",
           "$ref": "#/definitions/plugin"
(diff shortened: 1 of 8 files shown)


packages/knip/schema.json
```diff
@@ -603,14 +607,14 @@
...
-          "$ref": "#/definitions/plugin"
-        },
         "rspack": {
           "title": "rspack plugin configuration (https://knip.dev/reference/plugins/rspack)",
           "$ref": "#/definitions/plugin"
         },
+        "rstest": {
+          "title": "rstest plugin configuration (https://knip.dev/reference/plugins/rstest)",
+          "$ref": "#/definitions/plugin"
+        },
```
nyarthan, line 617: Looks like I messed up the sorting last time...

packages/knip/src/plugins/bumpp/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+// https://github.com/antfu-collective/bumpp#bumpp
+
+const title = 'bumpp';
+
+const enablers = ['bumpp'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const entry = ['bump.config.{mjs,ts,js,cjs,mts,cts}'];
```
nyarthan, line 12: The docs only mention a `bump.config.ts` but I tested it with the other extensions and they also work, which makes sense since it's using the [`c12`](https://github.com/unjs/c12) config loader

packages/knip/schema.json
webpro: automation ftw!

packages/knip/src/plugins/bumpp/index.ts
webpro: Thanks for looking that up. You could also consider using the `toC12config` helper (leaving that up to you).

nyarthan: Yeah for sure, would do it in a follow up since it's likely used by a few plugins, if you don't mind.

webpro: For sure 👍

webpro:

PR #1285, opened 2025-10-03
feat: improve `lefthook` plugin

nyarthan:
- include all valid configuration file locations / extensions
- enable plugin if config exists (I think it's common enough to install lefthook via a different method than as a node module e.g. in a nix setup)


packages/knip/src/plugins/lefthook/index.ts
```diff
@@ -11,11 +12,22 @@ const title = 'Lefthook';
...
+const patterns = [
+  'lefthook.{yml,yaml,json,toml}',
+  '.lefthook.{yml,yaml,json,toml}',
+  'lefthook-local.{yml,yaml,json,toml}',
+  '.lefthook-local.{yml,yaml,json,toml}',
+  '.config/lefthook.{yml,yaml,json,toml}',
+];
+
+const isEnabled: IsPluginEnabled = async ({ dependencies, cwd }) =>
+  Boolean(hasDependency(dependencies, enablers) || (await _firstGlob({ cwd, patterns })));
```
webpro, line 24: The issue with enabling the plugin by config file existence is that we'll no longer be able to determine the opposite: how can we mark the dependency as unused in the future (as it is still often listed)?

nyarthan: That makes sense - maybe we need another way of "force enabling" a plugin for these cases?

Issue #1244, opened 2025-09-01
🧩 Is confused by _comment in .lintstagedrc.json


rfalke-rtl:
### Discuss anything related to Knip

This works

```
{
  "*.+(css|json|md|html|yaml|yml)": [
    "prettier --write"
  ],
  "*.+(js|jsx|ts|tsx)": [
    "eslint --fix",
    "prettier --write"
  ],
}
```

But this

```
{
  "*.+(css|json|md|html|yaml|yml)": [
    "prettier --write"
  ],
  "*.+(js|jsx|ts|tsx)": [
    "eslint --fix",
    "prettier --write"
  ],
  "_comment": "This file is generated by ...."
}
```

results in

```
....
Analyzing workspace…
Unlisted binaries (3)
This     .lintstagedrc.json
Note     .lintstagedrc.json
changes  .lintstagedrc.json
```

webpro:
Is this an issue for many folks because it's from some example/template?

PR #1291, opened 2025-10-07
Skip comments in .lintstagedrc.json files using a heuristic

rfalke-rtl:
This fixes https://github.com/webpro-nl/knip/issues/1244


packages/knip/src/plugins/lint-staged/index.ts
```diff
@@ -29,7 +29,10 @@ const resolveConfig: ResolveConfig<LintStagedConfig> = async (config, options) =
 
   const inputs = new Set<Input>();
 
-  for (const entry of Object.values(config).flat()) {
+  for (const [key, entry] of Object.entries(config)) {
+    // Skip non-glob keys (comments, metadata, etc.)
+    if (key.startsWith('_') || key.startsWith('$')) continue;
```
webpro: Is this lint-staged behavior documented or encoded somewhere?

rfalke-rtl: No. lint-staged is agnostic. If a pattern doesn't match anything it will not be applied. 

https://stackoverflow.com/a/244858 and https://www.freecodecamp.org/news/comments-in-json/ both suggest "_comment".

webpro: Thanks. The thing is, the `_` and `$` aren't invalid in a pattern. Could we do with only skipping `_`-prefixed keys?

rfalke-rtl: I changed the code.

webpro:
:rocket: _This pull request is included in v5.64.3. See [Release 5.64.3](https://github.com/webpro-nl/knip/releases/tag/5.64.3) for release notes._

packages/knip/src/plugins/lint-staged/index.ts
webpro: Thanks! I released-it 🚀

PR #1297, opened 2025-10-10
Real JSON5 Support

kiliantyler:
# Changes

- Add testing for JSON, JSONC, and JSON5
- Move JSON, JSONC, and JSON5 parsing to the JSON5 package.

# Reasoning

Current implementation uses `strip-json-comments` package to parse all forms of JSON. JSON5 has a superset of features that are not available in the JSON parser, even with comments and trailing commas removed.

Knip is not a JSON linting tool so it should generally "not care" if JSON isn't valid and actually has comments in it, it just needs to parse it successfully. This accomplishes that by parsing all JSON/C/5 files with the JSON5 parser.

Linting/Formatting the JSON files used for testing causes the only files to be tested to be ones that are sanitized, which is not what exists in projects, so I have excluded them from formatting.

### AI Disclaimer
AI was used to generate the JSON, JSONC, and JSON5 files for testing. I reviewed the files manually after creation and formatted them to seem like real world examples.

### DIFF
--- a/packages/knip/fixtures/plugins/lint-staged-comment/.lintstagedrc.json
+++ b/packages/knip/fixtures/plugins/lint-staged-comment/.lintstagedrc.json
@@ -1,10 +1,5 @@
 {
-  "*.+(css|json|md|html|yaml|yml)": [
-    "prettier --write"
-  ],
-  "*.+(js|jsx|ts|tsx)": [
-    "eslint --fix",
-    "prettier --write"
-  ],
+  "*.+(css|json|md|html|yaml|yml)": ["prettier --write"],
+  "*.+(js|jsx|ts|tsx)": ["eslint --fix", "prettier --write"],
   "_comment": "This file is generated. Note that while it is possible to alter this file, changes should be made carefully."
 }


webpro:
Thanks for bringing this up! We need to consider this indeed.

Just to check up on the status quo first, we're already having this in the codebase:

- https://github.com/webpro-nl/knip/blob/ed2acecbdbcf3eece05c4e5777ac5bb4f3620e06/packages/knip/src/plugins/prettier/index.ts#L17
- https://github.com/webpro-nl/knip/blob/ed2acecbdbcf3eece05c4e5777ac5bb4f3620e06/packages/knip/src/util/plugin-config.ts#L55

That makes 4 plugins (prettier, bumpp, changelogen, changelogithub) at this point that "support json5". There's no way back I guess. But how many projects are actually _using_ json5 that throws in the current solution?

The json5 dependency is large and slow. Don't fancy including both dependencies either.

Looking at https://github.com/unjs/confbox which looks convenient with basically the same dependencies, but has the overhead of storing the format and holding the object in memory.

webpro:
The project also looks pretty much dead btw

```
❯ last-publish json5
The last publish date of json5 is 12/31/2022
```

kiliantyler:
> That makes 4 plugins (prettier, bumpp, changelogen, changelogithub) at this point that "support json5". There's no way back I guess. But how many projects are actually _using_ json5 that throws in the current solution?

I was! But that's why I'm here trying to fix it, I actually downgraded my prettier.json5 to be just a standard JSON so I could use knip ;)

IMO, removing JSON5 support would be better than supporting the extension but not the actual spec of it. 

> The project also looks pretty much dead btw

I considered that, the JSON5 spec hasn't changed so I don't think the package would *need* to be updated, but I can see the issue with a project that isn't actively maintained.

> Looking at https://github.com/unjs/confbox which looks convenient with basically the same dependencies

That is interesting, to keep dependencies to a minimum would you want to consider moving all available formats to something like this? ie removing `smol-toml` `js-yaml`

webpro:
> > That makes 4 plugins (prettier, bumpp, changelogen, changelogithub) at this point that "support json5". There's no way back I guess. But how many projects are actually _using_ json5 that throws in the current solution?
> 
> I was! But that's why I'm here trying to fix it, I actually downgraded my prettier.json5 to be just a standard JSON so I could use knip ;)
> 
> IMO, removing JSON5 support would be better than supporting the extension but not the actual spec of it.
> 
> > The project also looks pretty much dead btw
> 
> I considered that, the JSON5 spec hasn't changed so I don't think the package would _need_ to be updated, but I can see the issue with a project that isn't actively maintained.

Apart from the spec. What puts me off is PRs like this getting no attention at all: https://github.com/json5/json5/pull/362. Even if this particular PR isn't great (idk), performance isn't particularly great. So it'd be a shame if all JSON flavors go through this slow parser, effectively slowing down Knip for all users.

> > Looking at https://github.com/unjs/confbox which looks convenient with basically the same dependencies
> 
> That is interesting, to keep dependencies to a minimum would you want to consider moving all available formats to something like this? ie removing `smol-toml` `js-yaml`

Yes, but Knip potentially loads a lot of JSON* files and this seems to keep a ref of each obj in memory. We could try that out and PR downstream if necessary first, though, to optimize read perf and mem usage.

kiliantyler:
I threw together a benchmark for different parsing libs, confbox looks *promising*. Definitely has issues with memory, especially when it comes to JSON5 parsing, like you pointed out. Speed-wise it's pretty close to par with `strip-json-comments`

It also shows pretty clearly ***do not use json5 parsing on json files*** lol

[Results](https://kiliantyler.github.io/conf-parse-testing/)
[Repo](https://github.com/kiliantyler/conf-parse-testing)

webpro:
Thank you so much, that's amazing and super helpful.

> IMO, removing JSON5 support would be better than supporting the extension but not the actual spec of it.

Indeed, my bad. Gave it some thought and at this point I'm leaning towards:

1. For `.json` files use `JSON.parse` (builtin) in a try/catch with additional JSONC support as the fallback (builtin+strip-json-comments), since the built-in JSON parser is so optimized these days it'd be a shame **not** to _try_ and use it? Isn't it basically only `tsconfig.json` that does this so we could flow `tsconfig*.json` files directly through 2:
2. For `.jsonc` use builtin+strip-json-comment. 
3. Deprecate/drop official support for JSON5, and start showing a warning if parsing fails (there's quite the percentage of JSON5 that goes through the parser in 1).

Thinking of all the bytes & cycles we save. And a perf win too. WDYT?

kiliantyler:
I definitely agree with 1 & 2

for 3:
Thinking about the Dev experience I don't love the idea of dropping JSON5 support because, as you said, Knip supports some plugins that do support JSON5. So it puts it in a weird situation

Dev has JSON5 files in "supported" plugins:
- Likes the idea of using knip
- tries to run it `bunx knip`
- gets an error saying "Yeah technically this supports prettier, but not *your* configuration of prettier"
- Dev now has to choose to use new software that they aren't invested in, or changing the configuration file of the software they already are using.

I do realize that converting from json5 to jsonc isn't *hard*, but it does put up a small barrier that wasn't expected.

I don't think there's a good solution to fixing that considering that `confbox` just uses the same old `json5` package under the hood and `@croct/json5-parser` is pretty much an unknown package.

Short of writing a new json5 parser, you might be right that JSON5 just needs to be dropped, or attempted with `strip-json-comment` and then printing out an error if that fails

webpro:
> or attempted with `strip-json-comment` and then printing out an error if that fails

That's what I meant and went with this idea 💯

Thanks again for raising this and doing the benchmarks

webpro:
:rocket: _This issue has been resolved in v5.66.0. See [Release 5.66.0](https://github.com/webpro-nl/knip/releases/tag/5.66.0) for release notes._

PR #1331, opened 2025-11-03
fix: 🐛 re-support `"astro:env/*"` imports

jimmy-guzman:
Astro's virtual modules like `astro:env/client` and `astro:env/server` contain slashes, which prevented `sanitizeSpecifier` from treating them as Astro dependencies. Updated logic to handle all `astro:*` imports regardless of whether they contain slashes.

This regression started with v5.64.0

_I apologize if I didn't follow contribution guidelines correctly._

### DIFF
--- a/packages/knip/src/util/modules.ts
+++ b/packages/knip/src/util/modules.ts
@@ -48,22 +48,37 @@ export const getPackageFromDefinitelyTyped = (typedDependency: string) => {
   return typedDependency;
 };
 
+const CHAR_EXCLAMATION = 33; // '!'
+const CHAR_DASH = 45; // '-'
+const CHAR_SLASH = 47; // '/'
+const CHAR_COLON = 58; // ':'
+const CHAR_HASH = 35; // '#'
+const CHAR_QUESTION = 63; // '?'
+
 // Strip `?search` and other proprietary directives from the specifier (e.g. https://webpack.js.org/concepts/loaders/)
 export const sanitizeSpecifier = (specifier: string) => {
   if (isBuiltin(specifier) || isAbsolute(specifier) || specifier.startsWith(PROTOCOL_VIRTUAL)) return specifier;
-  let s = specifier;
-  let end = s.length;
-  let i = 0;
-  while (i < s.length && (s[i] === '!' || s[i] === '-')) i++;
-  s = s.substring(i);
-  for (let j = 0; j < s.length; j++) {
-    const char = s[j];
-    if (char === '!' || char === '?' || (char === '#' && j > 0)) {
-      end = j;
+  const len = specifier.length;
+  let start = 0;
+  let end = len;
+  let colon = -1;
+  let sawSlash = false;
+  for (let i = 0; i < len; i++) {
+    const ch = specifier.charCodeAt(i);
+    if (i === start && (ch === CHAR_EXCLAMATION || ch === CHAR_DASH)) {
+      start++;
+      continue;
+    }
+    if (ch === CHAR_SLASH && colon === -1) {
+      sawSlash = true;
+    }
+    if (colon === -1 && ch === CHAR_COLON && !sawSlash) {
+      colon = i;
+    }
+    if (ch === CHAR_EXCLAMATION || ch === CHAR_QUESTION || (ch === CHAR_HASH && i > start)) {
+      end = i;
       break;
     }
   }
-  s = s.substring(0, end);
-  if (/[^/]+:/.test(s)) s = s.split(':')[0];
-  return s;
+  return colon !== -1 && colon < end ? specifier.slice(start, colon) : specifier.slice(start, end);
 };


packages/knip/src/util/modules.ts
```diff
@@ -64,6 +64,7 @@ export const sanitizeSpecifier = (specifier: string) => {
     }
   }
   s = s.substring(0, end);
-  if (s.includes(':') && !s.includes('/')) s = s.split(':')[0];
+  if (s.startsWith('astro:')) s = 'astro';
+  else if (s.includes(':') && !s.includes('/')) s = s.split(':')[0];
```
webpro:
```suggestion
  if (/[^/]+:/.test(s)) s = s.split(':')[0];
```

So if there's no slash before the colon, return the part before the colon. Otherwise it could be a path param (`/:param`). Node built-ins like `node:fs` are caught with `isBuiltin`. Not that it's good or pretty or whatever, but here we are.. WDYT?

jimmy-guzman: Thanks, I like it, added in fb4313d2586de3f4ef87efd26115b1387c06ca4a

As a side note, 

Another approach to `sanitizeSpecifier` is instead of `while` that checks character by character we can use a replace. Then instead of using a for loop with a conditional break and end variable tracking we can use a `search` to get the index directly. We check `i > 0` instead of `i !== -1` because we want to preserve leading `#` (like `#specifier`), which is valid. Then instead of `test()` and `split()` we can just use a single `exec()` with optional chaining. 

Or in other words, replace manual loops with targeted regex operations. 

For example, 

```ts
export const sanitizeSpecifier = (specifier: string) => {
  if (isBuiltin(specifier) || isAbsolute(specifier) || specifier.startsWith(PROTOCOL_VIRTUAL)) return specifier;
  let s = specifier.replace(/^[!-]+/, "");
  const i = s.search(/[!?]|#./);
  if (i > 0) s = s.slice(0, i);
  return /^([^/]+):/.exec(s)?.[1] ?? s;
};
```

I can push this change as well if you're interested (all tests still pass)

webpro: I came from a regex solution and wanted to fix a bug, increase maintainability and also improve performance (this function is used _a lot_ in pretty hot code path - maybe results should even be cached).

So in that sense, your initial solution was good and my proposal was actually not that great in terms of performance: re-introduced a regex while there's already an iteration.

So now that we're on it.. what do you think of maximizing performance with a single iteration, ideally no regexes, while fixing your bug too? Do you maybe see a way to go about this?

jimmy-guzman: That makes sense and I might have an idea on how to tick all boxes, going to try a few things then push something up.

jimmy-guzman: Thoughts on 

```ts
const CHAR_EXCLAMATION = 33; // '!'
const CHAR_DASH = 45; // '-'
const CHAR_SLASH = 47; // '/'
const CHAR_COLON = 58; // ':'
const CHAR_HASH = 35; // '#'
const CHAR_QUESTION = 63; // '?'

// Strip `?search` and other proprietary directives from the specifier (e.g. https://webpack.js.org/concepts/loaders/)
export const sanitizeSpecifier = (specifier: string) => {
  if (isBuiltin(specifier) || isAbsolute(specifier) || specifier.startsWith(PROTOCOL_VIRTUAL)) return specifier;
  const len = specifier.length;
  let start = 0;
  let end = len;
  let colon = -1;
  let sawSlash = false;
  for (let i = 0; i < len; i++) {
    const ch = specifier.charCodeAt(i);
    if (i === start && (ch === CHAR_EXCLAMATION || ch === CHAR_DASH)) {
      start++;
      continue;
    }
    if (ch === CHAR_SLASH && colon === -1) {
      sawSlash = true;
    }
    if (colon === -1 && ch === CHAR_COLON && !sawSlash) {
      colon = i;
    }
    if (ch === CHAR_EXCLAMATION || ch === CHAR_QUESTION || (ch === CHAR_HASH && i > start)) {
      end = i;
      break;
    }
  }
  return colon !== -1 && colon < end ? specifier.slice(start, colon) : specifier.slice(start, end);
};
```
- Single-pass algorithm instead of multiple passes (while loop, substring, for loop, regex test, split)
- Uses `charCodeAt()` for numeric comparisons instead of string character comparisons
- Caches string length to avoid repeated `.length` lookups
- Extracts substrings only once at the end instead of multiple intermediate `substring()` calls
- Eliminates regex operations - replaces `/[^/]+:/.test()` and `.split(':')` with inline tracking

We can also add simple unbounded `Map` based cache if we want.

webpro: Wonderful! Ticks all the boxes 😃 If this passes the tests I'm more than happy to merge.

Let's defer caching for now. It could be beneficial, but at the cost of a bit of memory and I'm not yet sure how it'd stack up against other, perhaps more impactful areas such as creating TS programs and disk reads.

jimmy-guzman: Done in [a6415e6](https://github.com/webpro-nl/knip/pull/1331/commits/a6415e6e5b37c791f9c618472a1c9251fa0f80ab)

webpro:
Fixing a bug and improving performance while at it. Thanks for going the extra mile!

webpro:
:rocket: _This pull request is included in v5.68.0. See [Release 5.68.0](https://github.com/webpro-nl/knip/releases/tag/5.68.0) for release notes._

PR #1338, opened 2025-11-07
Add missing dependency on remark-gfm

remcohaszing:
This dependency is used as a transient dependency of `@astrojs/markdown-remark`. This is configured correctly in `tsconfig.json` so that the MDX language server and TypeScript plugin understand it. Previously this dependency was deduped, which worked fine. However, with the way pnpm installs dependencies, transient dependencies aren’t available directly. This causes a crash in the editor integrations.

To solve this, `remark-gfm` was added as a direct dependency.

### DIFF
--- a/packages/knip/src/plugins/mdx/index.ts
+++ b/packages/knip/src/plugins/mdx/index.ts
@@ -1,5 +1,5 @@
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
-import { toDependency } from '../../util/input.js';
+import { type Input, toDependency } from '../../util/input.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { MdxConfig } from './types.js';
 
@@ -11,13 +11,21 @@ const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependenc
 
 const config = ['tsconfig.json'];
 
-const takeDependencies = (config: MdxConfig) =>
-  Array.isArray(config.plugins) ? config.plugins.map(id => toDependency(id)) : [];
+const takeDependencies = (config: MdxConfig) => {
+  const inputs: Input[] = [];
+  if (Array.isArray(config.plugins)) {
+    for (const plugin of config.plugins) {
+      if (typeof plugin === 'string') inputs.push(toDependency(plugin));
+      else if (typeof plugin[0] === 'string') inputs.push(toDependency(plugin[0]));
+    }
+  }
+  return inputs;
+};
 
 const resolveConfig: ResolveConfig<MdxConfig | { mdx: MdxConfig }> = async (config, options) => {
   const { configFileName } = options;
 
-  // read by @mdx-js/typescript-plugin
+  // read by @mdx-js/typescript-plugin (https://github.com/mdx-js/mdx-analyzer#plugins)
   if (configFileName === 'tsconfig.json' && 'mdx' in config) return takeDependencies(config.mdx);
 
   return [];
--- a/packages/knip/fixtures/plugins/mdx/package.json
+++ b/packages/knip/fixtures/plugins/mdx/package.json
@@ -3,6 +3,7 @@
   "devDependencies": {
     "astro": "*",
     "remark-directive": "*",
+    "remark-frontmatter": "*",
     "typescript": "*"
   }
 }
--- a/packages/knip/fixtures/plugins/mdx/tsconfig.json
+++ b/packages/knip/fixtures/plugins/mdx/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "extends": "astro/tsconfigs/strict",
   "mdx": {
-    "plugins": ["remark-directive"]
+    "plugins": [["remark-frontmatter", ["toml", "yaml"]], "remark-directive"]
   }
 }
--- a/packages/knip/src/plugins/mdx/types.ts
+++ b/packages/knip/src/plugins/mdx/types.ts
@@ -1,3 +1,3 @@
 export type MdxConfig = {
-  plugins?: string[];
+  plugins?: (string | [string, unknown])[];
 };


webpro:
Thanks! Agreed.

Instead of adding it to `ignoreDependencies` I'll make Knip understand the TS config entries.

remcohaszing:
Even better!

webpro:
Added a minimal (and sligthly odd) MDX plugin to resolve this:

- enabled by having `astro` (e.g. `@astrojs/starlight` has it as a peer dep) or `mdxlint` listed in `package.json`
- reads config at `tsconfig.json#mdx` and currently just returns `plugins` as dependencies

For now I've just put it in a generic MDX plugin, but if that grows unwieldy or doesn't make sense already, the part we have here could become a separate plugin (to handle this `@mdx-js/typescript-plugin` part).

If you agree I'm happy to merge :)

packages/knip/src/plugins/mdx/index.ts
```diff
@@ -0,0 +1,40 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import { type Input, toDependency } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { MdxConfig } from './types.js';
+
+const title = 'MDX';
+
+const enablers = ['astro', 'mdxlint'];
```
remcohaszing, line 8: This is related to [`@mdx-js/typescript-plugin` / `@mdx-js/language-server`](https://github.com/mdx-js/mdx-analyzer), not `astro` or `mdxlint`

packages/knip/src/plugins/mdx/index.ts
```diff
@@ -0,0 +1,40 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import { type Input, toDependency } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { MdxConfig } from './types.js';
+
+const title = 'MDX';
```
remcohaszing, line 6: If you have a generic MDX plugin, you may want to add support for Turbopack MDX configuration as well. Though that might be out of scope for now. https://nextjs.org/docs/pages/guides/mdx#using-plugins-with-turbopack

packages/knip/src/plugins/mdx/index.ts
webpro: I know, but this is how plugins currently work: they're enabled cheaply by looking at dependencies in `package.json`. There's no extra disk reads needed here as both `package.json` (and `tsconfig.json`) are read already.

`isEnabled` could do anything async, but I've generally avoided not doing the default action like we do now. Might want to reconsider as we get into more odd situations like this.

packages/knip/src/plugins/mdx/index.ts
webpro: That's out of scope in this PR, but it's def useful to get a feel of what we'll need to handle.

packages/knip/src/plugins/mdx/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  return [];
+};
+
+export default {
+  title,
+  enablers,
+  isEnabled,
+  config,
+  resolveConfig,
+} satisfies Plugin;
```
remcohaszing, line 40: Just a tip: `satisfies` is not a type annotation. This tells TypeScript that the default export is of a new inferred type that has to satisfy `Plugin`. On a large scale this can decrease TypeScript performance. Also unnamed default exports result in a variable named `_default` in the emitted declaration files.

Typically I recommend to use a type annotation instead. `satisfies` has its uses, but should not be considered the default option.

Also splitting a default export into a declaration and a separate export, gives you control over the name. This affects autocompletions and readability of the type declarations.

I made a little [TypeScript playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAkgzgUQHYEMBGAbCATKBeKACgEp8A+KNAeyqxSQCgGBLJYCAJwDMUBjaAAoYArgHNWUAN4MosqBFSZOcAFxQ4wDq1EBtALoy5zRIqzY18ZOjMMAvk15UkGqGBHikaoWIkFpshWtlNX0AGkNjKyVzIlI8Ck1hCDsGCAAPMCoOYChsCB5hDBy3H0ZUjKycvIKiqUMA02CoMPqoSMaYknIoROTbdRRgYy5mCDgobw8gA) to elaborate what I mean. Have a look at the **.D.TS** tab.

packages/knip/src/plugins/mdx/index.ts
remcohaszing: In that case I think there could be more enablers, but this is a good start.

packages/knip/src/plugins/mdx/index.ts
webpro: Thanks for the explanation! Adding a type notation is a good idea, e.g.:

```
const plugin: Plugin = {};
export default plugin;
```

..leading to a `.d.ts` like

```ts
export declare const Plugins: {
    angular: import("../types/config.js").Plugin;
    astro: import("../types/config.js").Plugin;
}
```

Much better indeed. Will pick that up shortly.

I do think I want to keep the default exports, as I believe that's convenient for plugin authors. And I'd rather bundle them using a script¹, resulting in kebab-cased ~exports~ plugin names used by the actual schema for `knip.json`

¹ https://github.com/webpro-nl/knip/blob/main/packages/knip/scripts/generate-plugin-defs.js generates https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/index.ts

packages/knip/src/plugins/mdx/index.ts
webpro: I'm the first to admit this plugin is odd, it will likely not stay this way but let's just roll with it for now.

webpro:
:rocket: _This pull request is included in v5.69.0. See [Release 5.69.0](https://github.com/webpro-nl/knip/releases/tag/5.69.0) for release notes._

Issue #1062, opened 2025-04-30
💡 Mark prisma generators as used devDependencies


WikiRik:
### Suggest an idea for Knip

It's related to #1052 in the sense that it is a feature request for the (not yet existing) prisma plugin.

The prisma schema allows you to set custom generators from the community. See the documentation on that here; https://www.prisma.io/docs/orm/prisma-schema/overview/generators#community-generators

This results in for example the following snippet from a prisma schema file;
```prisma
generator json {
  provider = "prisma-json-types-generator"
}
```

The `prisma-json-types-generator` comes from the package with the same name which is now listed as a unused devDependency. It would be great if knip could recognize that it's actually used.

webpro:
Perhaps oddly, this might be a good candidate for a custom compiler: https://knip.dev/features/compilers

Essentially it's text in and text out, so this schema..

```prisma
generator json {
  provider = "prisma-json-types-generator"
}
```

...becomes TypeScript...

```ts
import("prisma-json-types-generator");
```

...by this `knip.ts` config...


```ts
export default {
  compilers: {
    schema: (text: string) => ""; // hack some regex(es) or use an actual compiler
  }
}
```

If there are no dependencies I'd consider adding it to Knip core.

The other thing that'd need to happen is either add the `src/**/*.schema` file patterns to `entry: []` files manually to your config, or add those to the new Knip plugin `entry` default patterns. But essentially all of this could be done apart from the plugin (obviously would be good to eventually merge it to something that's "zero-config").

WikiRik:
That would work for my use case, but I'm not entirely sure if the other community generators also follow the pattern where the name of the package is also the name of the generator. I haven't read enough about it to know for sure

webpro:
FYI, there is a Prisma plugin now, and you could add such a custom compiler to your own config.

webpro:
Closing this issue as part of a general cleanup to keep this project sustainable and optimize my time working on it. If you think this is inappropriate or if there is no workaround and you feel stuck, feel free to [open a new issue](./new/choose). Thanks for your understanding.

CHC383:
If anyone needs a working solution:

```ts
import type { KnipConfig } from "knip";

export default {
  entry: ["prisma/schema.prisma"],
  compilers: {
    prisma: (text: string) => {
      const directiveMatcher =
        /generator\s+(?!client)\w+\s*\{\s*provider\s*=\s*"([^"]+)"[^}]*\}/g;
      const imports = [];
[...]

PR #1339, opened 2025-11-07
feat(plugins): extend prisma plugin

CHC383:
- Add more Prisma config patterns
- Parse seed command from Prisma config file

PR #1340, opened 2025-11-08
feat(plugins): Add Prisma plugin entry and Prisma schema compiler

CHC383:
Close: https://github.com/webpro-nl/knip/issues/1062

- Add Prisma schemas as the entry of the Prisma plugin
- Add Prisma schema compiler

Note:
1. Please let me know if we don't want to add a built-in compiler for Prisma schema as it is not as widely used as the others, I am totally fine with that.
2. This stacks on top of https://github.com/webpro-nl/knip/pull/1339 and will be ready for review after that one is settled.
3. Please let me know if it is preferred to split this into two PRs, one for adding the entry (without processing) and one for the compiler.
4. Not sure if I need something like https://github.com/webpro-nl/knip/commit/4f3bd099fadedf59c40f130f1d5fa22da8b2b79c

### DIFF
--- a/packages/docs/src/content/docs/features/compilers.md
+++ b/packages/docs/src/content/docs/features/compilers.md
@@ -17,7 +17,7 @@ Knip has built-in "compilers" for the following file extensions:
 - `.svelte`
 - `.vue`
 - `.css` (only enabled by `tailwindcss`)
-- `.prisma` (only enabled by `prisma`)
+- `.prisma`
 
 Knip does not include real compilers for those files, but regular expressions to
 collect `import` statements. This is fast, requires no dependencies, and enough
--- a/packages/knip/src/plugins/prisma/index.ts
+++ b/packages/knip/src/plugins/prisma/index.ts
@@ -3,7 +3,7 @@ import type { Args } from '../../types/args.js';
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
 import { isDirectory } from '../../util/fs.js';
 import { type Input, toEntry } from '../../util/input.js';
-import { extname, join } from '../../util/path.js';
+import { join } from '../../util/path.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { PrismaConfig } from './types.js';
 


webpro:
Thanks, Henry! I like it. The idea is that they're "cheap" and not enabled without `prisma` in the deps so this is cool.

packages/knip/src/plugins/prisma/index.ts
```diff
@@ -11,21 +13,53 @@ const enablers = ['prisma', /^@prisma\/.*/];
...
+    // package.json
+    inputs.push(...options.getInputsFromScripts(config.seed));
+  } else if (config.migrations?.seed) {
+    // Prisma config file
+    inputs.push(...options.getInputsFromScripts(config.migrations.seed));
+  }
+
+  // Entry/Schema
+  // https://www.prisma.io/docs/orm/prisma-schema/overview/location#prisma-schema-location
+  const mergedEntry: string[] = [...entry];
```
CHC383: @webpro Just to double check, during my development, I found out that:
1. `resolveConfig` is not called if Prisma config is not presented and `package.json` doesn't include a section for `Prisma`, so the default `entry` is necessary.
2. If `resolveConfig` is called and the returned `Inputs[]` includes `entry` elements, the default `entry` will be overwritten instead of automatically being merged, so I need to merge the entry explicitly here.

webpro: Correct! Two reasons:

- The default `entry` is used to generate the plugin docs
- In general, that's how Knip works: `entry`, `project` and `config` can be overridden and the only way for end-users to get rid of an item is to replace all items of the array.

webpro: Also see this example in the Mocha plugin: the end result is either the entry paths from the local Mocha config (`spec`) or the default `entry`: https://github.com/webpro-nl/knip/blob/d922df43f501b42afd93de21641247b14d6807e3/packages/knip/src/plugins/mocha/index.ts#L19

packages/knip/src/plugins/prisma/index.ts
webpro, line 58: We don't need to parse script manually, this can replace `schemaFromScript`:

```suggestion
  config: true,
  resolveInputs: (parsed: ParsedArgs) => {
    const inputs: Input[] = [];
    if (parsed['schema']) inputs.push(toEntry(parsed['schema']));
    return inputs;
  },
```

packages/knip/src/plugins/prisma/index.ts
CHC383: Based on the [Prisma doc](https://www.prisma.io/docs/orm/prisma-schema/overview/location#prisma-schema-location), its CLI looks for the schema file in the following order:
1. The `schema` flag of the CLI command
2. The schema defined in the config/package.json
3. The default schemas

I guess it is fair to say that if the schemas are detected through flag or config, then we don't need the default `entry`, so using one or the other as the Mocha plugin makes sense.

And I just found out [multi-file Prisma schema](https://www.prisma.io/docs/orm/prisma-schema/overview/location#multi-file-prisma-schema), so I think the logic needs to be updated to check if the path is a directory and globs for the `.prisma` files inside. Is there any existing utils to scan the directory or can we use Node `fs.glob`?

I wonder if it is better to just define `entry` as `["**/*.prisma"]` as the suffix is pretty unique, then we don't need to worry about parsing the arguments or configs. On the one hand it might include unused schemas when people keeps backup/old schemas but not explicitly referencing them through the approaches above, but on the other hand it could also cover schemas that are used only in hand-typing commands and thus can't be detected by Knip. What's the general guidance for the Knip plugin? To be broader or more precise?

packages/knip/src/plugins/prisma/index.ts
CHC383: Thanks for the tip, if I understand correctly, these entries will be merged with the ones from `resolveConfig` right?

packages/knip/src/plugins/prisma/index.ts
webpro: - There is (sync) `isDirectory` (from util/fs helpers)
- In general, Knip tends to be precise as that leads to better results faster. That said, some plugins such as those for test runners have globs like `**/*.spec.ts` so it's not unheard of. Only if the plugin is enabled will the glob pattern be added. Users can still exclude files by having more specific or negated patterns in their `prisma` config:

```json
{
  "prisma": {
    "entry": ["src/**/*.prisma", "!src/backup/**/*.prisma"]
  }
}
```

User config like this always replaces (not merges) plugin defaults.

More than anything, would be great if we could have as many cases as possible work as expected without the user needing to configure anything.

packages/knip/src/plugins/prisma/index.ts
webpro: Indeed, `inputs` returned here are always added. There is no inter-play with `resolveConfig` (i.e. in `resolveConfig` we have the `config` but not the args, and vice-versa).

packages/knip/src/plugins/prisma/index.ts
CHC383: I tried to use `isDirectory`, and it does require full path of the schema folder to work properly. In `resolveConfig` we can join `options.cwd` with the schema path, is there a way to get the `cwd` in `resolveInputs`?

Update: For now I am just checking the extension of the path to determine if it is a directory or file as a workaround.

webpro: Thanks!

Good idea, I've added `options.cwd` and the `isDirectory` check, hope you don't mind :)

packages/docs/src/content/docs/features/compilers.md
```diff
@@ -17,6 +17,7 @@ Knip has built-in "compilers" for the following file extensions:
 - `.svelte`
 - `.vue`
 - `.css` (only enabled by `tailwindcss`)
+- `.prisma` (only enabled by `prisma`)
```
webpro:
```suggestion
- `.prisma`
```

webpro (commented):
Thank you so much, I think we're through, right?

Imho we could do with less comments, but really there are no blockers, happy to merge!

CHC383:
> Thank you so much, I think we're through, right?
> 
> Imho we could do with less comments, but really there are no blockers, happy to merge!

Yes it is good to go, thanks for your patience and the helps.

webpro:
🙏

webpro:
:rocket: _This pull request is included in v5.70.0. See [Release 5.70.0](https://github.com/webpro-nl/knip/releases/tag/5.70.0) for release notes._

Issue #1245, opened 2025-09-03
🐛 Astro compiler elides imports when there are template literals in the file


Ivo-Evans:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/webpro-nl/knip/compare/main...Ivo-Evans:knip:astro-fix-blacktick-support

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

#1041 added support for import syntax that isn't evaluated inside strings, e.g. 

```ts
dedent`
  import Button from '~/components/Button'
`
```

However, the regex used, `/\w+(?:\.\w+)*`[\s\S]*?`/g` is incorrect. The regex looks for word characters next to template tags. In this example the first and fourth backticks are not in the match, causing the compiler to elide everything between the two template strings, but nothing inside them, removing the import:

```astro
---
Astro.response.headers.set('Cache-Control', `max-age=1800, s-maxage=1800, stale-while-revalidate=120`);

import { used } from used

const message = `used ${used}`
---

{used}
```

I'm happy to work on this but I'd like to decide what to do first. One option is to revert #1041 - I can see that a lot of work was done on it, but on the other hand it's a niche feature with a high chance of backfiring if there's an edge case in the regex. It could be supported on an ad-hoc basis by custom compilers on specific knip installations.

Another option is to change the regex, either changing `w+` to `w*` or removing `w+` (this affects whether we match the string before the template literal or not). This does seem to correctly match the three cases I can think of: template literals used as strings, tagged template literals and markdown code examples inside comments: 

````astro
---
Astro.response.headers.set('Cache-Control', `max-age=1800, s-maxage=1800, stale-while-revalidate=120`);

/*
use this component like
```
---
import MyComponent from "./my-component"
---

<MyComponent />
```
*/

const foo = deftest`import foo from "foo"`

import Child from "./child"
---

<Child />
````

[...]

PR #1349, opened 2025-11-11
Use first-party Astro compiler

Ivo-Evans:
First off, this PR is not as big as it looks. It introduces two new sets of fixtures, which pumps up the diffs, but the significant changes are only in [`compilers/astro.ts`](https://github.com/webpro-nl/knip/pull/1349/files#diff-c20b01ad9b05b47cdae37f0ae3583d53a87d6159a46881ba22f0e147c5993580), [`compilers/index.ts`](https://github.com/webpro-nl/knip/pull/1349/files#diff-4af7e849a9e66964aa47b458b45d4747eb0a1f6227dca96baa1913faf3b0843a) and arguably [`test/plugins/astro.test.ts`](https://github.com/webpro-nl/knip/pull/1349/files#diff-be7cc4e9c38ef49b3a97dda5ed9ad55fb0b1f040dfaff759b46f22c70b0b4bca).

These changes introduce the Astro first-party compiler as Knip's compiler for `.astro` files. The new Astro compiler is the only built-in compiler which is async, so the PR also adds support for built-in async compilers.

The PR solves two specific problems. Firstly, we fix bug #1245 where we sometimes delete chunks of the file that contain imports. Secondly, it adds support for imports like `<script src="../fullstory.js" />`, which Astro supports. 

One thing I _haven't_ done is add support for setting `compilers: {astro: true }`, enabling an async compiler manually. I can see evidence for this functionality for sync compilers, e.g. here https://github.com/webpro-nl/knip/blob/68efbd55128947642f561284482b827d31d85fec/packages/docs/src/content/docs/features/compilers.md#L36 but when I tried I couldn't get it working and `compilers` isn't in the JSON schema [here](https://unpkg.com/knip@5.69.0/schema.json). Maybe you can advise.


Ivo-Evans:
I'm not sure why the tests are failing, I'll check tomorrow

webpro:
Thanks for your pull request, Ivo. The idea of the built-in "compilers" is that they don't require dependencies.

They are, indeed, not perfect, and that's why users can plug in their own with as little effort as possible.

In the past I've put a sequence of efforts into getting the [dependency graph](https://npmgraph.js.org/?q=knip) smaller. We can still improve in this area, and I'd prefer not to add dependencies. That's why I'm going to close this PR.

If there's other ways we can improve this part of Knip I'm obviously all ears. Did you experience issues with the built-in Astro compiler? Maybe we should add an example to the docs to help users adding the first-party Astro compiler themselves?

Issue #960, opened 2025-02-25
💡 Respect `pageExtensions` option in Next.js config


kachkaev:
### Suggest an idea for Knip

👋 @webpro! We use knip in a Next.js project and I’ve noticed one false-negative which is to do with how our app is setup. When [`pageExtensions` option](https://nextjs.org/docs/pages/api-reference/config/next-config-js/pageExtensions) is set in `next.config.ts`, standard globs from the plugin are no longer valid:

https://github.com/webpro-nl/knip/blob/c3664373828751945ebad38a617135df6047dced/packages/knip/src/plugins/next/index.ts#L14-L23

We’ve got this in `next.config.ts`:

```tsx
pageExtensions: ["page.tsx", "handler.ts"],
```

With this setting, the following files would be entry points:

```
pages/index.page.tsx
pages/about.page.tsx
pages/product/[slug].page.tsx
pages/ping.handler.ts
```

However, these files would not be entry points and should be reported, if unused:

```
pages/shared/logo.tsx
pages/about.page/map.tsx
pages/product/[slug].page/interactive-3d-model.tsx
pages/ping.handler/helpers.ts
```

At the moment, `knip` is unaware of `pageExtensions` so marks _all_ files inside the `pages` folder as used.

I understand that `pageExtensions` may be a tricky thing to support, because essentially `knip` needs to _run_ `next.config.ts` to get the right value. Perhaps, at least we could disable standard globs is `pageExtensions` is found anywhere in file. In the meantime, I’ll see if I can manually turn off the Next.js plugin to define custom globs. Open to other ideas too!

webpro:
Knip _could_ run `next.config.ts` (it does this for many plugins).

Feel free to take a stab at it, there are a few plugins that do something similar already:

- https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/nuxt/index.ts
- https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/eleventy/index.ts

So the idea is to implement `resolveEntryPaths` and return `toProductionEntry` inputs.

I'm not sure about potential issues with loading `next.config.ts` but in general this doesn't give issues (apart from users often not having environment variables set).

kachkaev:
I see, interesting! I don’t think I’ll have capacity to submit a PR in the next couple of weeks, so it’s up for grabs. In the meantime, I’ve added a local pnpm patch (`patches/knip.patch`) and it seems to be working!

```diff
[...]

PR #1351, opened 2025-11-12
Apply Next.js page extensions to app directory

remcohaszing:
The Next.js option `pageExtensions` applies to both the app router and the pages router. Knip however, only applied it to the pages router files, leading to many false positives.

### DIFF
--- a/packages/knip/test/plugins/next.test.ts
+++ b/packages/knip/test/plugins/next.test.ts
@@ -11,7 +11,6 @@ const cwd = resolve('fixtures/plugins/next');
 test('Find dependencies with the Next.js plugin', async () => {
   const options = await createOptions({ cwd });
   const { issues, counters } = await main(options);
-  console.log(issues);
 
   assert(issues.files.has(join(cwd, 'app/unused.ts')));
   assert(issues.files.has(join(cwd, 'pages/unused.jsx')));


webpro:
Thanks for the catch & fix, much appreciated

webpro:
:rocket: _This pull request is included in v5.69.1. See [Release 5.69.1](https://github.com/webpro-nl/knip/releases/tag/5.69.1) for release notes._

kachkaev:
I believe that files like `instrumentation.*` need to be treated like app files too: https://github.com/webpro-nl/knip/issues/960#issuecomment-3480697351

WDYT @remcohaszing?

remcohaszing:
What do you mean by that? Should for example `instrumentation.tsx` also be supported?

kachkaev:
It's already supported, but not with custom extensions. Looks like instrumentation (and maybe proxy / middleware) need the same trick with file extensions as files in app and pages dirs.

remcohaszing:
I’m not sure which files are affected by `pageExtensions`. I believe you if you say more files are supported, but I don’t know. I just handled additional ones of which I **know** they are affected.

I would wait for a user to report this. I can’t think of a scenario where the user wants to write `proxy.tsx` or `instrumentation.mdx`. If you are running into the problem that you use `instrumentation` with a custom extension, then I totally agree that should be fixed.

iwan-uschka:
I am working on a project with the following false positives reported by knip:
- `instrumentation.node.ts` (https://nextjs.org/docs/app/guides/open-telemetry)
  > We recommend creating a new file `instrumentation.node.ts` which you conditionally import only when using node:
- `instrumentation.page.ts` (https://nextjs.org/docs/pages/api-reference/config/next-config-js/pageExtensions)
  > Changing these values [of `pageExtensions`] affects _all_ Next.js pages, including the following:
  > 
  > - [`proxy.js`](https://nextjs.org/docs/pages/api-reference/file-conventions/proxy)
  > - [`instrumentation.js`](https://nextjs.org/docs/pages/guides/instrumentation)
  > - `pages/_document.js`
  > - `pages/_app.js`
  > - `pages/api/`
  >
  > For example, if you reconfigure `.ts` page extensions to `.page.ts`, you would need to rename pages like `proxy.page.ts`, `instrumentation.page.ts`, `_app.page.ts`.
- `middleware.page.ts` (https://nextjs.org/docs/15/pages/api-reference/file-conventions/middleware)
  > If you’ve customized [`pageExtensions`](https://nextjs.org/docs/15/app/api-reference/config/next-config-js/pageExtensions), for example to `.page.ts` or `.page.js`, name your file `middleware.page.ts` or `middleware.page.js` accordingly.

The `next.config.js` file contains `pageExtensions: ['page.ts', 'page.tsx', 'api.ts']` which extend the default Page extensions (`.tsx`, `.ts`, `.jsx`, `.js`) used by Next.js.

@remcohaszing does this provide some missing information?

I am new to knip so i don't feel confident enough to work on a PR right now.

remcohaszing:
I can’t tell without a reproduction. I suggest you create a minimal reproduction and report a new issue.

iwan-uschka:
Done: https://github.com/webpro-nl/knip/issues/1581. Thank you @remcohaszing.

PR #1361, opened 2025-11-19
Support Yarn PnP

parksb:
This PR continues the work from #924. Thanks @jwoo0122 :)

When a project uses Yarn PnP, this PR makes knip look for manifest files in the PnP file instead of searching in the `node_modules` directory. Previously, knip incorrectly reported peer dependencies required by packages as unused dependencies in Yarn PnP projects.

Below is an example where knip reports react and react-dom as unused dependencies in a project using styled-components. (styled-components requires react and react-dom as peer dependencies.)

| before |
|--|
| SCR-20251119-jwfr |

With Yarn PnP support in this PR, knip correctly reports no unused dependencies as shown below.

| after |
|--|
| SCR-20251119-jway |


parksb:
For test consistency, I've added the Yarn PnP cache files. This is equivalent to committing the `node_modules` directory to git. Without adding this, you would need to run yarn install once in the `fixtures/yarn-pnp` directory before running the tests.

packages/knip/src/manifest/helpers.ts
```diff
@@ -36,3 +49,63 @@ export const getFilteredScripts = (scripts: Scripts) => {
...
+    if (pnpStatus.dir !== dir) {
+      findNearestPnPFile(dir);
+    }
+
+    if (pnpStatus.enabled) {
+      const pnpApi = _require(pnpStatus.pnpPath);
+
+      if (pnpApi != null) {
+        const packageJsonPath = join(packageName, 'package.json');
+        const resolvedPath = pnpApi.resolveToUnqualified(packageJsonPath, dir);
```
parksb, line 99: I used [`resolveToUnqualified`](https://yarnpkg.com/advanced/pnpapi#resolvetounqualified) to bypass exports map restrictions. This is necessary for reading metadata (bin, peerDependencies, types, ...) even when:

1. Workspace packages don't have `./package.json` entry in exports map
2. Packages have no JavaScript entry point (e.g., `@types/*` packages)

Actual resolution (oxc-resolver) still respects exports map, so invalid imports will be correctly detected.

webpro:
Sorry for replying so late, but I just don't have the bandwidth to review and, more importantly, maintain this. That might change at some point, but it's been a hobby project for >3 years now so let's not hold our breath for it.

PR #1380, opened 2025-12-08
Add support for next-intl request.ts

remcohaszing:
Next-intl requires you to define a request configuration. By default this is loaded from intl/request.ts.

See https://next-intl.dev/docs/getting-started/app-router

### DIFF
--- a/packages/knip/src/plugins/next/index.ts
+++ b/packages/knip/src/plugins/next/index.ts
@@ -21,6 +21,7 @@ const productionEntryFilePatterns = [
   'app/**/sitemap.{js,ts}',
   'app/**/{icon,apple-icon}.{js,jsx,ts,tsx}',
   'app/**/{opengraph,twitter}-image.{js,jsx,ts,tsx}',
+  'i18n/request.{js,jsx,ts,tsx}',
   'mdx-components.{js,jsx,ts,tsx}',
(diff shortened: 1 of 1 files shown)


webpro:
This shouldn't be added to the Next.js plugin, but requires a new `next-intl` plugin.

remcohaszing:
I followed the implementation for `mdx-components`, which is only used if `@next/mdx` is used. Should that have its own plugin as well then?

webpro:
Yes. That has apparently slipped through back then.

And as mentioned recently, the `@next/mdx` case with `createMDX` options in `next.config.js` is very convoluted. We should probably split that one as well... the same AST (`ts.SourceFile`) for `next.config.ts` can be reused in separate plugins that have the same `config` file (it's still in memory so it's not inefficient).

In general the idea is to try and consolidate "enabling" packages with their own entry and config files in separate plugins as much as possible.

webpro:
[It's working](https://github.com/webpro-nl/knip/blob/main/AGENTS.md#plugins) 😬

webpro:
:rocket: _This issue has been resolved in v5.73.0. See [Release 5.73.0](https://github.com/webpro-nl/knip/releases/tag/5.73.0) for release notes._

PR #1384, opened 2025-12-10
fix(plugins/astro): srcDir config detection creating wrong entry patterns

viktormarinho:
My astro project was not detecting the entry files automatically via the plugin so i found it weird. Looking into the plugin seems like the regex had an errant backtick. Just removed that and the local build of the plugin now worked.


webpro:
Oh my 🤦 thanks!

webpro:
:rocket: _This pull request is included in v5.73.1. See [Release 5.73.1](https://github.com/webpro-nl/knip/releases/tag/5.73.1) for release notes._

PR #1386, opened 2025-12-12
fix: detect namespace imports used as default values in destructuring

jstaab:
Fixes detection of namespace imports used as default values in destructuring patterns.

Previously, code like this would report all exports from `./fallback` as unused:

```typescript
import * as Fallback from './fallback';
const { values = Fallback } = props;
values.someValue
```
Now knip correctly recognizes that the namespace is referenced and marks all its exports as used. We're trading in blanket false positive for potential false negative. I'm sure that could be improved to be more specific, but from my experience NS tracking is...hard.

### DIFF
--- a/packages/knip/src/typescript/ast-helpers.ts
+++ b/packages/knip/src/typescript/ast-helpers.ts
@@ -243,6 +243,7 @@ export const isConsiderReferencedNS = (node: ts.Identifier) =>
   ts.isSpreadAssignment(node.parent) ||
   ts.isArrayLiteralExpression(node.parent) ||
   ts.isExportAssignment(node.parent) ||
+  (ts.isBindingElement(node.parent) && node.parent.initializer === node) ||
   ts.isTypeQueryNode(node.parent.parent);
 
 export const isInOpaqueExpression = (node: ts.Node): boolean =>


webpro:
> Now knip correctly recognizes that the namespace is referenced and marks all its exports as used. We're trading in blanket false positive for potential false negative.

Agreed on the trade-offs and that this is better, thank you!

> I'm sure that could be improved to be more specific, but from my experience NS tracking is...hard.

We try to [track members when we can](https://github.com/webpro-nl/knip/blob/542e2ad11603d9652809fdd8c841f24dff402626/packages/knip/src/typescript/visitors/dynamic-imports/importCall.ts#L88).

TypeSript itself has `ts.findReferences` that could do it for us, but that's too slow at scale so the challenge we took upon ourselves is that Knip traverses the AST of any file only once. 

[Unless...](https://knip.dev/guides/performance#language-service), then it uses `findReferences` for exports/member that arent't already referenced in the module graph.

webpro:
:rocket: _This pull request is included in v5.73.4. See [Release 5.73.4](https://github.com/webpro-nl/knip/releases/tag/5.73.4) for release notes._

PR #1390, opened 2025-12-14
Add SWC plugin

ebroder:
Our (Meteor) project uses SWC directly without going through rspack or webpack or some other bundler, which I think is the more common usage model. This adds a plugin to parse and extract dependencies from the `.swcrc` config file.

### DIFF
--- a/packages/knip/src/plugins/swc/index.ts
+++ b/packages/knip/src/plugins/swc/index.ts
@@ -1,0 +1,30 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import { toDependency } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { SWCConfig } from './types.js';
+
+// https://swc.rs/
+// https://swc.rs/docs/configuration/swcrc
+
+const title = 'SWC';
+
+const enablers = ['@swc/core'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['.swcrc'];
+
+const resolveConfig: ResolveConfig<SWCConfig> = async config => {
+  const inputs = config?.jsc?.experimental?.plugins ?? [];
+  return inputs.map(([id]) => toDependency(id));
+};
+
+const plugin: Plugin = {
+  title,
+  enablers,
+  isEnabled,
+  config,
+  resolveConfig,
+};
+
+export default plugin;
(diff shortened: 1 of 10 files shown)


packages/knip/fixtures/plugins/swc/node_modules/@swc/core
```diff
@@ -0,0 +1 @@
+../../../../../../../node_modules/.pnpm/@swc+core@1.15.4_@swc+helpers@0.5.17/node_modules/@swc/core
```
webpro: Fixtures should not rely on nor influence the root (or other) packages. Please undo symlinks like this and the changes to the lockfile.

If any `node_modules/*` are needed, please create the bare minimum. Also see fixtures of other plugins, or AMA :)

packages/knip/src/plugins/swc/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const enablers = ['@swc/core'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['.swcrc'];
+
+const resolveConfig: ResolveConfig<SWCConfig> = async config => {
+  const inputs = config?.jsc?.experimental?.plugins ?? [];
+  return [...inputs].map(([id]) => toDeferResolve(id));
```
webpro: - Nitpick: no need to spread here.
- If we can do with `toDependency` that'd be better, unless there's a reasonable chance local files are referenced (in which case `toDeferResolve` is indeed the best choice here).

packages/knip/test/plugins/swc.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const cwd = resolve('fixtures/plugins/swc');
+
+test('Find dependencies with the swc plugin', async () => {
+  const options = await createOptions({ cwd });
+  const { counters } = await main(options);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    devDependencies: 1,
```
webpro: I'm guessing this is the `@swc/core` dependency. If that normally has some usage/reference e.g. in one of the `package.json#scripts` then it would be good to add that in for a more real-world simulation (also the comment in the test template).

webpro (commented):
Thanks for the PR! Happy to merge, but I did leave some comments.

packages/knip/test/plugins/swc.test.ts
ebroder: Aha, got it. I concluded that didn't apply to me because the issue was with `devDependencies` not with `binaries` (because the `swc` binary comes in with `@swc/cli` not `@swc/core`). Will clean this and the rest up.

ebroder:
Thanks for the review. I believe I've addressed those three issues; let me know if anything else looks off.

webpro:
Thanks, Evan! Nice one.

webpro:
:rocket: _This pull request is included in v5.74.0. See [Release 5.74.0](https://github.com/webpro-nl/knip/releases/tag/5.74.0) for release notes._

Issue #1399, opened 2025-12-18
🐛 Script discovery issues: `op run --env-file=.env -- *` and `--import=foo.js`


hilja:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/hilja/knip-script-discovery

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

I'm using [1Password cli](https://developer.1password.com/docs/cli/secrets-environment-variables#use-environment-env-files) `op` to read `.env` files before running scripts, Knip does not see the scripts coming after:

- `op run --env-file=.env --`
- Or `pnpm op --` if used as a script

```json
{
  "scripts": {
    "start": "op run --env-file=.env -- node server.ts",
    "start:opAsScript": "pnpm op -- node server.ts"
  }
}
```

Also scripts imported with `node --import` are not discovered (they works normally though):

```json
{
  "scripts": {
    "start:mocks": "op run --env-file=.env -- node --import=mocks.ts server.ts"
  }
}
```

Sometimes coupled with inline env vars: `NODE_ENV=production pnpm op -- node server.ts`.

There is a test for the double-dash, but in this case it's not working for some reason.

https://github.com/webpro-nl/knip/blob/20ee7d51110231a903f594ecf944710200954673/packages/knip/test/util/get-inputs-from-scripts.test.ts#L280

Huge thanks for your maintenance work :)

webpro:
:rocket: _This issue has been resolved in v5.75.2. See [Release 5.75.2](https://github.com/webpro-nl/knip/releases/tag/5.75.2) for release notes._

_Using Knip in a commercial project? Please consider [becoming a sponsor](https://knip.dev/sponsors)._

hilja:
Thank you!

Few issues with scripts outlined below. Let me know if I should open a new issues for these, or if they're even in scope.

## Arbitrary scrips with double-dash

Imports are not discovered if used as a script `pnpm op -- node server.ts`.

People can call their `op` script whatever of course, would it be possible to blanket allow all script to have `--`?

```
"pnpm <any_script> -- node server.ts",
```

[...]

PR #1404, opened 2025-12-19
Script discovery tests

hilja:
Not 100% if `pkgScripts` is how you tell it to treat `program` as a script. These obviously fail.

https://github.com/webpro-nl/knip/issues/1399

### DIFF
--- a/packages/knip/test/util/get-inputs-from-scripts.test.ts
+++ b/packages/knip/test/util/get-inputs-from-scripts.test.ts
@@ -193,7 +193,7 @@ test('getInputsFromScripts (bun)', () => {
   t('bun ci', []);
 });
 
-test.only('getInputsFromScripts (pnpm)', () => {
+test('getInputsFromScripts (pnpm)', () => {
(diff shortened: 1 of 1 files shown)


hilja:
I think a whole category of tests are missing, the `pnpm program ...` is covered really well, but there's just the basic case for `pnpm script ...` and `node -r script`.

webpro:
Feel free to add more cases. Please keep in mind I like to keep it a bit concise/sensible/maintainable, there's no need to test every possible case out there (e.g. that list of `bun [command]` cases are a bit much imho).

Cases that won't be fixed in this round are still useful to know about and we can `t.skip`.

webpro:
Fixed a few cases. Forgot to add: Knip does not aim to implement all the features all the various tools might have. It's on the basis of best-effort. That said, feel free to dive in!

webpro:
:rocket: _This issue has been resolved in v5.83.1. See [Release 5.83.1](https://github.com/webpro-nl/knip/releases/tag/knip%405.83.1) for release notes._

PR #1415, opened 2025-12-23
Add Nitro plugin

lynnntropy:
Adds a plugin for [Nitro](https://nitro.build).


packages/knip/fixtures/plugins/nitro/knip.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ignore": [".nitro/types/*.d.ts"]
```
webpro: Ideally, users don't need to add this to their configuration.

We can add it to `entry` patterns in the plugin so the type definitions are added to the TS program.

packages/knip/src/plugins/nitro/index.ts
```diff
@@ -0,0 +1,64 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import type { Input } from '../../util/input.js';
+import { toDependency, toProductionEntry } from '../../util/input.js';
+import { join } from '../../util/path.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { NitroConfig } from './types.js';
+
+const title = 'Nitro';
+
+const enablers = ['nitropack'];
```
webpro: What is the difference between https://www.npmjs.com/package/nitro and https://www.npmjs.com/package/nitropack?

packages/knip/fixtures/plugins/nitro/.nitro/types/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "paths": {
+      "#imports": ["./nitro-imports"],
+      "~/*": ["../../server/*"],
+      "@/*": ["../../server/*"],
+      "~~/*": ["../../*"],
+      "@@/*": ["../../*"]
+    }
+  },
+  "include": ["./nitro.d.ts", "../../**/*", "../../server/**/*"]
+}
```
webpro, line 14: Do we need all the TS configuration to cover the plugin's functionality properly? Path aliases, for instance, are covered elsewhere already. Please minimize to the essentials, if at all possible.

webpro (commented):
Thanks, Lynn! Nice one, happy to merge. Added a few notes, but nothing major, I think.

packages/knip/src/plugins/nitro/index.ts
lynnntropy: `nitro` is the new package for Nitro v3, `nitropack` is the package for v2 (the current stable release).

Now that you mention it, do you think it'd make sense for this plugin to support both v2 and v3, even though v3 isn't stable yet?

packages/knip/fixtures/plugins/nitro/.nitro/types/tsconfig.json
lynnntropy: Done.

packages/knip/fixtures/plugins/nitro/knip.json
lynnntropy: Done.

packages/knip/src/plugins/nitro/index.ts
webpro: If you think the plugin implementation is largely the same for both the `nitro` and `nitropack` enablers, I think it makes sense to have just a single plugin

If they're quite different or if they should really have their own distinct title, then we should split it up I guess.

webpro: > though v3 isn't stable yet?

From a Knip perspective this is likely stable _enough_?

lynnntropy: Alright, sounds good to me. I'm pretty sure I've made all the changes necessary for the plugin to work with both v2 and v3. Should I also add a separate test + fixture for v3, or are we fine with just having the one?

packages/knip/fixtures/plugins/nitro/node_modules/nitropack/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "nitropack"
```
webpro: [Please add a `bin` field](https://github.com/webpro-nl/knip/blob/57825f443509fb2afa0444f7831ea67a798364f8/packages/knip/test/plugins/_template.test.ts#L15) so that it's clear this package contains the `nitro` binary which should also clean up the test counters.

lynnntropy: Done. 😊

webpro:
Thank you so much! _Very_ cool plugin ✨

webpro:

PR #1418, opened 2025-12-24
1289 knex plugin

### LINKED ISSUE #1289: 💡 Knex Plugin
### Suggest an idea for Knip

When installing the `knex` library, one needs to also install an appropriate database library (e.g., `pg` for PostgreSQL). Knex lists these libraries as optional peer dependencies in its own [`package.json`](https://github.com/knex/knex/blob/01b177c485d696f1b72858dee728ba143c4fad76/package.json#L79-L101). dynamically imports that library depending on the knex configuration.

Knip reports the installed database libraries as unused, which is a false positive.

Knex docs: https://knexjs.org/guide/

## Workaround
My current workaround is to ignore all database libraries in `knip.json`:
```json
{
  "ignoreDependencies": [
    "pg",
    "pg-native",
    "sqlite3",
    "better-sqlite3",
    "mysql",
    "mysql2",
    "oracledb",
    "tedious"
  ]
}
```

## Proposed Solution
I think it would be useful to create a knex plugin which parses all instantiations of knex and checks which clients are configured. For example, consider this knex configuration:
```javascript
const knex = require('knex')({
  client: 'mysql', // implies that `mysql` or `mysql2` will appear as a dependency in package.json
  connection: {
    host: '127.0.0.1',
    port: 3306,
    user: 'your_database_user',
    password: 'your_database_password',
    database: 'myapp_test',
  },
});
```

Knip should detect the instantiation of knex and check the value of the `client` property. If its value is `mysql` for example, Knip should consider `mysql` and `mysql2` dependencies to be used.

AlexanderKaran:
Closes #1289

Detects database drivers (pg, mysql2, sqlite3, etc.) from knexfile.js client configurations, preventing false positives when these packages are listed as dependencies but only dynamically imported by Knex.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -192,4 +192,6 @@ jobs:
 
       - name: Run Knip in ${{ matrix.project.repo }}
         working-directory: ${{ matrix.project.name }}
-        run: ${{ matrix.project.commands }}
+        run: |
+          set -x
+          ${{ matrix.project.commands }}
--- a/packages/docs/.gitignore
+++ b/packages/docs/.gitignore
@@ -1,5 +1,4 @@
 .astro/
-.cache/
 .DS_Store
 .env
 dist/
--- a/packages/docs/src/pages/og/[...route].ts
+++ b/packages/docs/src/pages/og/[...route].ts
@@ -7,7 +7,7 @@ const require = createRequire(import.meta.url);
 const sharp = require('sharp');
 
 const template = readFileSync(resolve('src/assets/og-template.svg'), 'utf-8');
-const cacheDir = resolve('.cache/og');
+const cacheDir = resolve('node_modules/.astro/og');
 
 const getPages = async () => {
   const data = import.meta.glob(['/src/content/**/*.{md,mdx}'], { eager: true });
--- a/packages/knip/fixtures/exports-value-refs-default/refs.ts
+++ b/packages/knip/fixtures/exports-value-refs-default/refs.ts
@@ -22,6 +22,16 @@ function setLogger(log: typeof logger): void {
   log;
 }
 
+export type UnusedTypeInUnusedExport = { id: number };
+
+export interface UnusedInterface {
+  api: UnusedTypeInUnusedExport;
+}
+
+export class UnusedClass {}
+
+export type UnusedTypeWithClass = typeof UnusedClass;
+
 const obj = {
   destrRefObj: 1,
 };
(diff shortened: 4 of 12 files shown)


packages/knip/src/plugins/knex/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  for (const key in envConfigs) {
+    const envConfig = envConfigs[key];
+    if (typeof envConfig === 'object' && envConfig !== null) {
+      inputs.push(...processKnexConfig(envConfig));
+    }
+  }
+  return inputs;
+};
+
+const resolveFromAST: ResolveFromAST = (sourceFile: ts.SourceFile) => {
```
webpro: If I look at the fixtures, we can do with only the `resolveConfig`, and remove `resolveFromAST`.

packages/knip/src/plugins/knex/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (config.seeds?.directory) {
+    const dirs = Array.isArray(config.seeds.directory) ? config.seeds.directory : [config.seeds.directory];
+    inputs.push(...dirs.map(dir => toEntry(`${dir}/*.{js,ts}`)));
+  }
+  return inputs;
+};
+
+const resolveConfig: ResolveConfig<KnexConfig | Record<string, KnexConfig>> = config => {
+  if ('client' in config && config.client) {
+    return processKnexConfig(config as KnexConfig);
```
webpro: Perhaps we can "cast" configs to an array and keep only the block below to dedupe logic and iterate once.

packages/knip/src/plugins/knex/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    const dirs = Array.isArray(config.migrations.directory)
+      ? config.migrations.directory
+      : [config.migrations.directory];
+    inputs.push(...dirs.map(dir => toEntry(`${dir}/*.{js,ts}`)));
+  }
+  if (config.seeds?.directory) {
+    const dirs = Array.isArray(config.seeds.directory) ? config.seeds.directory : [config.seeds.directory];
+    inputs.push(...dirs.map(dir => toEntry(`${dir}/*.{js,ts}`)));
+  }
+  return inputs;
```
webpro, line 33: This helper is great. Would be great to actually have a migration and a seed file just to show they won't end up unused.

webpro:
Thank you! You're on a streak 🔥

webpro:

PR #1419, opened 2025-12-24
Fix Svelte plugin to detect svelte.config.ts

### LINKED ISSUE #1416: 🐛 Svelte plugin doesn't detect svelte.config.ts (only .js)
### Prerequisites

### Description

The Svelte plugin hardcodes `svelte.config.js` in its entry patterns:

```typescript
const entry = ['svelte.config.js', ...viteConfig];
```

However, SvelteKit projects can use `svelte.config.ts` (TypeScript config files). When using `.ts` extension, knip reports `@sveltejs/adapter-*` packages as unused because it never parses the config file.

### Expected Behavior

The Svelte plugin should detect `svelte.config.{js,ts}` like most other plugins do.

### Workaround

Override the svelte plugin config in knip.json:

```json
{
  "svelte": {
    "config": ["svelte.config.{js,ts}"]
  }
}
```

### Environment

- knip: 5.76.3
- SvelteKit: 2.49.2
- Node: 25.2.1

### Suggested Fix

In `packages/knip/src/plugins/svelte/index.ts`, change:

```typescript
const entry = ['svelte.config.js', ...viteConfig];
```

To:

```typescript
const entry = ['svelte.config.{js,ts,mjs,mts}', ...viteConfig];
```

This follows the pattern used by other plugins (vite, vitest, etc.) that support both JS and TS config files.

Pipboyguy:
Fixes #1416

Changed `svelte.config.js` to `svelte.config.{js,mjs,ts,cjs,mts,cts}` in the Svelte plugin entry pattern.


packages/knip/src/plugins/svelte/index.ts
```diff
@@ -12,7 +12,7 @@ const enablers = ['svelte'];
 
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
-const entry = ['svelte.config.js', ...viteConfig];
+const entry = ['svelte.config.{js,mjs,ts,cjs,mts,cts}', ...viteConfig];
```
webpro, line 15: This whole PR could be just this change.

Where is it actually documented or codified that Svelte reads anything else than `svelte.config.js`?

cylewaitforit: In relation to "official svelte docs" I found this 

```md
Besides inline options in Vite config, vite-plugin-svelte will also automatically resolve options from a Svelte config file if one exists. The default search paths are:

    svelte.config.js
    svelte.config.mjs
    svelte.config.ts
    svelte.config.mts

```


https://github.com/sveltejs/vite-plugin-svelte/blob/84294a99c299a1e15ee99be4bfa9f2d8a64a9262/docs/config.md#config-file-resolving

webpro: Thanks! That says a lot. A few lines below that there is:

> Try to stick with the `.js` extension whenever possible.

Which makes sense because Svelte itself won't pick up anything else?

Anyway, I'm going to close this PR. Sounds like a bit of an edge case, might be better off in a separate plugin.

webpro:
Closing this one. Also see comment https://github.com/webpro-nl/knip/pull/1419#discussion_r2659526672

PR #1424, opened 2025-12-28
Fix gitignore ancestor detection for git worktrees

altendky:
Git worktrees use a `.git` file (containing a pointer to the actual git
directory) instead of a `.git` directory. The gitignore ancestor traversal
was only checking for directories, causing it to incorrectly continue
searching beyond the worktree root into the parent repository.

This adds an `isGitRoot()` helper that checks for both `.git` directory
and `.git` file, fixing the boundary detection for worktrees.

### DIFF
--- a/packages/knip/test/util/find-and-parse-gitignores.test.ts
+++ b/packages/knip/test/util/find-and-parse-gitignores.test.ts
@@ -1,13 +1,9 @@
 import assert from 'node:assert/strict';
-import { copyFileSync } from 'node:fs';
+import fs from 'node:fs/promises';
 import { join } from 'node:path';
 import test from 'node:test';
 import { findAndParseGitignores } from '../../src/util/glob-core.js';
 import { resolve } from '../helpers/resolve.js';
-
-// Create the .git file for worktree tests (git won't track files named .git)
-const worktreeRoot = resolve('fixtures/glob-worktree/root');
-copyFileSync(join(worktreeRoot, 'dot-git'), join(worktreeRoot, '.git'));
 
 test('findAndParseGitignores', async () => {
   const cwd = resolve('fixtures/glob');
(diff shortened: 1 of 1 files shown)


altendky:
following up on my spam in https://github.com/webpro-nl/knip/issues/990

altendky:
i fixed some of the basic test failures.  i forgot to make sure git included the `.git` file for the tests, which it did not.  i'm not sure about some other failures that had occurred.

packages/knip/test/util/find-and-parse-gitignores.test.ts
```diff
@@ -1,8 +1,14 @@
 import assert from 'node:assert/strict';
+import { copyFileSync } from 'node:fs';
+import { join } from 'node:path';
 import test from 'node:test';
 import { findAndParseGitignores } from '../../src/util/glob-core.js';
 import { resolve } from '../helpers/resolve.js';
 
+// Create the .git file for worktree tests (git won't track files named .git)
+const worktreeRoot = resolve('fixtures/glob-worktree/root');
+copyFileSync(join(worktreeRoot, 'dot-git'), join(worktreeRoot, '.git'));
```
webpro: Probably good to clean up as well, also see https://github.com/webpro-nl/knip/blob/af0ebaa98537be9c95f32c2c3b58cb506922d9a3/packages/knip/test/plugins/lefthook-v1.test.ts#L30

webpro (commented):
Thanks! Hopefully we can get the test issues fixed.

packages/knip/test/util/find-and-parse-gitignores.test.ts
altendky: https://github.com/webpro-nl/knip/pull/1424/changes/7998413d07e353fed6d10ba59d6781dc4afcd379 made it not-global as well

webpro: Thanks! What I meant is to clean up after the test ran, but it does not seem to matter, so I guess we're good here.

webpro:
Thanks Kyle! Way to go 😉

webpro:

PR #1430, opened 2026-01-01
Added TanStack Router Plugin

AlexanderKaran:
Added a plugin for TanStack Router and React. I was going to add one for TanStack Start but it is based on Router which should be its own plugin and then will simplfy the Start one.

- Set up basic TanStack Router with React 
- Added other plugins into enablers, which should be find as they use the same file based strucuture:

```js
const enablers = [
  '@tanstack/react-router',
  '@tanstack/solid-router',
  '@tanstack/vue-router',
  '@tanstack/svelte-router',
  '@tanstack/router-cli',
  '@tanstack/router-plugin',
];
```

Let me know if I missed anything

### DIFF
--- a/packages/knip/src/plugins/tanstack-router/index.ts
+++ b/packages/knip/src/plugins/tanstack-router/index.ts
@@ -16,24 +16,13 @@ const enablers = [
 
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
-const config = ['tsr.config.json', 'vite.config.{js,ts,cjs,mjs}'];
-
 const entry = ['src/routeTree.gen.{ts,js}'];
-
-const production = [
-  'src/routes/__root.{tsx,jsx,ts,js}',
-  'src/routes/**/*.{tsx,jsx,ts,js}',
-  '!src/routes/**/-*',
-  '!src/routes/**/-*/**',
-];
 
 const plugin: Plugin = {
   title,
   enablers,
   isEnabled,
-  config,
   entry,
-  production,
 };
 
 export default plugin;


packages/knip/src/plugins/tanstack-router/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const config = ['tsr.config.json', 'vite.config.{js,ts,cjs,mjs}'];
+
+const entry = ['src/routeTree.gen.{ts,js}'];
+
+const production = [
+  'src/routes/__root.{tsx,jsx,ts,js}',
+  'src/routes/**/*.{tsx,jsx,ts,js}',
+  '!src/routes/**/-*',
+  '!src/routes/**/-*/**',
```
webpro: 👌

packages/knip/schema.json
```diff
@@ -741,6 +741,14 @@
...
           "$ref": "#/definitions/plugin"
         },
+        "tanstack-router": {
+          "title": "tanstack-router plugin configuration (https://knip.dev/reference/plugins/tanstack-router)",
+          "$ref": "#/definitions/plugin"
+        },
+        "tanstack-start": {
+          "title": "tanstack-start plugin configuration (https://knip.dev/reference/plugins/tanstack-start)",
+          "$ref": "#/definitions/plugin"
+        },
```
webpro, line 751:
```suggestion
```

webpro (commented):
Nice one! Looks clean & good to me.

packages/knip/src/plugins/tanstack-router/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  '@tanstack/solid-router',
+  '@tanstack/vue-router',
+  '@tanstack/svelte-router',
+  '@tanstack/router-cli',
+  '@tanstack/router-plugin',
+];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['tsr.config.json', 'vite.config.{js,ts,cjs,mjs}'];
```
webpro: Oh, one more thing: `config` entries do nothing if there's no `resolveConfig` (they are also used as `entry` though) and should be moved to `entry` (in which case `.json` entries can be omitted).

webpro: Maybe just leave it to the Vite plugin if that one's enabled? Otherwise pull in the Vite plugin's `resolveConfig`.

packages/knip/src/plugins/tanstack-router/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const entry = ['src/routeTree.gen.{ts,js}'];
```
webpro, line 19: If we would just have this one as a `production` (as that's what it looks like), can we omit the `src/routes/*` patterns altogether? They're all imported from the generated file.

packages/knip/src/plugins/tanstack-router/index.ts
AlexanderKaran: TanStack router can be used by other bundlers to as well just Vite is the most common, will fix this up as that should be left to Vite

packages/knip/src/plugins/tanstack-router/index.ts
AlexanderKaran: We can do this. I left it in incase the file was not kept up to date but it should be as part of the best practices of the router.

AlexanderKaran:
All fixed @webpro

webpro:
Thanks a bunch! Love how simple yet effective this plugin seems to be (kudos to TanStack too for that obviously).

AlexanderKaran:
> Thanks a bunch! Love how simple yet effective this plugin seems to be (kudos to TanStack too for that obviously).

Yeah TanStack router is amazing. Really fallen in love with it.

webpro:

Issue #1440, opened 2026-01-03
💡 Plugin for astro-og-canvas?


JoshuaKGoldberg:
### Suggest an idea for Knip

Coming over from https://github.com/flint-fyi/flint/pull/1229 -> https://github.com/delucis/astro-og-canvas/issues/19 -> https://github.com/delucis/astro-og-canvas/tree/4f0abfbb5dcae4a28095f97ebd53ea030a618c00/packages/astro-og-canvas#installation: the [`astro-og-canvas`](https://github.com/delucis/astro-og-canvas) package requires adding a direct dependency on `canvaskit-wasm` when using pnpm. That results in `canvaskit-wasm` being flagged by Knip:

```plaintext
> knip

Unused dependencies (1)
canvaskit-wasm  packages/site/package.json:32:4
```

Would you accept a PR to add a plugin for `astro-og-canvas` that makes `canvaskit-wasm` as a reasonable dependency?

webpro:
Yes, that plugin would be very welcome! It's odd though, seeing `canvaskit-wasm` being listed as a dependency of `astro-og-canvas` - hopefully this will be sorted out eventually (by pnpm?) and we won't need this workaround here.

> reasonable dependency

And [optional](https://knip.dev/writing-a-plugin/inputs#options) :)

JoshuaKGoldberg:
I took a stab at it in #1445. But there's a test failure I don't understand 😕. I tried!

joealden:
This can be closed FYI

PR #1445, opened 2026-01-06
Add astro-og-canvas plugin

JoshuaKGoldberg:
Adds a plugin for [`astro-og-canvas`](https://github.com/delucis/astro-og-canvas). As noted in https://github.com/webpro-nl/knip/issues/1440#issuecomment-3707952725, it sets `canvaskit-wasm` as an optional dependency.

### DIFF
--- a/packages/knip/src/plugins/astro-og-canvas/types.ts
+++ b/packages/knip/src/plugins/astro-og-canvas/types.ts
@@ -1,4 +1,0 @@
-export type PluginConfig = {
-  plugins?: string[];
-  entryPathsOrPatterns?: string[];
-};


JoshuaKGoldberg:
🤔 I poked around for a few minutes but couldn't figure this out. I don't see anything wrong with this PR's code?

```plaintext
not ok 189 - Resolve imports in separate workspaces without entry file
  ---
  duration_ms: 218.728072
  type: 'test'
  location: '/home/runner/work/knip/knip/packages/knip/test/workspaces-cross-reference.test.ts:1:364'
  failureType: 'testCodeFailure'
  error: "Cannot read properties of undefined (reading 'unused')"
  code: 'ERR_TEST_FAILURE'
  name: 'TypeError'
  stack: |-
    TestContext.<anonymous> (/home/runner/work/knip/knip/packages/knip/test/workspaces-cross-reference.test.ts:18:50)
    process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    async Test.run (node:internal/test_runner/test:1054:7)
    async startSubtestAfterBootstrap (node:internal/test_runner/harness:296:3)
  ...
```

webpro:
Yeah sorry that's just a flaky test I think. Gave it a nudge!

packages/knip/src/plugins/astro-og-canvas/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const entry: string[] = [];
+
+const production: string[] = [];
+
+const resolveConfig: ResolveConfig<PluginConfig> = async config => {
+  const inputs = config?.plugins ?? [];
+  return [
+    ...inputs.map(id => toDeferResolve(id)),
+    toDependency('canvaskit-wasm', { optional: true }),
```
webpro: For this `resolveConfig` function to run we need a `config` entry that will then be loaded, and default-exported value passed in here as `config`. Does that apply?

If not, we can implement a [resolve](https://github.com/webpro-nl/knip/blob/b9dc80e355f490cea5ee7fd153c8821049722f8a/packages/knip/src/types/config.ts#L166) instead without a config file (which I now realized is undocumented.. sorry).

packages/knip/src/plugins/astro-og-canvas/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const enablers = ['astro-og-canvas'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = [];
+
+const entry: string[] = [];
+
+const production: string[] = [];
```
webpro: The items that end up empty can be removed 😇

packages/knip/test/plugins/astro-og-canvas.test.ts
```diff
@@ -0,0 +1,21 @@
...
+const cwd = resolve('fixtures/plugins/astro-og-canvas');
+
+test('Find dependencies with the astro-og-canvas plugin', async () => {
+
+  const options = await createOptions({ cwd });
+  const { counters } = await main(options);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    dependencies: 1,
```
webpro: So this should end up with no dependencies unused which isn't the case yet, because `resolveConfig` is not executed at all.

packages/knip/src/plugins/astro-og-canvas/index.ts
JoshuaKGoldberg: If I'm understanding them, no, I don't think there's a `config` entry. Would a `resolve` be something like this?

```ts
const resolve: Resolve = async () => {
  return [toDependency('astro-og-canvas')];
};
```

webpro: Yes exactly! This is pretty uncommon, most plugins exist because of the config files they parse :)

JoshuaKGoldberg: Ok cool! I added one and removed the `dependencies: 1`. But tests are failing on that 😕

webpro: Sorry I spoke too fast. This is all we need:

```ts
const resolve: Resolve = () => [toDependency('canvaskit-wasm', { optional: true })];
```

No `resolveConfig` as `astro-og-canvas` does not have a config file.

Then likely `astro-og-canvas` itself is still unused, perhaps just add `index.js` that imports it, just to reference it (as we would in some Astro page or something).

JoshuaKGoldberg: Oh, I see, that makes sense given the lack of config file. 👍

packages/knip/src/plugins/astro-og-canvas/index.ts
```diff
@@ -0,0 +1,24 @@
...
+// https://github.com/delucis/astro-og-canvas
+
+const title = 'astro-og-canvas';
+
+const enablers = ['astro-og-canvas'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const resolve: Resolve = async () => {
+  return [toDependency('astro-og-canvas')];
```
webpro:
```suggestion
  return [toDependency('canvaskit-wasm', { optional: true })];
```

packages/knip/test/plugins/astro-og-canvas.test.ts
```diff
@@ -0,0 +1,19 @@
...
+const cwd = resolve('fixtures/plugins/astro-og-canvas');
+
+test('Find dependencies with the astro-og-canvas plugin', async () => {
+  const options = await createOptions({ cwd });
+  const { counters } = await main(options);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    processed: 0,
+    total: 0,
```
webpro:
```suggestion
    processed: 1,
    total: 1,
```

webpro (commented):
Almost there 💖✨

webpro:
The lights aren't green yet because there is an unused file lying around. Let me know whether you'd like to wrap this up or I'm happy to do so otherwise!

JoshuaKGoldberg:
Ah, sorry, I'd meant to check back in on this and got distracted. Thanks for the prod 🙂

webpro:
Nah sorry sometimes I'm all over the place. Thank you! This is neat.

webpro:
:rocket: _This pull request is included in v5.80.2. See [Release 5.80.2](https://github.com/webpro-nl/knip/releases/tag/knip%405.80.2) for release notes._

PR #1447, opened 2026-01-06
Replace fast-glob with tinyglobby

### LINKED ISSUE #1446: 🧩 Replace `fast-glob` with `tinyglobby`

I will make a PR

gameroman:
Closes #1446

### DIFF
--- a/packages/docs/src/content/docs/reference/cli.md
+++ b/packages/docs/src/content/docs/reference/cli.md
@@ -23,6 +23,34 @@ Shortcut: `-n`
 Don't show dynamic progress updates. Progress is automatically disabled in CI
 environments.
 
+### `--config [file]`
+
+Use an alternative path for the configuration file. Default locations:
+
+- `knip.json`
+- `knip.jsonc`
+- `.knip.json`
+- `.knip.jsonc`
+- `knip.js`
+- `knip.ts`
+- `package.json#knip`
+
+Shortcut: `-c`
+
+### `--use-tsconfig-files`
+
+Use `tsconfig.json` to define project files (override `project` patterns).
+
+### `--tsConfig [file]`
+
+Shortcut: `-t`
+
+Use an alternative path for the TypeScript configuration file.
+
+Using `-t jsconfig.json` is also supported.
+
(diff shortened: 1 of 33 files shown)


webpro:
Thanks for bringing this up again, and the efforts already put in. I'm going to change the title to make it clear that we should really only merge if we can get this right.

gameroman:
looks like there are 4 places where the `firstGlob` function is used

https://github.com/webpro-nl/knip/blob/5b98269b0ce28890a8a16559846188dd2016bf25/packages/knip/src/plugins/yarn/index.ts#L10

https://github.com/webpro-nl/knip/blob/5b98269b0ce28890a8a16559846188dd2016bf25/packages/knip/src/plugins/taskfile/index.ts#L21

https://github.com/webpro-nl/knip/blob/5b98269b0ce28890a8a16559846188dd2016bf25/packages/knip/src/plugins/github-actions/index.ts#L15

https://github.com/webpro-nl/knip/blob/5b98269b0ce28890a8a16559846188dd2016bf25/packages/knip/src/plugins/pnpm/index.ts#L11

gameroman:
@webpro I removed 2 of the usages and will see if I can deal with the other 2

gameroman:
removed another one

gameroman:
removed the last one

silverwind:
Could use [fsPromises.glob](https://nodejs.org/api/fs.html#fspromisesglobpattern-options) as well, given that it's available in Node 22 and greater which matches the minimum version of this module.

webpro:
Knip requires at least Node.js v18, tinyglobby v12. And I don't expect Knip to reach the `fs.glob` threshold too soon, as it might be helpful in legacy projects trying to keep up. That said, using built-ins is usually a good idea once we can (but tbh haven't looked into differences here recently).

silverwind:
Ah sorry, I was looking at the [wrong package.json](https://github.com/webpro-nl/knip/blob/ecf5e34b51f3817dabcf50ca6705141aea3f56e5/package.json#L51), carry on :D

gameroman:
#1462

PR #1449, opened 2026-01-07
Improved TanStack Router Configs

AlexanderKaran:
After testing the TanStack Router plugin on a few different size codes bases in CI and others there are times where we got false positive warnings with exports from the route files. Followed examples from other router set ups in Knip and now treat route files as production entries to skip export analysis. Also makes the plugin more robust to cases where routeTree.gen.ts doesn't exist yet (fresh clones, clean builds, CI).

### DIFF
--- a/packages/knip/src/plugins/tanstack-router/index.ts
+++ b/packages/knip/src/plugins/tanstack-router/index.ts
@@ -1,6 +1,6 @@
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
 import { toProductionEntry } from '../../util/input.js';
-import { join, toAbsolute } from '../../util/path.js';
+import { toAbsolute } from '../../util/path.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { TanStackRouterConfig } from './types.js';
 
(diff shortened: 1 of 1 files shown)


packages/knip/src/plugins/tanstack-router/index.ts
```diff
@@ -16,13 +19,36 @@ const enablers = [
...
+  const { configFileDir } = options;
+
+  const routesDirectory = localConfig.routesDirectory ?? './src/routes';
+  const generatedRouteTree = localConfig.generatedRouteTree ?? './src/routeTree.gen.ts';
+
+  const routesDir = toAbsolute(routesDirectory, configFileDir);
+  const routeTreePath = toAbsolute(generatedRouteTree, configFileDir);
+
+  const resolved = [
+    toEntry(routeTreePath),
```
webpro: Then I guess `src/routeTree.gen.{ts,js}` is a production entry as well.

packages/knip/src/plugins/tanstack-router/index.ts
```diff
@@ -16,13 +19,36 @@ const enablers = [
...
+const config = ['tsr.config.json'];
+
 const entry = ['src/routeTree.gen.{ts,js}'];
+const production = ['src/routes/**/*.{ts,tsx,js,jsx}'];
+
+const resolveConfig: ResolveConfig<TanStackRouterConfig> = async (localConfig, options) => {
+  const { configFileDir } = options;
+
+  const routesDirectory = localConfig.routesDirectory ?? './src/routes';
+  const generatedRouteTree = localConfig.generatedRouteTree ?? './src/routeTree.gen.ts';
```
webpro, line 29: The fallback/default could point to `src/routeTree.gen.{ts,js}` that might end up in `production: []`? See other comment.

webpro (commented):
Thanks! I wonder if we should pick one strategy though: either only use the generated tree file, or routes glob. Unless you think results are better/more predictable by using both?

AlexanderKaran:
@webpro Going to close this PR. After more digging turns our the `routeTree.gen.{ts,js}` was not in Git and that was the issue. TanStack position (was not always the case) is this file should be in Git. There just older projects.

webpro:
Ah gotcha. Yet maybe `routeTree.gen` could be a production entry?

AlexanderKaran:
@webpro moved it to `production` and added support if they change the location in the tanstack config. So better support for TanStack but does not reward incorrect setup. :)

webpro:
Oohh nice 👌 Thanks!

webpro:
:rocket: _This pull request is included in v5.80.1. See [Release 5.80.1](https://github.com/webpro-nl/knip/releases/tag/knip%405.80.1) for release notes._

PR #1454, opened 2026-01-08
Replace glob with fs helpers

gameroman:
Step 1 of #1447

### DIFF
--- a/packages/knip/src/plugins/github-actions/index.ts
+++ b/packages/knip/src/plugins/github-actions/index.ts
@@ -1,6 +1,5 @@
-import { readdir } from 'node:fs/promises';
 import type { IsPluginEnabled, Plugin, PluginOptions, ResolveConfig } from '../../types/config.js';
-import { isDirectory } from '../../util/fs.js';
+import { hasFilesWithExtensions } from '../../util/fs.js';
 import { type Input, isDeferResolveEntry, toEntry } from '../../util/input.js';
 import { findByKeyDeep } from '../../util/object.js';
 import { join, relative } from '../../util/path.js';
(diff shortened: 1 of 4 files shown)


packages/knip/src/plugins/github-actions/index.ts
```diff
@@ -11,8 +12,17 @@ const title = 'GitHub Actions';
...
 
-const isEnabled: IsPluginEnabled = async ({ cwd }) =>
-  Boolean(await _firstGlob({ cwd, patterns: ['.github/workflows/*.{yml,yaml}'] }));
+const isEnabled: IsPluginEnabled = async ({ cwd }) => {
+  const workflowsDir = join(cwd, '.github/workflows');
+  if (!isDirectory(workflowsDir)) return false;
+
+  try {
+    const files = await readdir(workflowsDir);
+    return files.some(file => file.endsWith('.yml') || file.endsWith('.yaml'));
```
webpro: Plugins shouldn't need this complexity. It isn't much, and it's generally not an issue if a piece of logic in one plugin is an outlier for some reason, yet in this case this should somehow be part of the `fs` or `glob` helpers. I think that keeps things organized in a centralized place (I could imagine we might want to use this more often). Within the overarching story of the migration to tinyglobby it feels a bit out of place (although I'm fine with moving things around in the process a bit and wrapping up properly after the fact).

gameroman: Probably fs helper as there is no glob logic

gameroman: Done

webpro:
Nice work, thanks!

webpro:

Issue #1441, opened 2026-01-05
💡 `--workspace` Configuration: Expected Support for Selecting Multiple Workspaces


fightZy:
### Suggest an idea for Knip

 Currently, the `--workspace` option only supports a single directory. There are dozens of workspaces in my project, and I want to check only the specified ones in certain pipelines. However, I currently have to run the command multiple times, such as 
```
pnpm knip --workspace A & pnpm knip --workspace B & pnpm knip --workspace C
```
This actually involves redundant analysis logic (such as build graphs). I hope to have a multi-workspace selection and filtering command similar to [turbo](https://turborepo.com/docs/reference/run#--filter-string) and [pnpm](https://pnpm.io/filtering) to optimize the program.
For example
```
pnpm knip --workspace A,B,C
```

webpro:
It's a good idea, I'd welcome a pull request for this if you're up to it.

fightZy:
My pleasure. I will attempt to start working on it and submit a pull request.

webpro:
Closing this as it has been merged and released. Thanks again, glad this is in 🙏

PR #1455, opened 2026-01-09
Expand workspace filter

fightZy:
Feature https://github.com/webpro-nl/knip/issues/1441

Filter the workspace by package names, directory paths, globs and Git ranges.

Key Changes
- Multiple Selectors: Support multiple `--workspace` (`-W`) arguments.
- Negated Patterns: Ability to exclude workspaces using `!`.
- Glob Support: Match package names and directories using wildcards.
- Git Integration: Select workspaces based on changed files in a Git range (e.g., `[main...HEAD]`).

### DIFF
--- a/packages/knip/src/ConfigurationChief.ts
+++ b/packages/knip/src/ConfigurationChief.ts
@@ -27,6 +27,7 @@ import { toRegexOrString } from './util/regex.js';
 import { ELLIPSIS } from './util/string.js';
 import { byPathDepth } from './util/workspace.js';
 import { selectWorkspaces } from './util/workspace-selectors.js';
+import { createWorkspaceFilePathFilter, type WorkspaceFilePathFilter } from './util/workspace-file-filter.js';
 
 const defaultBaseFilenamePattern = '{index,cli,main}';
 
@@ -91,6 +92,7 @@ export class ConfigurationChief {
   config: Configuration;
   workspace: string | string[] | undefined;
   selectedWorkspaces: string[] | undefined;
+  workspaceFilePathFilter: WorkspaceFilePathFilter | undefined;
 
   workspaces: string[];
   ignoredWorkspacePatterns: string[] = [];
@@ -193,6 +195,12 @@ export class ConfigurationChief {
     this.workspaceGraph = createWorkspaceGraph(this.cwd, this.availableWorkspaceNames, wsPkgNames, packages);
 
     this.selectedWorkspaces = this.getSelectedWorkspaces();
+
+    this.workspaceFilePathFilter = createWorkspaceFilePathFilter(
+      this.cwd,
+      this.selectedWorkspaces,
+      this.availableWorkspaceNames
+    );
 
     this.includedWorkspaces = this.getIncludedWorkspaces();
 
--- a/packages/knip/src/IssueCollector.ts
+++ b/packages/knip/src/IssueCollector.ts
@@ -5,7 +5,7 @@ import { partition } from './util/array.js';
 import type { MainOptions } from './util/create-options.js';
 import { initCounters, initIssues } from './util/issue-initializers.js';
 import { relative } from './util/path.js';
-import { createWorkspaceFilePathFilter } from './util/workspace-file-filter.js';
+import type { WorkspaceFilePathFilter } from './util/workspace-file-filter.js';
 
 const createMatcher = (patterns: Set<string>) => {
   const [negated, positive] = partition(patterns, p => p[0] === '!');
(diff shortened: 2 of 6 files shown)


packages/knip/src/ConfigurationChief.ts
```diff
@@ -258,27 +262,25 @@ export class ConfigurationChief {
   }
 
   private getIncludedWorkspaces() {
-    if (this.workspace) {
-      const dir = path.resolve(this.cwd, this.workspace);
-      if (!isDirectory(dir)) throw new ConfigurationError('Workspace is not a directory');
-      if (!isFile(dir, 'package.json')) throw new ConfigurationError('Unable to find package.json in workspace');
```
webpro, line 264: Are we giving up on this? If someone does `knip -W ./not-found` or if there's no `package.json` then Knip should err.

packages/knip/src/IssueCollector.ts
```diff
@@ -39,11 +40,16 @@ export class IssueCollector {
...
-    this.filter = options.workspace ? join(options.cwd, options.workspace) : undefined;
     this.isMatch = () => false;
     this.isFileMatch = () => false;
   }
 
+  setWorkspaceFilter(selectedWorkspaces: string[], availableWorkspaceNames: string[]) {
+    const isSelectedFilePath = createWorkspaceFilePathFilter(this.cwd, selectedWorkspaces, availableWorkspaceNames);
+    if (!isSelectedFilePath) return;
+    this.workspaceFilter = (filePath: string) => !isSelectedFilePath(filePath);
+  }
```
webpro, line 50: How about creating the filter earlier on, e.g. in `createOptions`? Then we can set in the `constructor` directly. Perhaps same for as much as possible for the logic happening in `ConfigurationChief`.

fightZy:
> Thanks for the pull request! This is a valuable feature for large projects indeed. Happy to review, but given the size of the changeset and the fact I'm not sure we should include the Git-based range filter (at least not yet), could you please remove that part?

Regarding Git range filtering, my consideration is that it can conveniently check and report the scope of changes in the CI environment of monorepo projects, thus shortening the execution time.

packages/knip/src/ConfigurationChief.ts
fightZy: https://github.com/webpro-nl/knip/pull/1455/changes#diff-3c4f4094662c809c4485f88ceda6a292f04134e4a5f9f5c8cae93228c387e7c9R169-R195

Yes, an error will be reported when parsing the path.

webpro:
> > Thanks for the pull request! This is a valuable feature for large projects indeed. Happy to review, but given the size of the changeset and the fact I'm not sure we should include the Git-based range filter (at least not yet), could you please remove that part?
> 
> Regarding Git range filtering, my consideration is that it can conveniently check and report the scope of changes in the CI environment of monorepo projects, thus shortening the execution time.

Agreed. Yet once Knip lands support for targeting multiple workspaces at once, you can still achieve that (using e.g. a shell script).

This type of use case is also why Knip supports [dynamic configs](https://knip.dev/reference/dynamic-configuration). Although that won't fly for CLI arguments... Perhaps it makes sense to add a new `workspace: string[]` to the [configuration options](https://knip.dev/reference/configuration#workspaces).

Next to the idea that smaller PRs are generally better than larger ones, for both users and for me as a maintainer, not landing everything at once has multiple benefits:

Does the feature work well..

- in other projects/environments?
- in tandem with other features and/or their user configurations?
- in a codebase without Git — or with another CVS, although I must admit I haven't tested Knip much myself in a not-Git VCS, it's just that bringing in a new feature depending on Git reminds me of this

Once we have a better understanding of this, I'm happy to discuss next steps.

fightZy:
> Agreed. Yet once Knip lands support for targeting multiple workspaces at once, you can still achieve that (using e.g. a shell script).
> 
> This type of use case is also why Knip supports [dynamic configs](https://knip.dev/reference/dynamic-configuration). Although that won't fly for CLI arguments... Perhaps it makes sense to add a new `workspace: string[]` to the [configuration options](https://knip.dev/reference/configuration#workspaces).
> 
> Next to the idea that smaller PRs are generally better than larger ones, for both users and for me as a maintainer, not landing everything at once has multiple benefits:
> 
> Does the feature work well..
> 
> * in other projects/environments?
> * in tandem with other features and/or their user configurations?
> * in a codebase without Git — or with another CVS, although I must admit I haven't tested Knip much myself in a not-Git VCS, it's just that bringing in a new feature depending on Git reminds me of this
> 
> Once we have a better understanding of this, I'm happy to discuss next steps.

Your points make total sense—fully agree!

packages/knip/src/IssueCollector.ts
fightZy: Optimized https://github.com/webpro-nl/knip/pull/1455/commits/2dd85125da2bc0fa6a06060d74d2af58a6b102ee

webpro:
Thanks, this is a great PR! Much appreciated. While reviewing I saw few things I wanted to refactor and figured it would be an unnecessary back & forth to ask you. No functional/API changes, just internals. Hope you don't mind! It's all in c5e66db.

webpro:

PR #1457, opened 2026-01-09
Fix manual enabled compilers

digitsum:
Fixes the bug that manually enabled compilers as described [here](https://knip.dev/features/compilers#built-in-compilers) are not actually enabled. The config gets lost in the transformation. See provided fixture for a reproduction example.

During analysis I already found the culprit code location. That's why I directly opened a pull request to fix it. Please let me know if you prefer a different solution or if I should open an issue first.

### DIFF
--- a/packages/knip/fixtures/plugins/vue-webpack/knip.ts
+++ b/packages/knip/fixtures/plugins/vue-webpack/knip.ts
@@ -1,6 +1,3 @@
 export default {
   entry: ['parent.vue'],
-  compilers: {
-    vue: true,
-  },
 };


packages/knip/fixtures/plugins/vue-webpack/knip.ts
```diff
@@ -1,6 +1,3 @@
 export default {
   entry: ['parent.vue'],
-  compilers: {
-    vue: true,
-  },
```
digitsum, line 5: Here the `vue` compiler is already automatically activated via the `vue` dependency in package.json. So the manual setting is not needed here and this also explains why this test is green.
That's why I created a new separate fixture. I removed the manual setting here to avoid confusion. Please let me know if this is beyond the scope of this pull request.

webpro:
Dang you are right! Thank you very much for the fix ✨ 🙏

webpro:

PR #1462, opened 2026-01-12
Replace `fast-glob` with `tinyglobby`

### LINKED ISSUE #768: 💡 Replace `fast-glob` by `tinyglobby`
### Suggest an idea for this project

Most of `knip` deps come from `fast-glob`.

https://npmgraph.js.org/?q=knip

https://bundlephobia.com/package/fast-glob vs https://bundlephobia.com/package/tinyglobby

repo: https://github.com/SuperchupuDev/tinyglobby

see: https://github.com/es-tooling/module-replacements/blob/main/docs/modules/glob.md

### LINKED ISSUE #1446: 🧩 Replace `fast-glob` with `tinyglobby`

gameroman:
Replaces `fast-glob` with `tinyglobby`
Closes #768
Closes #1446


gameroman:
@superchupudev maybe you know better what could be causing the issue?

SuperchupuDev:
hi! what is the function being called with? you can find out via the `debug` option or the `TINYGLOBBY_DEBUG` env var then running the tests

gameroman:
> hi! what is the function being called with? you can find out via the `debug` option or the `TINYGLOBBY_DEBUG` env var then running the tests

Should I enable it in CI temporary or try running it locally?

gameroman:
@SuperchupuDev, enabled it in CI

webpro:
@gameroman If you're using VS Code or WebStorm, there are some tips in https://github.com/webpro-nl/knip/blob/main/.github/DEVELOPMENT.md. In this case, you could open a failing test and then set a breakpoint to see for instance the `patterns` that come in and the resulting file paths. Here's a quick example:

- Open `subpath-imports-outdir` test
- Open `glob.ts` and set breakpoint
- Run "Debug test with Bun" (or tsx/Node) launch config (`F5`) from inside the test file
- See `patterns` come in, maybe it has something to do with a pattern being an absolute path, or missing `cwd`:

Screenshot 2026-01-13 at 6 46 28 AM

Does that help?

gameroman:
To be honest, I don't really know what to do with that debug information though

I will try again tomorrow

webpro:
My first guess was that, given the input arguments are the same, apparently fast-glob handles this differently than tinyglobby, and so their output is not the same (it's not a drop-in replacement).

packages/knip/src/util/glob.ts
```diff
@@ -53,8 +53,15 @@ const defaultGlob = async ({ cwd, dir = cwd, patterns, gitignore = true, label }
...
 };
 
-const syncGlob = ({ cwd, patterns }: { cwd?: string; patterns: string | string[] }) =>
-  fg.sync(patterns, { cwd, followSymbolicLinks: false });
+const syncGlob = ({ cwd, patterns }: { cwd?: string; patterns: string | string[] }) => {
+  const hasAbsolutePattern = [patterns].flat().some(p => isAbsolute(p.replace(/^!/, '')));
+  return globSync(patterns, {
+    cwd,
+    absolute: hasAbsolutePattern,
+    followSymbolicLinks: false,
```
webpro, line 72: - Why not hard-code `absolute: true`?
- Why does the circular symlink test fail if we use `absolute: true`?

From the [tinyglobby docs](https://superchupu.dev/tinyglobby/documentation) alone it isn't immediately clear to me why one would affect the other (return absolute paths ↔ follow symlinks). Maybe @SuperchupuDev could shed a light on this?

webpro (commented):
This is really great! Feels like tinyglobby has indeed become more of drop-in compared to last time I looked. Minimal set of changes for maximum effect. Hopefully we can tackle that absolute/symlinks issue as well.

webpro:
Took a random repo (https://github.com/TanStack/query) for a perf spin on my Apple M1 machine (ran 3x each):

tinyglobby:

```
❯ knip --performance-fn defaultGlob
✂️  Excellent, Knip found no issues.

Name         size  min    max   median   sum  
-----------  ----  ----  -----  ------  ------
defaultGlob   371  0.00  21.64    0.67  717.82
defaultGlob   371  0.00  39.99    0.70  774.05
defaultGlob   371  0.00  21.98    0.67  724.47
```

main branch/fast-glob:

```
 ❯ knip --performance-fn defaultGlob
✂️  Excellent, Knip found no issues.

Name         size  min    max   median   sum  
-----------  ----  ----  -----  ------  ------
defaultGlob   371  0.00  20.66    0.51  618.13
defaultGlob   371  0.00  18.44    0.46  638.13
defaultGlob   371  0.00  31.69    0.46  629.07
```

Did not expect tinyglobby to be (a little bit) slower, would be good to understand that part better as well. I'll do so more digging.

packages/knip/src/util/glob.ts
webpro: Turns out this was an issue with the failing Knip test itself: relative ↔ absolute path assertion, because fast-glob magically returns absolute paths if we pass in absolute patterns without setting the `absolute` option (and vice versa).

webpro:
I think we're functionally good here, it's only the perf degradation for projects with a lot of `.gitignore` files and thus a lot of `ignore` patterns, for which I've opened https://github.com/SuperchupuDev/tinyglobby/pull/190.

AlexanderKaran:
That's great news wooooooo!

packages/knip/src/util/glob.ts
SuperchupuDev: hi, sorry i forgot to reply but yeah fast-glob magically returns absolute paths if the pattern is absolute even if the option is set to `false`

webpro (approved):
Thanks everyone. In particular @gameroman for putting in the work and sticking to it, and @SuperchupuDev for tinyglobby ❤️

Going to not-major bump knip with the [differences as documented](https://superchupu.dev/tinyglobby/comparison).

There have been some internal hassles re. un-ignore patterns, and we've covered/fixed at least the infamous Yarn plugins case in tests.

AlexanderKaran:
Woooooooo

webpro:

Issue #1127, opened 2025-06-09
💡 Could it be possible to add `op`, `fly` (`lfyctl`), and `cloudflared` as binaries?


hilja:
### Suggest an idea for Knip

- `op` - [1Password cli](https://developer.1password.com/docs/cli/)
- `flyctl` - (alias `fly`) [fly.io cli](https://fly.io/docs/flyctl/)
- `cloudflared` - [Cloudflare tunnel cli](https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/get-started/)

webpro:
Binaries that apply to this description can be added to the list:

https://github.com/webpro-nl/knip/blob/b346903187cebf19fc0516d2ea46bd8dec7d4dfe/packages/knip/src/constants.ts#L34-L37

If there are other things Knip should support (such as positional arguments that are entry points or arguments such as for configuration files or references to dependencies, etc etc), feel free to keep the conversation going and I'm happy to assist/implement.

webpro:
Closing this issue as part of a general cleanup to keep this project sustainable and optimize my time working on it. If you think this is inappropriate or if there is no workaround and you feel stuck, feel free to [open a new issue](./new/choose). Thanks for your understanding.

PR #1469, opened 2026-01-14
Add cloudflared, fly, uv, gcloud, and op binaries

hilja:
- `cloudflared`: cloudflare tunnels
- `flyctl` and it's alias `fly`: fly.io hosting
- `op`: 1Password
- ~~`biome`: there's a biome plugin, but some of use a globally installed biome to avoid downloading the package~~

https://github.com/webpro-nl/knip/issues/1127


joealden:
I feel like `biome` isn't something we should ignore, as I believe they recommend using it locally? As I presume ignoring it globally would reduce the benefit of the biome plugin?

@webpro other than the code comment, are there any guidelines on when something should be ignored globally?

I can create a separate PR, but I'd also add `uv` and `gcloud`, as these are the only two I need to ignore at work, and feel like ones that do make sense, as you can't install/access their bins via a JS package manager (to my knowledge).

hilja:
> I feel like biome isn't something we should ignore, as I believe they recommend using it locally? As I presume ignoring it globally would reduce the benefit of the biome plugin?

You make a good point.

webpro:
Please see https://github.com/webpro-nl/knip/blob/31eaaf544c88bd4c26003025da515a0f8ac134dd/packages/knip/src/constants.ts#L34-L38

The 6 year threshold is to avoid conflicts/surprises in case the package is available/installed locally. That's an arbitrary number, but have to draw the line somewhere.

```
$ last-publish cloudflared
The last publish date of cloudflared is 8/1/2025
$ last-publish fly
The last publish date of fly is 2/4/2025
$ last-publish flyctl
The last publish date of flyctl is 11/21/2025
$ last-publish gcloud
The last publish date of gcloud is 3/10/2017
$ last-publish op
The last publish date of op is 11/21/2018
$ last-publish uv
The last publish date of uv is 12/15/2022
```

hilja:
@webpro oh, the binaries should strictly be available in npm? Most of these are just brew installs, there are packages with the same name in npm, like [flyctl](https://www.npmjs.com/package/flyctl) but it's a js wrapper for the cli which is not maintained by fly.io.

> last npm publish was at least 6 years ago.

"Last", is this correct? Should it say "first"? Last publish 6 years ago would mean the package is kinda dead.

webpro:
Sounds like you got things backwards. That list contains binaries that are NOT in npm OR "kinda dead".

hilja:
Okay that makes more sense now 😅

joealden:
@webpro I feel like there is some nuance to be had with the NPM last publish date though?

Take `uv` for example - https://docs.astral.sh/uv/ is that we're actually using, and https://www.npmjs.com/package/uv doesn't even look to have a bin?

webpro:
> and https://www.npmjs.com/package/uv doesn't even look to have a bin?

Good point. In that case we can include it, regardless of last publish. Especially since the package is deprecated. In theory `bin` could have been removed or could still be added.

There is also the case where an entirely different package might have, say, the `eslint` or `tsc` bin defined.

webpro:
If there's still interest in getting this merged, then please at least remove [cloudflared](https://npmx.dev/package/cloudflared).

webpro:

PR #1473, opened 2026-01-17
Detect and install none installed packages

AlexanderKaran:
Adding commands that detect the package manager and give install options for import packages that are not installed.

I was initialy very defensive here walking up until I found a file in case someone opened a nested package of a workspace. I think this was me being silly so simplfied it with an error if it can not find the details it needs for an install.


webpro:
Thanks, I like it! Just tried in a monorepo using Yarn and it installed the unlisted dev dependency in the right workspace.

AlexanderKaran:
@webpro Phew, ....so glad it works on other computers to

webpro:
Right! I know the feeling 😅 😂

webpro:
Let's go, YOLO. Thanks again!

webpro:
:rocket: _This pull request is included in v5.82.0. See [Release 5.82.0](https://github.com/webpro-nl/knip/releases/tag/knip%405.82.0) for release notes._

MathiasWP:
Does this implementation support auto-installing with bun or deno?

webpro:
Looks like that's not the case. Might be a nice opportunity to contribute, if you're up to it! No worries if not.

AlexanderKaran:
That's not a bad idea. I will update it.

PR #1478, opened 2026-01-19
Added plugin for Serverless framework

BenCrinion-IW:
I've created the basics of a plugin for Serverless Framework.

Serverless Framework config file links to handles for functions which I've added as a productionEntryPoint.


BenCrinion-IW:
Hi, I need some help with this PR please.

Is this part of the test identifying there's an unused dependency? The serverless package isn't always imported by the handler, how do I say it's a necessary dependency?

```
dependencies: 1,
```

Also, how can I run Knip against my codebase with this plugin while it's in development?

packages/knip/src/plugins/serverless-framework/index.ts
```diff
@@ -0,0 +1,39 @@
...
+const resolveConfig: ResolveConfig<PluginConfig> = async config => {
+  const functions = Object.keys(config.functions);
+
+  return [...functions].map(id => convertHandlerToFileToProductionEntry(config.functions[id].handler));
+};
+
+const convertHandlerToFileToProductionEntry = (handler: string) => {
+  const lastIndexOfDot = handler.lastIndexOf('.');
+  const file = `${handler.substring(0, lastIndexOfDot)}.{js,ts}`;
+  return toDeferResolveProductionEntry(file)
```
webpro: From the fixture alone I'd say these could be `toProductionEntry` as no module resolution is needed? Unless it could point to an external dependency, which might not be the case here.

packages/knip/test/plugins/serverless-framework.test.ts
```diff
@@ -0,0 +1,21 @@
...
+const cwd = resolve('fixtures/plugins/serverless-framework');
+
+test('Find dependencies with the serverless-framework plugin', async () => {
+  const options = await createOptions({ cwd });
+  const { counters } = await main(options);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    dependencies: 1,
+    files: 2,
```
webpro: The `Unused files` means the plugin isn't working as intended yet, the plugin should return entries and entry files are then referenced/used.

packages/knip/test/plugins/serverless-framework.test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const cwd = resolve('fixtures/plugins/serverless-framework');
+
+test('Find dependencies with the serverless-framework plugin', async () => {
+  const options = await createOptions({ cwd });
+  const { counters } = await main(options);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    dependencies: 1,
```
webpro: Indeed, this means `serverless` isn't referenced anywhere.

Typically such dependencies are referenced in some source or JS/TS config file, or from one of the `package.json#scripts` (in that case, note the message from the test template "Missing binaries? Add: node_modules/pkg/package.json with bin → ./index.js").

webpro (commented):
Thanks Ben!

> Also, how can I run Knip against my codebase with this plugin while it's in development?

Using Bun or tsx: https://github.com/webpro-nl/knip/blob/main/.github/DEVELOPMENT.md#without-compilation

packages/knip/test/plugins/serverless-framework.test.ts
BenCrinion-IW: I don't understand the note from the test template but I've added scripts to the package.json in the fixture. Is that what the note means?

https://github.com/webpro-nl/knip/pull/1478/changes#diff-547da02503d0da639a21fec79177fead936f72c72f245223a05c48f71f973a0dR7

packages/knip/test/plugins/serverless-framework.test.ts
BenCrinion-IW: I've updated the plugin to use the productionEntry helper function. 

Does `files` in this counter represent unused files? I guess that means that my plugin isn't matching the paths of the handlers in the fixture correctly?

BenCrinion-IW: Okay, I've realised the issue with this. I had the incorrect relative path in the config file for the folder structure of the fixture.

This is fixed

packages/knip/fixtures/plugins/serverless-framework/serverless.yml
```diff
@@ -0,0 +1,21 @@
...
+
+provider:
+  name: aws
+  runtime: nodejs20.x
+  region: eu-west-1
+  logRetentionInDays: 1
+  stage: '${opt:stage, "dev"}'
+  environment:
+    STAGE: ${opt:stage, "dev"}
+    CLOUDFORMATION_VARIABLE: "!Sub 'arn:aws:iam::${AWS::AccountId}:role/role-name'"
```
BenCrinion-IW, line 14: Serverless uses CloudFormation intrinsic functions which make the YAML invalid.

This line shouldn't be in quotes but if I remove the quotes the parser fails because it doesn't expect `!Sub`

` CLOUDFORMATION_VARIABLE: !Sub 'arn:aws:iam::${AWS::AccountId}:role/role-name'`

Is there a way to tell the config parser to ignore errors? I've seen some documentation say that some config files are treated as plain text. If I can force that behaviour I can use regex to find the `handler: path/blah.blah` pattern.

@webpro

BenCrinion-IW:
@webpro I've improved the PR and fixed the issue with the unreferenced files but have further questions. Are you able to take another look at my comments please?

packages/knip/test/plugins/serverless-framework.test.ts
webpro: The note means that if you have such scripts in `package.json`, _then_ please add the following files to the fixture(s):

```jsonc
// node_modules/serverless/package.json
{
  "name": "serverless",
  "bin": {
    "serverless": "index.js"
  }
}
```

```
// node_modules/serverless/index.js (empty file)
```

The tests are much like "integration tests" and Knip looks at `package.json#bin` to connect binaries with packages (hence, the unlisted binary + unused dependency issues/counters were not 0 in the test).

packages/knip/fixtures/plugins/serverless-framework/serverless.yml
webpro: There is currently no way to ignore such parsing errors. And also no way to parse `.yml` as text and handle things differently. I'll see what I can do here, but it's not a priority for me at the moment.

webpro:
@BenCrinion-IW please see fb042ae235e57b340d167bb7bb7d7ddb0fa5b1fb - any chance you could give this PR another shot?

webpro:
Just went ahead, synced with main and wrapped up the PR. Thanks again Ben!

webpro:

BenCrinion-IW:
Thank you! I've been meaning to get around to it but it's been low on my list of priorities

Issue #1483, opened 2026-01-23
💡 Add --no-dependents flag for --workspace


WooWan:
## Problem

`--workspace` expands selection to include dependent workspaces. In CI with partial installs (e.g. `yarn workspaces focus`), dependents don't have their dependencies installed, causing Knip to fail when loading their configs.

`--strict` excludes dependents but implies --production, skipping devDependencies analysis. However, we need to analyze devDependencies while skipping dependent workspaces, so I propose a --no-dependents flag.

## Proposal
                                                                                
  Add --no-dependents flag: includes selected workspaces + ancestors, excludes dependents, without   
  implying --production.                                                                             
  
| Flag | Scope | devDeps |
|------|-------|---------|
| default | selected + ancestors + dependents | ✓ |
| `--strict` | selected only | ✗ (production) |
| `--no-dependents` | selected + ancestors | ✓ |

webpro:
Closing this one, please see #1485 for more details.

PR #1484, opened 2026-01-24
Add --no-dependents flag to exclude dependent workspace expansion  

WooWan:
- resolves: https://github.com/webpro-nl/knip/issues/1483
##  Summary                                                                                            
                                                                                                     
  Adds --no-dependents CLI flag that includes selected workspaces and their ancestors, but excludes  
  dependent workspaces. Unlike --strict, this does not imply --production, so devDependencies are    
  still analyzed.                                                                                    
                                                                                                     
##  Motivation                                                                                         
                                                                                                     
  In CI pipelines using partial installs (yarn workspaces focus, pnpm --filter), only explicitly     
  specified workspaces have their dependencies installed. Knip's default --workspace behavior expands
   to dependents, which then fail to load configs due to missing dependencies.                       
                                                                                                     
  --strict solves the expansion issue but also skips devDependencies analysis — this flag provides a 
  middle ground.

### DIFF
--- a/packages/knip/fixtures/workspaces-no-dependents/package.json
+++ b/packages/knip/fixtures/workspaces-no-dependents/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "@fixtures/no-dependents",
+  "workspaces": ["packages/*"],
+  "knip": {
+    "workspaces": {
+      "packages/*": {
+        "entry": "index.ts!",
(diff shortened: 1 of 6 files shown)


WooWan:
This PR contains unrelated commits, so I'll close it and open a new one.

PR #1485, opened 2026-01-24
Add --no-dependents flag to exclude dependent workspace expansion

WooWan:
- resolves: https://github.com/webpro-nl/knip/issues/1483
##  Summary                                                                                            
                                                                                                     
  Adds --no-dependents CLI flag that includes selected workspaces and their ancestors, but excludes  
  dependent workspaces. Unlike --strict, this does not imply --production, so devDependencies are    
  still analyzed.                                                                                    
                                                                                                     
##  Motivation                                                                                         
                                                                                                     
  In CI pipelines using partial installs (yarn workspaces focus, pnpm --filter), only explicitly     
  specified workspaces have their dependencies installed. Knip's default --workspace behavior expands
   to dependents, which then fail to load configs due to missing dependencies.                       
                                                                                                     
  --strict solves the expansion issue but also skips devDependencies analysis — this flag provides a 
  middle ground.

### DIFF
--- a/packages/knip/fixtures/workspaces-no-dependents/package.json
+++ b/packages/knip/fixtures/workspaces-no-dependents/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "@fixtures/no-dependents",
+  "workspaces": ["packages/*"],
+  "knip": {
+    "workspaces": {
+      "packages/*": {
+        "entry": "index.ts!",
+        "project": "**/*.ts!"
+      }
+    }
+  }
+}
--- a/packages/knip/fixtures/workspaces-no-dependents/packages/app/index.ts
+++ b/packages/knip/fixtures/workspaces-no-dependents/packages/app/index.ts
@@ -1,0 +1,3 @@
+import { helper } from '@fixtures/no-dependents__lib';
+
+export const result = helper;
--- a/packages/knip/fixtures/workspaces-no-dependents/packages/app/package.json
+++ b/packages/knip/fixtures/workspaces-no-dependents/packages/app/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "@fixtures/no-dependents__app",
+  "main": "./index.js",
+  "dependencies": {
+    "@fixtures/no-dependents__lib": "1.0.0"
+  }
+}
--- a/packages/knip/fixtures/workspaces-no-dependents/packages/lib/index.ts
+++ b/packages/knip/fixtures/workspaces-no-dependents/packages/lib/index.ts
@@ -1,0 +1,1 @@
+export const helper = 1;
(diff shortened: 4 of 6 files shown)


WooWan:
This is the only blocker to adopting knip in our repository. I'd appreciate it if you could take a look when you have time. Thanks!

webpro:
Thanks for the PR. Though I don't think this is a good way to solve your problem:

> Knip's default `--workspace` behavior expands to dependents, which then fail to load configs due to missing dependencies.

We shouldn't add another flag to tweak things leading to behavior that should not be in the program, as [dependents should be in the workspaces graph](https://knip.dev/features/monorepos-and-workspaces#filter-workspaces):

> Dependent workspaces may reference exports from the linted workspace.

Granted, the behavior of `--strict` is a bit arbitrary. Yet I'm not looking to add more ambiguity here.

So I guess my question is: why don't you install the dependencies?

WooWan:
> So I guess my question is: why don't you install the dependencies?

We manage a large monorepo, and managing CI time is critical to ship services on time. In a monorepo, there are tons of services and dependencies, and the workspace graph can get complicated. So we only install dependencies partially (full installation takes a significant amount of time in CI).

We need a way to run Knip only in the specific `--workspace` we specify. I believe the slight added complexity is worth it to give users this option, considering partial dependency installation is a common practice in monorepos.

webpro:
> So we only install dependencies partially

That's cool but it's not in line with how Knip works.

Here's another idea that you can do today: configure Knip so that the plugins loading the missing dependencies are disabled in the problematic dependent workspaces, e.g.:

```jsonc
{
  "workspaces": {
    "packages/lib": {
      "vite": { "config": [] },  // keep entry files, but do not load the config file
      "vite": false,             // disable plugin entirely
    }
  }
}
```

This way, you need to deal with the disabled plugin(s), but keep other inter-workspace connections properly. I would recommend trying this, regardless of what this PR offers.

WooWan:
Thanks for the suggestion! However, I see a limitation with the static config approach:                                                                                                                                                    
The knip.json configuration is static, so disabling plugins for specific workspaces would affect all scenarios:                                                                                                                                                                                                                                                                                                                                                                       

- When running knip --workspace lib → app's vite plugin should be disabled (deps not installed)
- When running knip --workspace app → app's vite plugin should be enabled (deps are installed)

Manually maintaining plugin settings to match actual dependency installation is error-prone and can easily lead to mismatches.  With --no-dependents, users can just run Knip for the workspaces they specify — no config maintenance, no mismatch issues.

That said, while discussing this, I realized there might be a more explicit and clearer approach than --no-dependents.

webpro:
> The knip.json configuration is static

The JSON was just to clarify my idea, you can use a function from e.g. `knip.config.ts` and get argument values from CLI.

> With --no-dependents, users can just run Knip for the workspaces they specify — no config maintenance, no mismatch issues.

Again, this not the solution we're after. It leads to false positives because dependent workspaces are not included in the workspaces graph. That's why I'm going to close this issue.

PR #1488, opened 2026-01-26
Avoid highlighting path-like specifiers

azat-io:
**Summary**

- Normalize `file:` specifiers to paths before creating issues
- Treat path‑like specifiers as paths in the symbols reporter (don’t inline‑highlight)
- Use `lastIndexOf` instead of `split()` to avoid `//` when a package name appears multiple times

**Problem**

When configs use `import.meta.resolve()` for plugins (e.g. Prettier), the returned `file:` URLs were shown as‑is in text output:

```
file:///Users/.../node_modules/.pnpm/prettier-plugin-astro@0.14.1/node_modules//dist/index.js  prettier.config.ts
```

**Issues:**

1. Full URL shown instead of package name
2. Double slash (`//`) in pnpm paths where the package name repeats JSON output was already correct (`"name": "prettier-plugin-astro"`).

**Solution**

1. Convert `file:` URLs to file paths via `fileURLToPath()`
2. Detect path‑like specifiers and show only the symbol
3. Use `lastIndexOf` instead of `split().join()` for highlighting

**After**

```
prettier-plugin-astro   prettier.config.ts
```


webpro:
Can you please add a fixture + test (or add something to an existing related test), so we can see the source of the problem first?

azat-io:
@webpro Done ✅

webpro:
Thanks Azat! We could simplify things a bit by just filing the issue with the package name instead of that file path (which is why we have `getPackageNameFromFilePath` in the first place).

azat-io:
Thanks!

webpro:
:rocket: _This pull request is included in v5.83.0. See [Release 5.83.0](https://github.com/webpro-nl/knip/releases/tag/knip%405.83.0) for release notes._

PR #1496, opened 2026-01-28
Add JetBrains plugin to docs

niklas-wortmann:


webpro:
It's a blog post, adding this weeks later is stretching that idea a bit too far imho. Also might give the impression that the JB plugin has the same feature-set as the VS Code extension.

I'll try to come up with something, like a similar dedicated page and of course link to your plugin as well.

niklas-wortmann:
Ohh didn't realize it's a blog post, looked like a normal docs page to me, sorry for that 😂 the /blog/... Kinda gave it away though 😂

webpro:
Starlight didn't have a clear option to blog within site at the time, didn't bother to find a better solution after that.

Just added the https://knip.dev/reference/integrations page. I've moved the VS Code screenshots all the way down so JB plugin is still prominent.

PR #1499, opened 2026-01-31
Move File Watching Request from vscode extension to LSP and auto start

phibr0:
I am working on a [Zed extension](https://github.com/phibr0/zed-knip) for the Knip LSP, but am facing two issues:

1. The LSP waits for a start request, which Zed won't send
2. File watching is implemented via the vscode extension (which I also can't in Zed)

By auto starting and watching files via a request of the LSP it works seamlessly and it should be even easier to make it work for other editors aswell.

### DIFF
--- a/packages/language-server/src/server.js
+++ b/packages/language-server/src/server.js
@@ -39,7 +39,7 @@ const toPosix = value => value.split(path.sep).join(path.posix.sep);
 
 /**
  * @import { Issues, Rules } from 'knip/session';
- * @import { Connection, Diagnostic, CodeAction, ClientCapabilities } from 'vscode-languageserver';
+ * @import { Connection, Diagnostic, CodeAction } from 'vscode-languageserver';
  * @import { CodeActionParams, DidChangeWatchedFilesParams } from 'vscode-languageserver';
  * @import { Config, IssuesByUri } from './types.js';
  *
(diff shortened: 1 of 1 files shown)


webpro:
Thanks for the PR! Agreed this should become more flexible.

The issue with auto-start is that there should still be a way for users to defer creating the session, e.g. for large and/or multiple workspaces.

So I've started work in the `feat/ls-ext-2` branch. Would be really great if we could align on that one. What do you think? Any chance you could work with and adapt if necessary? Including the file watcher part?

There's some more major change in there though, sorry if that's overwhelming. But at least there's a start with auto-start :)

phibr0:
Yes, that's a better implementation for the auto start behavior! Zed's extensions are still quite primitive though (you are really just writing a function that provides a shell command that spawns a language server) and don't allow for extra configuration. So we can't enable the file watching/synchronization via the extension, and I believe the LSP needs to request it. I could write a wrapping LSP that does this, which then forwards everything, but I think that's unnecessary.

webpro:
Feel free to add to that branch (or PR against it) so I can see what you need / what works with Zed.

webpro:
Any chance you could give this a shot? Install latest using e.g.

```
npm i https://pkg.pr.new/@knip/language-server@51b7dc1
```

phibr0:
Will take a look at the changes this weekend, seems like it doesn't watch for changes just yet. :/

webpro:

PR #1500, opened 2026-02-01
Added new Ignore Tag

AlexanderKaran:
Working on large internal projects, we often take routes or code offline, either by using feature flags or by removing routes and components. Knip then picks these up as code to clean up however, I don't want to flat out ignore them forever (we all forget we did this)

Added a new JSDoc tag with a date value. This way the code is ignored until the date passes, at which point Knip will flag it again, and you need to action it


webpro:
Thanks for the PR! Doing some housekeeping. As discussed, closing this in favor of #1532, hoping that will be the way forward.

PR #1507, opened 2026-02-04
Handle Prettier config re-exports

azat-io:
Fix false positive “unlisted dependencies” for Prettier plugins when config is re-exported.

If a Prettier config file is a pure re-export (ESM `export { default } from 'pkg'` or CJS `module.exports = require('pkg')`), Knip should not report plugins from that external config as dependencies of the consumer.

Implementation: detect pure re-export via AST and skip `plugins` resolution for that config.

### DIFF
--- a/packages/knip/fixtures/plugins/prettier-reexport/node_modules/@org/prettier-config/index.js
+++ b/packages/knip/fixtures/plugins/prettier-reexport/node_modules/@org/prettier-config/index.js
@@ -1,0 +1,3 @@
+export default {
+  plugins: [import.meta.resolve('prettier-plugin-test')],
+};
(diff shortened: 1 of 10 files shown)


azat-io:
For context, the problem occurs when using this Prettier config:
https://github.com/azat-io/prettier-config/blob/main/index.ts#L104-L109

webpro:
Thanks, Azat!

Since I think this is a generic challenge, I'm trying to come up with a more generic solution. This PR has logic I'm hoping to optimize in that solution.

webpro:
Any chance you could give my attempt a shot in your project(s)? E.g.:

```
npm i -D https://pkg.pr.new/knip@8678df2
```

azat-io:
I tried testing it. Looks good, problem solved! 👍

webpro:
Thanks for testing! I've added an additional improvement: if the config file is a barrel file that re-exports only external dependencies, we don't need to load it at all anymore. Knip will just do it's regular AST traversal to get the imports, but not actually load up the file: perf win + less false positives.

webpro:

PR #1516, opened 2026-02-08
Fix path resolution

intoop:

### DIFF
--- a/packages/language-server/src/server.js
+++ b/packages/language-server/src/server.js
@@ -57,6 +57,9 @@ export class LanguageServer {
   connection;
 
   /** @type {undefined | string} */
+  workspaceRoot;
+
+  /** @type {undefined | string} */
   cwd;
 
   /** @type Set<string> */
@@ -91,7 +94,8 @@ export class LanguageServer {
 
       if (!uri) return { capabilities: {} };
 
-      this.cwd = fileURLToPath(uri);
+      this.workspaceRoot = fileURLToPath(uri);
+      this.cwd = this.workspaceRoot;
 
       const capabilities = {
         codeActionProvider: {
@@ -186,10 +190,10 @@ export class LanguageServer {
       const configFilePath = config.configFilePath
         ? path.isAbsolute(config.configFilePath)
           ? config.configFilePath
-          : path.resolve(this.cwd ?? process.cwd(), config.configFilePath)
+          : path.resolve(this.workspaceRoot ?? process.cwd(), config.configFilePath)
         : undefined;
 
-      if (configFilePath) this.cwd = path.dirname(configFilePath);
+      this.cwd = configFilePath ? path.dirname(configFilePath) : this.workspaceRoot;
       if (this.cwd) process.chdir(this.cwd);
 
       this.connection.console.log('Creating options');


webpro:
Thanks for the PR! What was the error? I need a bit more details before I can merge.

intoop:
Sorry for insufficient information.

---

When knip.json is not in the root, I have to modify the settings:

```
{
"knip.configFilePath": "./src/knip.json",
}
```
But this only works for the first time. After each time I "Restart Knip Server", the path becomes buggy.
```
Error: Error: ENOENT: no such file or directory, chdir 'd:\Projects\test\src' -> 'd:\Projects\test\src\src'
Error: Error: ENOENT: no such file or directory, chdir 'd:\Projects\test\src' -> 'd:\Projects\test\src\src\src'
Error: Error: ENOENT: no such file or directory, chdir 'd:\Projects\test\src' -> 'd:\Projects\test\src\src\src\src'
```
And this PR fixes this bug

packages/language-server/src/server.js
```diff
@@ -186,10 +190,10 @@ export class LanguageServer {
       const configFilePath = config.configFilePath
         ? path.isAbsolute(config.configFilePath)
           ? config.configFilePath
-          : path.resolve(this.cwd ?? process.cwd(), config.configFilePath)
+          : path.resolve(this.workspaceRoot ?? process.cwd(), config.configFilePath)
         : undefined;
 
-      if (configFilePath) this.cwd = path.dirname(configFilePath);
+      this.cwd = configFilePath ? path.dirname(configFilePath) : this.workspaceRoot;
```
webpro, line 196: iiuc, `this.cwd` might end up as `./src` in your case, right? that likely will cause issues if `knip.json` is not in root.

The documentation at https://knip.dev/overview/configuration#location indicates you could put the `knip.json` file anywhere, but that's currently not true in combination with using VS Code extension.

So either we need a better fix or you just need to move `knip.json` one directory up.

webpro (commented):
Thanks for the explanation. Please see my comment.

packages/language-server/src/server.js
intoop: I tried to move `knip.json` to the root, but in SvelteKit projects, it causes `No Svelte config file found in D:\Projects\my_project - using SvelteKit's default configuration without an adapter.`

Just like the documentation mentions, `knip.json` must be in the same directory as Svelte configurations. However, I cannot move Svelte configurations because only one directory uses Svelte, not the whole project.

> This plugin is enabled if "svelte" is listed in "dependencies" or "devDependencies" in package.json.
> This plugin is enabled if "@sveltejs/kit" is listed in "dependencies" or "devDependencies" in package.json.

---

So rather than move `knip.json` one directory up or finish the `knip.json` searching logic for VS Code extension, set `"knip.configFilePath": "./tauri/knip.json"` is an easier way to fix this. And I believe something like this is not the expected behavior anyways:
```
Error: ENOENT: no such file or directory, chdir 'd:\Projects\test\src' -> 'd:\Projects\test\src\src'
Error: ENOENT: no such file or directory, chdir 'd:\Projects\test\src' -> 'd:\Projects\test\src\src\src'
Error: ENOENT: no such file or directory, chdir 'd:\Projects\test\src' -> 'd:\Projects\test\src\src\src\src'
```

intoop:
Thanks. Please see my reply.

webpro:
Did you try an absolute path already? e.g.

```json
{
"knip.configFilePath": "d:\Projects\test\src\knip.json",
}
```

intoop:
I tried an absolute path. It works, but I want to share the settings with other contributors so that they don't have to do this manually. That's why I use a relative path.

webpro:
Any chance you could remove the `process.chdir()` logic, so that we only need a single prop for working dir?

That's a cleaner implementation to begin with — I'll deal with that missing `chdir` later if necessary.

webpro:
While I was working on some related things I figured I could handle this as well.

Would be great if you could give it a spin before it goes GA!

webpro:

PR #1520, opened 2026-02-08
Fix issue with function expo plugins

tomhicks:
Fixes an issue where ExpoConfig types are wrong, and plugins can be functions. This gets some kind of string out of the function. If it's a named function, it's nice. If it's an inline function or arrow function it's probably not gonna be great but it's better than a crash.

Alternatively we could like have an autoincrementing integer name like `function_plugin_1`


webpro:
Thanks for the pull request! Any chance you could add a test so we can see what's being fixed?

Edit: there's existing fixtures you could add it to.

tomhicks:
given that a go. i'm working in the web ide because i don't have knip checked out so might need a bit of formatting or might not even pass, who knows.

PR #1528, opened 2026-02-13
Ignore RNC CLI and metro internals in react-native projects

wojtekmaj:
This PR resolves two issues in React Native projects:

- @react-native-community/cli and its siblings (-platform-ios for example) are listed as unused
- metro-transform-worker and metro-minify-terser get flagged as unresolved imports - yet they aren't directly used in metro.config.ts, they are pulled in via getDefaultConfig().

PR #1533, opened 2026-02-16
Treat react-native.config.js as config file in react-native projects

wojtekmaj:
Missed that one in #1528

### DIFF
--- a/packages/knip/fixtures/plugins/metro/react-native.config.js
+++ b/packages/knip/fixtures/plugins/metro/react-native.config.js
@@ -1,4 +1,4 @@
-export default {
+module.exports = {
   project: {
     ios: {},
   },


packages/knip/src/plugins/metro/index.ts
```diff
@@ -13,7 +13,7 @@ const enablers = ['metro', 'react-native'];
 
 const isEnabled: IsPluginEnabled = options => hasDependency(options.dependencies, enablers);
 
-const config = ['metro.config.{js,cjs,json}', 'package.json'];
+const config = ['metro.config.{js,cjs,json}', 'package.json', 'react-native.config.js'];
```
webpro, line 16: `config` files are loaded and processed by `resolveConfig` — can both config files be used interchangeably?

- No, we just need to add it as normal entry file → move to `const entry = []` in this plugin.
- No, it might contain `Input` items like dependencies or entry files that need custom parsing (like this plugin) → add switch/case in this plugin's `resolveConfig` or move to new react-native plugin.
- Yes → we're good.

webpro (commented):
Thanks for the PR! I wonder if this is the intention though, please see my comment.

packages/knip/src/plugins/metro/index.ts
webpro: Btw `config` files are always added to the program as an `entry` automatically so their `import` statements are included/resolved.

wojtekmaj: They are not "interchangeable". While React Native and Metro are quite tightly coupled, RN config is not the same as Metro config. Yet from Knip's perspective they should be, as far as I understand, treated equally:

- They both may, or may not be present
- They should both be treated as used, if present in a React Native project
- They can both import other dependencies
- They can potentially include references that allow further, more complex validation mechanisms, e.g. via `assets` path specified - kinda like Playwright config can reference Playwright setup file and mark it "used" even if it's not technically imported anywhere in our codebase

webpro: Sounds like we should split up in two plugins.

- if `metro` is listed in `package.json` → handle `metro.config`
- same for `react-native` → `react-native.config.js`

If this is not a viable route then that'd be pretty messed up. I'll try to whip something up.

wojtekmaj: I think that would be a step in the right direction. Especially since Metro can definitely be used outside React Native.

However, in most React Native projects, Metro isn't a direct dependency.

In my project, the only thing with "metro" in name in package.json is `@react-native/metro-config`, and `yarn why metro -R` returns:

```
└─ app-native@workspace:.
   ├─ @react-native/metro-config@npm:0.84.0 (via npm:0.84.0)
   │  └─ metro-config@npm:0.83.3 (via npm:^0.83.3)
   │     └─ metro@npm:0.83.3 (via npm:0.83.3)
   │        └─ metro-config@npm:0.83.3 (via npm:0.83.3)
   └─ react-native@npm:0.84.0 [13114] (via npm:0.84.0 [13114])
      └─ @react-native/community-cli-plugin@npm:0.84.0 [d8f1f] (via npm:0.84.0 [d8f1f])
         ├─ metro@npm:0.83.3 (via npm:^0.83.3)
         └─ metro-config@npm:0.83.3 (via npm:^0.83.3)
```

So if I had to guess, we'd indeed need 2 plugins:

- react-native one with `react-native` being an exclusive enabler
- metro one with `metro` _AND_ `react-native` being an enabler

How does that sound?

webpro: Thanks! Would be great if we could pick it up from here: https://github.com/webpro-nl/knip/pull/1538

wojtekmaj:
Abandoning in favor of obviously superior #1538

PR #1538, opened 2026-02-17
Extract and extend react-native plugin

webpro:
Let's see if we can untangle Metro and RN plugins properly.

### DIFF
--- a/packages/knip/fixtures/plugins/metro-defaults/package.json
+++ b/packages/knip/fixtures/plugins/metro-defaults/package.json
@@ -1,8 +1,6 @@
 {
   "name": "@plugins/metro-defaults",
   "dependencies": {
-    "metro": "*",
-    "react": "*",
-    "react-native": "*"
+    "metro": "*"
   }
 }
--- a/packages/knip/fixtures/plugins/metro/package.json
+++ b/packages/knip/fixtures/plugins/metro/package.json
@@ -1,10 +1,8 @@
 {
   "name": "@plugins/metro",
   "dependencies": {
-    "expo-asset": "*",
-    "metro": "*",
-    "react": "*",
-    "react-native": "*",
+    "@react-native/metro-config": "*",
+    "metro-minify-esbuild": "*",
     "react-native-svg-transformer": "*"
   },
   "metro": {
--- a/packages/knip/src/plugins/react-native/index.ts
+++ b/packages/knip/src/plugins/react-native/index.ts
@@ -1,4 +1,4 @@
-import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import type { IsPluginEnabled, Plugin, Resolve, ResolveConfig } from '../../types/config.js';
 import { type Input, toDependency } from '../../util/input.js';
 import { hasDependency } from '../../util/plugin.js';
 import type { ReactNativeConfig } from './types.js';
@@ -34,11 +34,11 @@ const resolveConfig: ResolveConfig<ReactNativeConfig> = async config => {
     }
   }
 
-  for (const pkg of RN_CLI_PACKAGES) {
-    inputs.push(toDependency(pkg, { optional: true }));
-  }
-
   return inputs;
+};
+
+const resolve: Resolve = () => {
+  return RN_CLI_PACKAGES.map(pkg => toDependency(pkg, { optional: true }));
 };
 
 const plugin: Plugin = {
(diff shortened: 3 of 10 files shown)


packages/knip/fixtures/plugins/metro-defaults/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@plugins/metro-defaults",
   "dependencies": {
+    "metro": "*",
     "react": "*",
     "react-native": "*"
```
wojtekmaj: I'd even consider removing these deps, to ensure Metro enabler works well.

packages/knip/fixtures/plugins/metro/package.json
```diff
@@ -1,10 +1,8 @@
...
   "name": "@plugins/metro",
   "dependencies": {
-    "@react-native-community/cli": "*",
-    "@react-native-community/cli-platform-android": "*",
-    "@react-native-community/cli-platform-ios": "*",
     "expo-asset": "*",
+    "metro": "*",
     "react": "*",
     "react-native": "*",
     "react-native-svg-transformer": "*"
```
wojtekmaj, line 6: ...and since this fixture is kinda-sorta about both Metro AND React Native, perhaps leave these intact as you did

packages/knip/src/plugins/metro/index.ts
```diff
@@ -9,7 +9,7 @@ import type { MetroConfig } from './types.js';
 
 const title = 'Metro';
 
-const enablers = ['metro', 'react-native'];
+const enablers = ['metro', '@react-native/metro-config'];
```
wojtekmaj, line 12: :+1:

packages/knip/src/plugins/react-native/index.ts
```diff
@@ -0,0 +1,52 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.js';
+import { type Input, toDependency } from '../../util/input.js';
+import { hasDependency } from '../../util/plugin.js';
+import type { ReactNativeConfig } from './types.js';
+
+// https://github.com/react-native-community/cli/blob/main/docs/configuration.md
+
+const title = 'React Native';
+
+const enablers = ['react-native', '@react-native-community/cli'];
```
wojtekmaj: That's not technically incorrect but there's absolutely zero point of using RNC CLI without react-native, so we could just:

```suggestion
const enablers = ['react-native'];
```

packages/knip/src/plugins/react-native/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['react-native.config.{js,cjs,ts,mjs}'];
```
wojtekmaj: At the moment, it's ONLY .js, other extensions are not accepted. It might be handy to let people know that the non-.js files they created might not get picked up. On the other hand, should RN ever add support for more exts, we'd need to update that back.

```suggestion
const config = ['react-native.config.js'];
```

webpro:
Thanks for the review @wojtekmaj. I've fixed and cleaned up a bit more, to have better tests moving forward (e.g. cleaner metro tests and a combined metro-react-native test).

Personally I'd be good to go with this, rather have you and others run it in actual projects with real reproductions (I'm happy to improve fixtures/tests as we go, but not so much in isolation like here in this PR, gonna be ping-pong).

wojtekmaj:
`npx https://pkg.pr.new/knip@1538` looks good in my RN project on the happy path.

`babelTransformerPath` dependency picked up correctly.
Making a typo in `babelTransformerPath` -> errored correcrly
Renaming `react-native.config.js` to `react-native.config.ts` -> errored correctly.
Removing `react-native.config.js` still correctly recognized project as RN.

Sanity check on non-RN project produced satisfactory results.

wojtekmaj:
packages/knip/schema.json + docs require updating with new plugin?

webpro:
Thanks for test driving this!

> packages/knip/schema.json + docs require updating with new plugin?

`schema.json` is updated indeed; docs are generated by `packages/docs`

webpro:
:rocket: _This pull request is included in v5.84.0. See [Release 5.84.0](https://github.com/webpro-nl/knip/releases/tag/knip%405.84.0) for release notes._

PR #1540, opened 2026-02-17
Add pm2 plugin

### LINKED ISSUE #1539: 💡 Read PM2 config to derive entry files
### Suggest an idea for Knip

In PM2, you can define `pm2.config.json` file that describes "apps". Example:

```json
{
  "apps": [
    {
      "name": "api",
      "script": "./src/index.ts",
    },
    {
      "name": "cron",
      "script": "./src/cron/index.ts",
    },
    {
      "name": "uptime_monitor",
      "script": "./src/uptime_monitor.ts",
    }
  ]
}
```

At the moment, Knip does not read this configuration file and would consider `./src/cron/index.ts` as unused.

wojtekmaj:
Closes #1539

### DIFF
--- a/packages/knip/test/plugins/pm2.test.ts
+++ b/packages/knip/test/plugins/pm2.test.ts
@@ -7,6 +7,7 @@ import { createOptions } from '../helpers/create-options.js';
 import { resolve } from '../helpers/resolve.js';
 
 const cwd = resolve('fixtures/plugins/pm2');
+const cwdEcosystem = resolve('fixtures/plugins/pm2-ecosystem');
 
 test('Find entry files with the pm2 plugin', async () => {
   const options = await createOptions({ cwd });
@@ -14,6 +15,21 @@ test('Find entry files with the pm2 plugin', async () => {
 
   assert(!issues.files.has(join(cwd, 'src/another.js')));
   assert(issues.files.has(join(cwd, 'src/unused.js')));
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    files: 1,
+    processed: 3,
+    total: 3,
+  });
+});
+
+test('Find entry files with the pm2 plugin from ecosystem config', async () => {
+  const options = await createOptions({ cwd: cwdEcosystem });
+  const { issues, counters } = await main(options);
+
+  assert(!issues.files.has(join(cwdEcosystem, 'src/another.js')));
+  assert(issues.files.has(join(cwdEcosystem, 'src/unused.js')));
 
   assert.deepEqual(counters, {
     ...baseCounters,
--- a/packages/knip/fixtures/plugins/pm2-ecosystem/ecosystem.config.json
+++ b/packages/knip/fixtures/plugins/pm2-ecosystem/ecosystem.config.json
@@ -1,0 +1,12 @@
+{
+  "apps": [
+    {
+      "name": "api",
+      "script": "./src/index.js"
+    },
+    {
+      "name": "another",
+      "script": "./src/another.js"
+    }
+  ]
+}
--- a/packages/knip/fixtures/plugins/pm2-ecosystem/node_modules/pm2/package.json
+++ b/packages/knip/fixtures/plugins/pm2-ecosystem/node_modules/pm2/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "pm2",
+  "bin": {
+    "pm2": "index.js"
+  }
+}
--- a/packages/knip/fixtures/plugins/pm2-ecosystem/package.json
+++ b/packages/knip/fixtures/plugins/pm2-ecosystem/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "@plugins/pm2-ecosystem",
+  "scripts": {
+    "start": "pm2 start ecosystem.config.json"
+  },
+  "devDependencies": {
+    "pm2": "*"
+  }
+}


wojtekmaj:
"Works on my machine".

Before:

```
wojciech.maj@MacBook-Pro server % npx knip                             
Unused files (6)
src/cron/[redacted].ts                                    
src/cron/index.ts                                      
src/cron/[redacted].ts                    
src/cron/[redacted].ts
src/instrument.ts                                      
src/[redacted].ts 
```

After:

```
wojciech.maj@MacBook-Pro server % npx https://pkg.pr.new/knip@1540
Unused files (1)
src/instrument.ts
```

packages/knip/src/plugins/pm2/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  'pm2.config.cjs',
+  'pm2.config.mjs',
+  'ecosystem.config.json',
+  'ecosystem.config.js',
+  'ecosystem.config.cjs',
+  'ecosystem.config.mjs',
+];
+
+const isEnabled: IsPluginEnabled = ({ cwd, dependencies }) =>
+  hasDependency(dependencies, enablers) || configFileNames.some(name => isFile(cwd, name));
```
webpro: A bit on the fence with this one: not sure if we should support the globally-installed scenario. I know it's pretty common for pm2 but also it's just not how Knip's designed. Might set a precedent. And not best practice to not version/control any npm-available dependency? And this function is O(1) with `hasDependency` in nearly all cases but here we are doing fs ops (adds up unnecessarily if many workspaces not even using pm2).

packages/knip/test/plugins/pm2.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    files: 1,
+    processed: 3,
+    total: 3,
+  });
+});
+
+test('Find entry files with the pm2 plugin without pm2 in dependencies', async () => {
```
webpro: Maybe use second test for `ecosystem.config.*` instead? 😬

packages/knip/src/plugins/pm2/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const config = ['pm2.config.{json,js,cjs,mjs}', 'ecosystem.config.{json,js,cjs,mjs}'];
+
+const addApplicationEntry = (application: PM2Application, entries: ReturnType<typeof toEntry>[]) => {
+  if (application.script && isInternal(application.script)) entries.push(toEntry(application.script));
```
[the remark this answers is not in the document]
webpro: Can you please turn off that Copilot rubbish. It's distracting the review process and actually not very good/complete (in general agents aren't very good with Knip yet because they're trained on different single-file linter logic).

We can probably omit `isInternal`. Then use one of:

- If we know it's a local file, use `toEntry`
- If we know it's a local production file, use `toProductionEntry`
- If it could be a dependency as well, use `toDeferResolve`

Core will handle absolute/relative paths. Filter out URLs indeed if applicable.

Also see https://knip.dev/writing-a-plugin/inputs

webpro (commented):
Thanks Wojciech! Happy to merge, but not sure whether we should support the global case.

packages/knip/src/plugins/pm2/index.ts
wojtekmaj: Ahhh, I see where you're going. Sadly pm2 is highly recommending a global install in its docs and consequently most of the folks will have this installed as a separate tool, and finding its config file is the only way to determine that it might be in use.

We only really have 3 options:
- Keep it as I implemented it
- Add a way to enable plugins manually
- Remove global pm2 support entirely

packages/knip/src/plugins/pm2/index.ts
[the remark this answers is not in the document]
wojtekmaj: Honestly no clue how it got there, I thought it's enabled by maintainer on a per-repo basis. Sorry!

packages/knip/schema.json
```diff
@@ -636,6 +636,10 @@
           "title": "playwright-test plugin configuration (https://knip.dev/reference/plugins/playwright-test)",
           "$ref": "#/definitions/plugin"
         },
+        "pm2": {
+          "title": "pm2 plugin configuration (https://knip.dev/reference/plugins/pm2)",
```
wojtekmaj, line 640: Note: This would need to be actually added

packages/knip/src/plugins/pm2/index.ts
webpro: Projects can do pm2:true in Knip config! Per workspace or "rooted".

packages/knip/test/plugins/pm2.test.ts
wojtekmaj: Done! Good idea. I used ecosystem.config.json because that was the "most different" case.

packages/knip/src/plugins/pm2/index.ts
wojtekmaj: Geez I should be reading docs more carefully!

wojtekmaj:
pkg.pr.new still works after recent pushes :)

packages/knip/src/plugins/pm2/index.ts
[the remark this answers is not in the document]
webpro: No worries. There's copy like "Copilot started reviewing on behalf of wojtekmaj 18 hours ago" in the UI. I've disabled "Automatic Copilot code review" at https://github.com/settings/copilot/features.

webpro:
Thank you!

webpro:
:rocket: _This pull request is included in v5.84.0. See [Release 5.84.0](https://github.com/webpro-nl/knip/releases/tag/knip%405.84.0) for release notes._

PR #1541, opened 2026-02-18
Make knip compatible with erasableSyntaxOnly

wojtekmaj:
By making knip compatible with erasableSyntaxOnly, we're making knip compatible with other systems that could potentially parse or transpile knip codebase, but are unable (or rather: unwilling) to parse TS features that require JS emit changes.

For example, we could:
- replace `tsc` with another compiler like swc
- drop `tsx` dependency from knip repository (Node.js supports TypeScript natively, but with erasable syntax only)
  - Making smoke tests significantly faster in the process

This change also eliminates a whole class of potential bugs related to enums, namespaces legacy decorators and other syntax this setting is forbidding.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,9 @@ jobs:
       - name: Install earliest supported peer dependencies
         run: pnpm add typescript@5.0.4 @types/node@18
         working-directory: packages/knip
+      - name: Remove erasableSyntaxOnly from tsconfig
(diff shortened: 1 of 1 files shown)


webpro:
Great! I was also thinking of going `.js` → `.ts` import specifier everywhere so we can run `node src/cli.ts` directly.

wojtekmaj:
> Great! I was also thinking of going `.js` → `.ts` import specifier everywhere so we can run `node src/cli.ts` directly.

Yes, I think that's a fantastic direction. Only it was too disruptive to just throw a PR at you without discussing first :D

webpro:

PR #1544, opened 2026-02-18
Fix false positives for arrow-wrapped dynamic imports assigned to variables

### LINKED ISSUE #1543: 🐛 false positives for lazy imports
### Prerequisites

### Reproduction url

https://github.com/jantimon/knip/blob/c9f590f1666035d5c335a6ee919b8d37cc6fc0ca/packages/knip/fixtures/imports-opaque/index.ts#L23-L30

### Reproduction access

### Description of the issue

We have a large React codebase where we use lazy loading pretty much everywhere, routes, heavy components, feature-flagged modules etc. 

The pattern looks like this:

```ts
const loadDashboard = () => import('./Dashboard');
const loadSettings = () => import('./Settings');

registerRoutes([
  { path: '/dashboard', load: loadDashboard },
  { path: '/settings', load: loadSettings },
]);
```

knip reported a ton of false positives. Every named export from these lazily imported modules shows up as unused

Minimal repro:

```ts
// module.ts
export default MyComponent;
export type Props = { ... };

// routes.ts
const loader = () => import('./module.js');
loadLazy(loader);
```

Knip reports `Props` (and any other named export from `module.ts`) as unused. Seems like it treats `const loader = () => import('./module.js')` the same as `const loader = await import('./module.js')`, as if `loader` holds the module directly rather than a function that returns it

Since theres no property accesses on `loader` it just assumes only `default` is imported

Expected: the import should be treated as **opaque** (like it already is for bare `() => import(...)` not assigned to a variable) since theres no way to statically know which exports are used at the call site

jantimon:
fixes #1543

the variable declaration branch in `importCall.ts` was matching `const loader = () => import('./module.js')` because `node.parent.parent` skips over the arrow function and lands on the `VariableDeclaration`

The fix  is to checking for a function boundary and if one exists the import is marked as opaque instead of resolving to `default`

Test: 
added the `const loader = () => import('./variable.js')` case to the opaque imports fixture, plus a `const direct = import('./direct.js')` negative case to verify direct imports still resolve normally

### DIFF
--- a/packages/knip/fixtures/imports-opaque/direct.ts
+++ b/packages/knip/fixtures/imports-opaque/direct.ts
@@ -1,0 +1,1 @@
+export default 1;
--- a/packages/knip/fixtures/imports-opaque/index.ts
+++ b/packages/knip/fixtures/imports-opaque/index.ts
@@ -19,3 +19,12 @@ const obj = {
   key2: await import('./awaited-assignment.js'),
   ...(await import('./obj-spread.js')),
 };
+
+const loader = () => import('./variable.js');
+function fnRun(load: () => Promise<unknown>) {
+  return load();
+}
+fnRun(loader);
+
+const direct = import('./direct.js');
+fn(direct, 0);
--- a/packages/knip/fixtures/imports-opaque/variable.ts
+++ b/packages/knip/fixtures/imports-opaque/variable.ts
@@ -1,0 +1,2 @@
+export default 1;
+export const named = 1;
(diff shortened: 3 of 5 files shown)


webpro:
Thanks Jan Nicklas! Glad you dug in and fixed it properly, happy to merge this.

webpro:
:rocket: _This pull request is included in v5.84.1. See [Release 5.84.1](https://github.com/webpro-nl/knip/releases/tag/knip%405.84.1) for release notes._

jantimon:
@webpro thank you soo much for the quick release - awesome maintained!

PR #1547, opened 2026-02-18
Fix require.context regex matching for path-aware patterns

jstaab:
Webpack's `require.context` returns keys prefixed with `./` (like`./components/Button.js`), so users write their filter regexes against that format. But we use `fast-glob` here which returns bare relative paths `components/Button.js`, causing any regex that references the `./` prefix to never match.

This fix prepends `./` to each path before testing the regex, aligning with webpack's actual behavior. 

I've added a new test and accompanying fixture which gets reported as unused without the fix. With the fix, it reports as used.


webpro:
Perfect. Thank you!

webpro:

PR #1548, opened 2026-02-19
Rewrite import specifiers to use .ts extensions, remove tsx

wojtekmaj:
Follow-up after discussion in #1541 

Likely to fail on older versions of Node (<22.18.0) and/or TypeScript (<5.7), will have to figure out what to do with tests (prebuild on newer version before running on older one? freezing this PR until Node.js 20 is dropped?) 

<details>

<summary>Script used for mass rewrite</summary>

```ts
node <<'NODE'
const fs=require('fs');
const path=require('path');
const exts=['.ts','.tsx','.mts','.cts'];
function walk(dir, out=[]){
  for(const e of fs.readdirSync(dir,{withFileTypes:true})){
    if(e.name==='node_modules' || e.name==='.git' || e.name==='dist') continue;
    const p=path.join(dir,e.name);
    if(e.isDirectory()) walk(p,out);
    else if(e.isFile() && p.endsWith('.ts') && "'!p.endsWith('"'.d.ts')) out.push(p);
  }
  return out;
}
const files=walk('.');
const re=/(?:from\\s*['"]|import\\s*['"]|import\\s*\\(\\s*['"]|require\\s*\\(\\s*['"])(\\.{1,2}\\/["'^'"'"\\n?#]+?)\\.js(?=['"])/g;
const hits=[];
for(const file of files){
  const text=fs.readFileSync(file,'utf8');
  let m;
  while((m=re.exec(text))){
    const rel=m[1];
    const base=path.resolve(path.dirname(file),rel);
    let hasTs=false;
    for(const ext of exts){
      if(fs.existsSync(base+ext)){hasTs=true;break;}
    }
    if(hasTs) hits.push(file);
  }
}
console.log([...new Set(hits)].sort().join('\\n'));
NODE
```

</details>

### DIFF
--- a/packages/knip/package.json
+++ b/packages/knip/package.json
@@ -46,9 +46,9 @@
     "lint": "biome lint",
     "format": "biome format --write",
     "test": "node scripts/run-test.ts",
-    "test:node": "node --test test/*.test.ts test/**/*.test.ts",
+    "test:node": "tsx --test test/*.test.ts test/**/*.test.ts",
     "test:bun": "bun test test/*.test.ts test/**/*.test.ts",
-    "test:smoke": "glob-bin -c \"node --test\" \"test/*.test.ts\" && glob-bin -c \"node --test\" \"test/{plugins,util}/*.test.ts\"",
+    "test:smoke": "glob-bin -c \"tsx --test\" \"test/*.test.ts\" && glob-bin -c \"tsx --test\" \"test/{plugins,util}/*.test.ts\"",
     "test:bun:smoke": "bun test test/*.test.ts test/{plugins,util}/*.test.ts",
     "watch": "npm link && tsc --watch",
     "prebuild": "pnpm run generate-plugin-defs && node rmdir.js dist",
@@ -93,6 +93,7 @@
     "@wdio/types": "^9.20.0",
     "codeclimate-types": "^0.3.1",
     "glob-bin": "^1.0.0",
+    "tsx": "^4.21.0",
     "typescript": "^5.8.3"
   },
   "engines": {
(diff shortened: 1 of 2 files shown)


webpro:
We could leave in `tsx` to run `.ts` test files and... restore `.js` to build with ts 5.0.4?

wojtekmaj:
Hint on node -e: https://bsky.app/profile/wojtekmaj.pl/post/3mf5mhsxros2v

wojtekmaj:
> We could leave in `tsx` to run `.ts` test files and... restore `.js` to build with ts 5.0.4?

So, like... Rolling back all the changes? Yeah that's what I was afraid of when thinking about Node.js 18 and thought we might wanna wait and just try a PoC :D

webpro:
No I mean only like I did in the TS 5.0.4 pipeline, it's all green now

webpro:
Oh wait dumb oversight on my end. Would still publish with .ts everywhere 🙈

webpro:
Looks like `rewriteRelativeImportExtensions` properly rewrites both file extensions and import specifiers (to build using `tsc` and publish). I think we're good to go actually. WDYT? Am I missing something?

wojtekmaj:
> Looks like `rewriteRelativeImportExtensions` properly rewrites both file extensions and import specifiers (to build using `tsc` and publish). I think we're good to go actually. WDYT? Am I missing something?

Yeah, this option will indeed do exactly what we want. I just need to resolve conflicts carefully...

webpro:
Let's go! Thanks a bunch.

webpro:

PR #1552, opened 2026-02-20
Fix incorrect coverage provider being reported

rexxars:
If you have a vitest config that sets a coverage provider to `istanbul`, and a script has a `--coverage` flag, knip incorrectly reports a missing `@vitest/coverage-v8` (as well as `@vitest/coverage-istanbul`).

This PR changes this behavior:
- If a vitest config has a `coverage` config and `enabled` is not set to `false`, _or_ you've defined a `provider` property, use the configured coverage provider (or v8 as fallback). This is slightly cheating as we are _assuming_ that you'll be wanting to use coverage if you've defined an explicit provider
- If passing a `--coverage.provider` flag, add that provider as a used dependency - otherwise, leave it up to config to determine

### DIFF
--- a/packages/knip/src/plugins/vitest/index.ts
+++ b/packages/knip/src/plugins/vitest/index.ts
@@ -38,8 +38,8 @@ const findConfigDependencies = (localConfig: ViteConfig, options: PluginOptions,
       : [];
   const reporters = getExternalReporters(testConfig.reporters);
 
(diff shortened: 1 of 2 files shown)


webpro:
That seems more correct indeed, thank you! 👍

webpro:

PR #1557, opened 2026-02-22
Add Qwik plugin

azat-io:
Add a new `qwik` plugin for Knip (Qwik/Qwik City).

It supports standard Qwik entries/routes, custom `srcDir`, `routesDir` (including `string[]`), Qwik City virtual modules, and `.md/.mdx` route files (including imports inside `.mdx` via the MDX compiler).

References:

- https://github.com/QwikDev/qwik
- https://qwik.dev/docs/routing/

### DIFF
--- a/packages/knip/src/plugins/qwik/index.ts
+++ b/packages/knip/src/plugins/qwik/index.ts
@@ -2,15 +2,16 @@ import MDX from '../../compilers/mdx.ts';
 import type { IsPluginEnabled, Plugin, RegisterCompilers, ResolveFromAST } from '../../types/config.ts';
 import { type Input, toEntry, toIgnore, toProductionEntry } from '../../util/input.ts';
 import { hasDependency } from '../../util/plugin.ts';
+import { config } from '../vite/index.ts';
 import { getRoutesDirs, getSrcDir } from './resolveFromAST.ts';
+
+// https://qwik.dev/docs/project-structure/
 
 const title = 'Qwik';
 
 const enablers = ['@builder.io/qwik'];
 
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
-
-const config = ['vite.config.{js,mjs,ts,cjs,mts,cts}'];
 
 const entry = ['src/entry.dev.tsx'];
 


packages/knip/src/plugins/qwik/index.ts
```diff
@@ -0,0 +1,61 @@
...
+import { hasDependency } from '../../util/plugin.ts';
+import { getRoutesDirs, getSrcDir } from './resolveFromAST.ts';
+
+const title = 'Qwik';
+
+const enablers = ['@builder.io/qwik'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['vite.config.{js,mjs,ts,cjs,mts,cts}'];
```
webpro: Please import/reuse from vite plugin

packages/knip/src/plugins/qwik/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  return [
+    ...entry.map(setSrcDir).map(path => toEntry(path)),
+    ...production.map(setSrcDir).map(path => toProductionEntry(path)),
+    ...routeEntries,
+    toIgnore('@qwik-city-plan', 'unlisted'),
+    toIgnore('@qwik-city-sw-register', 'unlisted'),
+    toIgnore('@qwik-client-manifest', 'unlisted'),
```
webpro, line 41: Why ignore these?

If they're an exception/problematic I'd like to:

- link to the source of the problem (qwik docs or source code)
- if you didn't already, try `toDependency` with `optional: true` first

azat-io: These are virtual modules generated by Qwik's Vite plugins at build time - they don't exist as npm packages and should never be in `package.json`, so `toDependency` with `optional: true` wouldn't be appropriate here.

They're registered in resolveId/load hooks and generate code on the fly (routes, SW script, client manifest). Same pattern as $app/$env/$service-worker in SvelteKit.

Sources:

- `@qwik-city-plan` - https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/src/buildtime/vite/plugin.ts#L325
- `@qwik-city-sw-register` - https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/src/buildtime/vite/plugin.ts#L328
- `@qwik-client-manifest` - https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/optimizer/src/plugins/plugin.ts#L1128

packages/knip/src/plugins/qwik/index.ts
azat-io: Done ✅

packages/knip/src/plugins/qwik/index.ts
webpro: Thanks for clarification, all good then.

Any chance you know of a good public repo I could give it a spin on, just for a quick smoke test?

azat-io: I just used:

```
pnpm create qwik@latest
```

I also tested it on my own project. But it has a non-standard configuration that requires creating and configuring a Knip config:
https://github.com/azat-io/eyecons

webpro: Works well on the qwik playground app. And it's slightly odd but a config like this does seem to do the job on eyecons:

```json
{
  "entry": ["extension/index.ts"],
  "qwik": {
    "config": ["vite.config.documentation.ts"]
  }
}
```

WDYT? Good to go?

azat-io: Yes, I think that's normal for a non-standard configuration. 👍

webpro: Thank you! Let's go ⚡

webpro:

PR #1559, opened 2026-02-23
Fix Astro sharpImageService() false positive for unused sharp

azat-io:
This PR fixes a false positive in the Astro plugin when using `sharpImageService()` in `astro.config.*`.

Astro requires installing `sharp` when `sharpImageService()` is configured, but Knip did not detect that usage and could report `sharp` as an unused dependency. 

The Astro plugin now detects `sharpImageService()` usage in Astro config files (via AST) and marks `sharp` as a referenced dependency.

### DIFF
--- a/packages/knip/src/plugins/astro/resolveFromAST.ts
+++ b/packages/knip/src/plugins/astro/resolveFromAST.ts
@@ -31,34 +31,14 @@ export const getSrcDir = (sourceFile: ts.SourceFile): string => {
 };
 
 export const usesSharpImageService = (sourceFile: ts.SourceFile) => {
-  const sharpImageServiceNames = new Set<string>();
-
   for (const statement of sourceFile.statements) {
-    if (ts.isImportDeclaration(statement) && ts.isStringLiteral(statement.moduleSpecifier)) {
-      if (statement.moduleSpecifier.text !== ASTRO_CONFIG_SPECIFIER) continue;
-      const importClause = statement.importClause;
-      if (!importClause) continue;
-
-      if (importClause.namedBindings && ts.isNamedImports(importClause.namedBindings)) {
-        for (const element of importClause.namedBindings.elements) {
-          if (
-            element.propertyName?.text === 'sharpImageService' ||
-            (!element.propertyName && element.name.text === 'sharpImageService')
-          ) {
-            sharpImageServiceNames.add(element.name.text);
-          }
-        }
-      }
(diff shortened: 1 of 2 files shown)


packages/knip/src/plugins/astro/resolveFromAST.ts
```diff
@@ -27,3 +29,36 @@ export const getSrcDir = (sourceFile: ts.SourceFile): string => {
...
+          ) {
+            sharpImageServiceNames.add(element.name.text);
+          }
+        }
+      }
+    }
+  }
+
+  function visit(node: ts.Node): boolean {
+    if (ts.isCallExpression(node)) {
```
webpro: Nitpick: I think we don't even need to look for the call expression. The fact that `sharpImageService` is imported should be enough.

packages/knip/test/plugins/astro-sharp-image-service.test.ts
```diff
@@ -0,0 +1,14 @@
...
+import { createOptions } from '../helpers/create-options.ts';
+import { resolve } from '../helpers/resolve.ts';
+
+const cwd = resolve('fixtures/plugins/astro-sharp-image-service');
+
+test('Find no unused sharp dependency when Astro sharpImageService is configured', async () => {
+  const options = await createOptions({ cwd });
+  const { issues } = await main(options);
+
+  assert(!issues.dependencies['package.json']?.['sharp']);
```
webpro: Please keep the `counters` in. Use default values and `dependencies: 0` is automatically included and we don't even need this assertion.

packages/knip/src/plugins/astro/resolveFromAST.ts
azat-io: Done ✅️️️️️️️

packages/knip/test/plugins/astro-sharp-image-service.test.ts
azat-io: Done ✅️️️️️️️

webpro:
Thanks, Azat! Truly appreciate all the efforts 🙏

webpro:

controversial:
Is the sharp image service not the default, even when it’s unspecified in config? See https://docs.astro.build/en/reference/errors/missing-sharp/

I think `sharp` is still a false-positive when users have _no_ configured value of `image.service` (even if they didn’t explicitly configure `sharpImageService`)

webpro:
Agreed, pushed a patch. @azat-io do you see any issues with this implementation, that basically reverses the logic?

PR #1561, opened 2026-02-23
Fix Vite plugin to respect root option for index.html entries

azat-io:
In one of my projects, Vite uses the `root` option, so `index.html` is not located at the project root. Because of this, Knip did not detect entries referenced from `index.html` and incorrectly reported related files/dependencies as unused.

This PR makes Knip respect Vite’s `root` option when resolving `index.html` entries, and adds a test case for a custom root setup.

### DIFF
--- a/packages/knip/src/plugins/vitest/index.ts
+++ b/packages/knip/src/plugins/vitest/index.ts
@@ -4,8 +4,9 @@ import type { Args } from '../../types/args.ts';
 import type { IsPluginEnabled, Plugin, PluginOptions, ResolveConfig } from '../../types/config.ts';
 import { _glob } from '../../util/glob.ts';
 import { type Input, toAlias, toConfig, toDeferResolve, toDependency, toEntry } from '../../util/input.ts';
-import { isAbsolute, isInternal, join, toPosix } from '../../util/path.ts';
+import { isAbsolute, isInternal, join, toAbsolute, toPosix } from '../../util/path.ts';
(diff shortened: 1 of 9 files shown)


packages/knip/src/plugins/vitest/index.ts
```diff
@@ -146,7 +147,15 @@ export const resolveConfig: ResolveConfig<ViteConfigOrFn | VitestWorkspaceConfig
     }
   };
 
+  const seenRoots = new Set<string>();
+
   for (const cfg of configs) {
+    const viteRoot = toAbsolute(cfg.root ?? '.', options.cwd);
+    if (!seenRoots.has(viteRoot)) {
+      seenRoots.add(viteRoot);
+      for (const entry of await getIndexHtmlEntries(viteRoot)) inputs.add(entry);
```
webpro, line 156: If we do `getIndexHtmlEntries` here, can we remove `resolve` from the Vite plugin? As we're executing `resolveConfig` from both plugins already.

azat-io: `resolveConfig` only runs when a config file exists. Without resolve, config-less Vite projects (like the `vite3` fixture) would lose `index.html` entry detection.

webpro: Good point! Thanks.

webpro:
Thanks Azat! ✨

webpro:

PR #1569, opened 2026-02-24
Tailwind-4 css config

### LINKED ISSUE #1551: 🐛 unused tailwind.css file
### Prerequisites

### Reproduction url

https://stackblitz.com/edit/github-pnh13q2n?file=package.json

### Reproduction access

### Description of the issue

With Tailwind 4, the config is written in CSS.
The CSS file (`tailwind.css` for example) don't need to be referenced explicitly. The tool (such as `@tailwindcss/vite`) will detect it automatically.

Also, even when it is referenced explicitly in `package.json`:

```jsonc
{
  "scripts": {
    "build:css": "tailwindcss -i tailwind.css -o styles.css"
  },
  "devDependencies": {
    "tailwindcss": "*"
  }
}
```

It is still not detected.

`knip` reports 3 errors:

```sh
Unused files (1)
tailwind.css
Unused devDependencies (1)
tailwindcss  package.json:16:6
Unlisted binaries (1)
tailwindcss  package.json
```

unional:
closes #1551

tailwind css config can be any css. So need to include all css files.

### DIFF
--- a/packages/knip/src/plugins/tailwind/index.ts
+++ b/packages/knip/src/plugins/tailwind/index.ts
@@ -11,7 +11,7 @@ const enablers = ['tailwindcss'];
 
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
-const entry = ['tailwind.config.{js,cjs,mjs,ts}'];
+const entry = ['tailwind.config.{js,cjs,mjs,ts}', '*.css'];
 
 const registerCompilers: RegisterCompilers = ({ registerCompiler, hasDependency }) => {
   if (hasDependency('tailwindcss')) registerCompiler({ extension: '.css', compiler });
--- a/packages/knip/fixtures/tailwind-4/any.css
+++ b/packages/knip/fixtures/tailwind-4/any.css
@@ -1,0 +1,1 @@
+@import 'tailwindcss' layer(base);
--- a/packages/knip/fixtures/tailwind-4/package.json
+++ b/packages/knip/fixtures/tailwind-4/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "@fixtures/tailwind-4",
+  "dependencies": {
+    "tailwindcss": "^4.0.3"
+  }
+}
--- a/packages/knip/fixtures/tailwind-4/tailwind.css
+++ b/packages/knip/fixtures/tailwind-4/tailwind.css
@@ -1,0 +1,1 @@
+@import 'tailwindcss';


packages/knip/src/plugins/tailwind/index.ts
```diff
@@ -11,7 +11,7 @@ const enablers = ['tailwindcss'];
 
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
-const entry = ['tailwind.config.{js,cjs,mjs,ts}'];
+const entry = ['tailwind.config.{js,cjs,mjs,ts}', '*.css'];
```
webpro, line 14: If we would add all `.css` files as entry points, we can no longer figure out if there's an unused css file.

Ideally, there is a default entry point, often this is resolved in `resolveConfig` function. But the tailwind plugin doesn't have this yet.

If it can't be resolved, such as if it is referenced only in some `index.html` file, then users can configure it like so:

```json
{
  "tailwind": {
    "entry": ["./styles.css"]
  }
}
```

That should then result in all other css files begin resolved through imports, plugins, etc. by the [tailwind "compiler"](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/tailwind/compiler.ts#L14-L15).

If you think there is a good and generic way to find the entry file(s) then we could work on that.

webpro (commented):
Thanks for the PR! There are better ways to deal with this, so I've added a comment.

packages/knip/src/plugins/tailwind/index.ts
unional: Yes, likely user config is needed.

It seems like my assumption of how Tailwind 4 handles the config is wrong. However, the conclusion seems to be still the same: yes, we can't reliably detect the config file. So a board `*.css` or user config is needed.

In [their doc](https://tailwindcss.com/docs/installation/using-vite):

- The html in Vite example reference the config css `/src/style.css'. So likely in Vite it process the html file and able to locate where the config is. (however it could be complicate outside as some path resolution to map `/src/style.css` to `./src/style.css` is needed)
- The PostCSS example is not doing any reference, so probably it scans the code (so no reference here)
- Tailwind CLI usage can be detect in script or `package.json`. So the reference may or may not be there.

For Storybook usage, we do need to reference the config file in `.storybook/preview.tsx`:

```tsx
import '../tailwind.css'
```

So technically that can be detected from reference/usage.

webpro: Tailwind might do a lot of work to find CSS (entry) files and I'm not sure whether Knip should re-implement all that. For instance, Vite looks in HTML file for JS entry: https://github.com/webpro-nl/knip/blob/4e95ffb45748fa1ae84548c1992d2947826d2667/packages/knip/src/plugins/vite/index.ts#L25-L27 but it's regex-based and we shouldn't do more of this tbh.

For the Storybook preview case: if `.css` compiler is enabled then `import '../tailwind.css'` should add the `.css` file and the files it resolves/imports.

unional: Yes. So I think your approach is the right one. Unfortunately user needs to provide that knowledge through config.

Maybe can shortcut that a bit with some defaults like `tailwind.css` or `tailwind*.css`.

webpro: Yes, if `tailwind.css` is a common entry we could add it. E.g. if it's in the official documentation or official tailwind app/boilerplate init/generator (so we can link to it in source code for future reference).

webpro, line 14: Most setups will import from JS/TS files directly, and don't need any entry points configured by Knip. However, we could still do something like this:

```suggestion
const entry = ['tailwind.config.{js,cjs,mjs,ts}'];

const production = ['{,src/}{global{,s},style{,s},index}.css'];
```

Feels like `tailwind.css` isn't a widely used default?

unional: Not sure how widely used `tailwind.css` is, I can just speak for myself. I do that as well as using some conventions to create multiple of them. Here is an example: https://github.com/justland/just-web-foundation/tree/main/libs/toolkits/.storybook

But those are referenced in code, so `knip` should be able to detect them through normal means.

The case it wouldn't be detected in if it is used manually. e.g. export as part of the package for consumption.

But we can justify that is a relatively rare case and ask user to specify them in the knip config.

webpro: I feel like we could close this PR, especially since it does not even solve your specific issue, right?

unional: Sounds good. Seems like user config is the way to go.

Thanks for your help! 🍻

PR #1570, opened 2026-02-24
chore: add `k` script for easy testing

unional:
Adding this to make it easier to run fixture test:

```shell
pnpm k --directory packages/knip/fixture/tailwind-4

// or with @antfu/ni
nr k --directory packages/knip/fixture/tailwind-4
```

### DIFF
--- a/.github/DEVELOPMENT.md
+++ b/.github/DEVELOPMENT.md
@@ -92,6 +92,9 @@ Run Knip without compilation:
 
 ```shell
 node path/to/knip/packages/knip/src/cli.ts
+
+// or using the alias script
+pnpm k
 ```
 
 #### Alias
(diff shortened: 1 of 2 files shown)


webpro:
Happy to add this but I'm using an alias for this:

```
❯ alias k
alias k='node --inspect ~/p/knip/knip/packages/knip/src/cli.ts'
```

Now I have `k` anywhere. Isn't there `alias` or an equivalent on your system?

Also see https://github.com/webpro-nl/knip/blob/main/.github/DEVELOPMENT.md#alias

webpro:
Now that we can use `node src/cli.ts` from anywhere I don't think this script is very useful anymore, as it only works from the knip source directory (and something like `pnpm --dir [dir]` is just unnecessary overhead).

unional:
sure, no problem. Just a suggestion as this is attached to the repo so new contributor can benefit from it and don't need to add the alias in their rc file.

PR #1523, opened 2026-02-11
Fix Yarn (Berry) plugins and binaries reported as unused

wojtekmaj:
Fixes #1505
Fixes #1522

Closes #1509

This PR adds support for `plugins` and `yarnPath` that may be referenced in `.yarnrc.yml`. Thanks to this change, we no longer raise a false positive for `.yarn/*` while also avoiding false _negatives_ that #1509 would have introduced by completely ignoring `.yarn` directory.

> [!NOTE]
> This PR was created with AI assistance (GPT 5.3 Codex).

PR #1574, opened 2026-02-25
Detect Yarn plugins that are listed by their path alone

robintown:
In a .yarnrc.yml file, plugins can be listed either as a dict or simply as a string. In the latter case, Yarn interprets the string as the plugin's path. https://github.com/webpro-nl/knip/pull/1523 did not yet account for these kinds of plugin entries, causing the plugin files to still be flagged as unused.

([Here's](https://github.com/element-hq/element-call/actions/runs/22407328480/job/64870758723?pr=3712) where I ran into this issue in the real world)


webpro:
Thanks Robin! ✨

webpro:

PR #1578, opened 2026-02-28
Detect cross-package imports by nearest package boundary

azat-io:
Fix false positive unused dependencies for internal workspace packages when the dependency key differs from the target package name, or when Knip runs from a leaf package.

Aliased workspace dependencies were missed because usage detection relied on discovered workspace package names, which aren't available when running from a leaf workspace.

Detect cross-package imports by comparing the resolved file's nearest package boundary against the current workspace directory.


webpro:
> Fix false positive unused dependencies for internal workspace packages when the dependency key differs from the target package name

This might've been an issue with the fixture? I fixed this.

> or when Knip runs from a leaf package.

This isn't supported. And I'd rather not add too much overhead trying to.

webpro:
What problem does it solve? Or can we close?

azat-io:
Thanks, you're right.

After rebasing on latest `main`, the original "dependency key differs from target package name" part appears to have been fixed by fbee3426ccc05eaf2fcaa069826cd3594f946db9.

PR #1580, opened 2026-03-01
feat: add WXT plugin

adelin-b:
## Summary

- Adds a new plugin for [WXT](https://wxt.dev/) (Web Extension Framework)
- Resolves `modules` array entries in `wxt.config.ts` as dependencies
- Defines `entrypoints/**/*` as entry file patterns
- Prevents false positive "unused dependency" reports for WXT modules (e.g. `@wxt-dev/module-react`, `wxt-module-console-forward`)

## Motivation

WXT uses a `modules` array in `wxt.config.ts` similar to Nuxt's module system. Without this plugin, knip reports these modules as unused devDependencies because they're only referenced in the WXT config and never directly imported in source code.

## Implementation

Follows the same pattern as the Nuxt plugin:
- `setup()` defines a global `defineConfig` helper so knip can load the config
- `resolveConfig()` extracts string entries from the `modules` array as dependencies
- Entry patterns match WXT's conventional `entrypoints/` directory

## Test plan

- [x] Plugin test passes (`node --test test/plugins/wxt.test.ts`)
- [x] Correctly identifies unused dependencies not in `modules`
- [x] Does not flag modules listed in `wxt.config.ts` as unused

### DIFF
--- a/packages/knip/src/plugins/wxt/index.ts
+++ b/packages/knip/src/plugins/wxt/index.ts
@@ -1,0 +1,52 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.ts';
+import type { Input } from '../../util/input.ts';
+import { toDependency } from '../../util/input.ts';
+import { hasDependency } from '../../util/plugin.ts';
+import type { WxtConfig } from './types.ts';
+
+// https://wxt.dev/guide/configuration.html
+
+const title = 'WXT';
+
+const enablers = ['wxt'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['wxt.config.{js,mjs,ts}'];
+
+const entry = ['entrypoints/**/*.{ts,tsx,js,jsx,html}'];
+
+const production = ['entrypoints/**/*.{ts,tsx,js,jsx,html}'];
+
+const setup = async () => {
+  if (globalThis && !('defineConfig' in globalThis)) {
+    Object.defineProperty(globalThis, 'defineConfig', {
+      value: (id: any) => id,
+      writable: true,
+      configurable: true,
+    });
+  }
+};
+
+const resolveConfig: ResolveConfig<WxtConfig> = async localConfig => {
+  const deps =
+    localConfig?.modules?.reduce<Input[]>((acc, id) => {
+      if (typeof id === 'string') acc.push(toDependency(id));
+      return acc;
+    }, []) ?? [];
+
+  return deps;
+};
+
+const plugin: Plugin = {
+  title,
+  enablers,
+  isEnabled,
+  config,
+  entry,
+  production,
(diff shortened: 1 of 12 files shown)


packages/knip/src/plugins/wxt/index.ts
```diff
@@ -0,0 +1,52 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.ts';
+import type { Input } from '../../util/input.ts';
+import { toDependency } from '../../util/input.ts';
+import { hasDependency } from '../../util/plugin.ts';
+import type { WxtConfig } from './types.ts';
+
+// https://wxt.dev/guide/configuration.html
```
webpro, line 7: 404

packages/knip/src/plugins/wxt/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const enablers = ['wxt'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['wxt.config.{js,mjs,ts}'];
+
+const entry = ['entrypoints/**/*.{ts,tsx,js,jsx,html}'];
+
+const production = ['entrypoints/**/*.{ts,tsx,js,jsx,html}'];
```
webpro, line 19: Production entries are `entry` points in default mode as well:

```suggestion
const production = ['entrypoints/**/*'];
```

This glob should target **all** files, right?

https://wxt.dev/guide/essentials/entrypoints.html

packages/knip/src/plugins/wxt/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+  }
+};
+
+const resolveConfig: ResolveConfig<WxtConfig> = async localConfig => {
+  const deps =
+    localConfig?.modules?.reduce<Input[]>((acc, id) => {
+      if (typeof id === 'string') acc.push(toDependency(id));
+      return acc;
+    }, []) ?? [];
```
webpro, line 36: This is a bit hard to read. Perhaps a `for..of` iteration can do the job.

packages/knip/test/plugins/wxt.test.ts
```diff
@@ -0,0 +1,23 @@
...
+test('Find dependencies with the wxt plugin', async () => {
+  const options = await createOptions({ cwd });
+  const { issues, counters } = await main(options);
+
+  assert(issues.dependencies['package.json']['unused-module']);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    files: 1,
+    dependencies: 1,
```
webpro, line 19: Ideally there are no unused files, as both the config file and the entry point should be consumed by the plugin.

webpro (commented):
Thank you! Happy to merge, left a few remarks though.

webpro:
We can close this now that #1703 has been merged.

PR #1584, opened 2026-03-02
feat: add AI agent skill for Knip usage guidance

Rasaboun:
## Summary

- Adds a ready-to-use AI agent skill (`skill/`) with structured instructions for AI coding agents (GitHub Copilot, Claude Code, Cursor, etc.)
- Covers setup, configuration, result interpretation, false positive handling, and safe auto-fixing
- Includes config templates for common setups (Next.js, Expo, Astro, monorepos) in `skill/references/`
- Updates README with an "AI Agent Skill" section documenting installation via `npx skills add` and manual copy

### DIFF
--- a/packages/knip/README.md
+++ b/packages/knip/README.md
@@ -26,6 +26,34 @@ performance, less maintenance and easier refactorings.
 - Follow [@webpro.nl on Bluesky][13] for updates
 - [Sponsor Knip!][14]
 
+## AI Agent Skill
+
+This repository includes a ready-to-use [agent skill](https://docs.github.com/en/copilot/concepts/agents/about-agent-skills)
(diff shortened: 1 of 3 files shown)


Rasaboun:
@webpro Thanks for taking a look at this but why PR was closed, could you share the reason ?

webpro:
It's mostly generated, isn't it? If you or the agent doesn't bother to first understand Knip itself, then how can you or the agent explain it using a skill?

Did you see the MCP Server? See https://github.com/webpro-nl/knip/blob/main/packages/mcp-server/src/texts.js specifically for better instructions.

Rasaboun:
You're right the initial version was largely generated as a starting point

It's based on something I’m using in my own project, but I didn't align it enough with Knip itself
I'll go through the MCP and rethink it with that in mind

Thanks

PR #1589, opened 2026-03-05
Use module resolution for package manifest loading

### LINKED ISSUE #1590: Package manifest resolution fails in Yarn PnP environments
## Description

`loadPackageManifest` in `packages/knip/src/manifest/helpers.ts` uses hardcoded `node_modules` paths to resolve package manifests:

```ts
_require(join(dir, 'node_modules', packageName, 'package.json'))
```

This fails in Yarn PnP environments where there is no `node_modules` directory, causing peer dependency metadata (`bin`, `peerDependencies`, etc.) to not be read. As a result, binaries are reported as unlisted and dependencies as unused incorrectly.

The original code even had a TODO comment acknowledging uncertainty about the approach:
> *"Not sure what's the most efficient way to get a package.json, but this seems to do the job across package managers (npm, Yarn, pnpm)"*

## Proposed fix

Replace hardcoded `node_modules` path lookups with `createRequire` from `node:module`, which uses Node.js standard module resolution. Yarn PnP hooks into this resolution, so it works without `node_modules`.

I have a working implementation and will open a PR shortly.

sekyungk:
Closes #1590

## Summary

- `loadPackageManifest` previously constructed absolute paths with hardcoded `node_modules` and passed them to `createRequire()` directly — bypassing Node.js module resolution entirely
- Replaced with `createRequire().resolve()` so that Node.js standard module resolution handles package lookup, which works across all package managers including Yarn PnP
- The original code had a TODO comment: *"Not sure what's the most efficient way to get a package.json, but this seems to do the job across package managers (npm, Yarn, pnpm)"* — however it did not work in Yarn PnP environments since there is no `node_modules` directory

## Test plan

- [x] All 578 smoke tests pass (`pnpm run test:bun:smoke`)
- [x] TypeScript type check passes (`tsc --noEmit`)
- [x] Verified manually on pnpm monorepo and Yarn PnP project
- No test changes needed

### DIFF
--- a/packages/knip/src/manifest/helpers.ts
+++ b/packages/knip/src/manifest/helpers.ts
@@ -1,16 +1,33 @@
-import { createRequire } from 'node:module';
+import Module from 'node:module';
 import type { Scripts } from '../types/package-json.ts';
 import { join } from '../util/path.ts';
+import { _require } from '../util/require.ts';
 
 type LoadPackageManifestOptions = { dir: string; packageName: string; cwd: string };
 
 export const loadPackageManifest = ({ dir, packageName, cwd }: LoadPackageManifestOptions) => {
-  try {
-    const _require = createRequire(join(dir, 'package.json'));
-    const manifestPath = _require.resolve(join(packageName, 'package.json'));
-    if (manifestPath.startsWith(cwd)) return _require(manifestPath);
-  } catch (_error) {
-    // Explicitly suppressing errors here
+  if (Module.findPackageJSON) {
+    try {
+      const manifestPath = Module.findPackageJSON(packageName, join(dir, 'package.json'));
+      if (manifestPath?.startsWith(cwd)) return _require(manifestPath);
+    } catch (_error) {
+      // Explicitly suppressing errors here
+    }
+  } else {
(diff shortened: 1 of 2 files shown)


webpro:
Thanks for the PR. There was an issue in main I've fixed. But still most ecosystem tests seem to fail with this solution. Any chance you know what that could be?

sekyungk:
@webpro

Thank you for the review!

The previous approach used `createRequire().resolve()` , which fails with `ERR_PACKAGE_PATH_NOT_EXPORTED` for packages that have an exports field without a ./package.json entry. I've pushed a new commit that uses Module. findPackageJSON (Node >=22.14.0) instead, which bypasses the exports restriction. For Node versions below 22.14.0, it falls back to the original node_modules path approach.

Apologies if anything is unclear — English is not my first language.

sekyungk:
Closing this PR. After further testing, I found that Module. findPackageJSON does not work in Yarn PnP environments.
This means the approach in this PR doesn't actually solve the original issue.
I apologize for not testing this properly before opening the PR.  I'll look into alternative approaches.

PR #1598, opened 2026-03-08
Add oxlint.config.ts to oxlint plugin

DaniFoldi:
Hi :wave:,

Oxlint now supports `oxlint.config.ts` as a configuration file (https://oxc.rs/docs/guide/usage/linter/config.html), so I've added that to the list of config entrypoints.

In reality, it complains if both the json and the ts files are present, but since it's a stub here it should be fine for the fixture.


webpro:
Thanks Dániel!

PR #1599, opened 2026-03-08
Add oxfmt plugin

DaniFoldi:
Hi :wave:,

Similar to oxlint, here's a simple plugin for oxfmt.

I have to say, this repo has one of the best agent preparations, 5.4 was able to do this pretty much on its own, the only input it needed was the supported configuration formats! (And if it messed up and I didn't see something, let me know and I'm happy to fix it but I had to give it a try, and looked all good to me 😃)


packages/knip/src/plugins/oxfmt/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// https://oxc.rs/docs/guide/usage/formatter/config.html
+
+const title = 'Oxfmt';
+
+const enablers = ['oxfmt'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['.oxfmtrc.json', '.oxfmtrc.jsonc'];
```
webpro, line 12: Just saying, `config` entries are normally accompanied by `resolveConfig` and the oxfmt configuration object becomes the first argument. If there isn't anything custom to resolve, they should be `entry` items. However, since they're JSON files they're not actually used analyzed at all (hence the zeroes in the test assertion counters), and it makes sense to keep it as it is, because once there _is_ something interesting we can just add `resolveConfig`.

packages/knip/fixtures/plugins/oxfmt/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@plugins/oxfmt",
+  "private": true,
+  "scripts": {
+    "format": "oxfmt",
+    "format:custom": "oxfmt --config .oxfmtrc-1.jsonc"
```
webpro, line 6: ✨

webpro:
Thanks for the kind words! And this PR!

Issue #1565, opened 2026-02-23
💡 Enable metro plugin when expo is installed


DaniFoldi:
### Suggest an idea for Knip

Hi :wave:,

Expo is a complicated framework, because (unlike others, eg. next or astro which rely on build tools and wrap their configuration) it uses babel/metro config files. Expo depends on `@expo/metro`, which depends on all the metro tools, but that's not important here.

A project does not need metro in its package.json directly installed, and thus the metro plugin is not activated, and the metro config file is reported as unused. Reference: https://docs.expo.dev/guides/customizing-metro/
Babel is in a similar situation, although in all of the RN/expo projects I've seen where babel.config.js was explicitly added (neither is required, but for customization it's supported), `@babel/transform` and others were directly depended upon, so it's not as useful to infer in my opinion.

I was looking at the existing expo and metro plugins, and my current best idea is to add `expo` to the enablers of the metro plugin - but it's possible that there is a better way, hence this issue instead of a PR. Let me know if this is the right approach, or I'd missed a way to cross-reference plugins, happy to add it.

webpro:
That could work.

Just last week we've managed to split up React Native from Metro plugin so at least things are organized slightly better today. The matter sounds pretty similar to e.g. #1528 and #1538.

Probably a good way to go about this is to create a new minimal fixture directory that represents the real-world scenario that should be supported (see other related fixtures/tests to get the idea). Or maybe adjust an existing one if that makes sense. Make the change and make the new test pass... 🚦

Typically plugins have their own "enablers", and sometimes e.g. `entry` or `config` or `resolveConfig` from another plugin is referenced/used (e.g. vite just uses vitest `resolveConfig`).

webpro:
Closing this with #1600 merged.

PR #1600, opened 2026-03-08
Enable metro plugin when expo is installed

DaniFoldi:
Hi :wave:,

As discussed in #1565, this change enables the metro plugin when expo is detected, and adds a new minimal fixture for this setup.

Also, I saw a tweet recently that the next version of metro would gain support for .ts config files, so I updated the config file detection to support all the formats in https://github.com/facebook/metro/blob/main/packages/metro-config/src/__tests__/loadConfig-test.js - not all projects are going to update immediately (especially expo projects where the version is internally managed), but I assume the number of projects with an existing metro.config.ts file where this would temporarily cause a false negative is probably near 0.

### DIFF
--- a/packages/knip/test/plugins/expo4.test.ts
+++ b/packages/knip/test/plugins/expo4.test.ts
@@ -1,7 +1,6 @@
 import assert from 'node:assert/strict';
 import test from 'node:test';
 import { main } from '../../src/index.ts';
-import { join } from '../../src/util/path.ts';
 import baseCounters from '../helpers/baseCounters.ts';
 import { createOptions } from '../helpers/create-options.ts';
 import { resolve } from '../helpers/resolve.ts';
@@ -10,9 +9,7 @@ const cwd = resolve('fixtures/plugins/expo4');
 
 test('Find Metro config in Expo project without direct Metro dependency', async () => {
   const options = await createOptions({ cwd });
-  const { issues, counters } = await main(options);
-
-  assert(counters.files === 0);
+  const { counters } = await main(options);
 
   assert.deepEqual(counters, {
     ...baseCounters,
--- a/packages/knip/fixtures/plugins/expo4/metro.config.js
+++ b/packages/knip/fixtures/plugins/expo4/metro.config.js
@@ -1,1 +1,5 @@
-module.exports = {};
+module.exports = {
+  transformer: {
+    babelTransformerPath: 'react-native-svg-transformer',
+  },
+};
--- a/packages/knip/fixtures/plugins/expo4/node_modules/react-native-svg-transformer/package.json
+++ b/packages/knip/fixtures/plugins/expo4/node_modules/react-native-svg-transformer/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "react-native-svg-transformer"
+}
--- a/packages/knip/fixtures/plugins/expo4/package.json
+++ b/packages/knip/fixtures/plugins/expo4/package.json
@@ -7,5 +7,8 @@
     "expo": "*",
     "react": "*",
     "react-native": "*"
+  },
+  "devDependencies": {
+    "react-native-svg-transformer": "*"
   }
 }


packages/knip/test/plugins/expo4.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import { createOptions } from '../helpers/create-options.ts';
+import { resolve } from '../helpers/resolve.ts';
+
+const cwd = resolve('fixtures/plugins/expo4');
+
+test('Find Metro config in Expo project without direct Metro dependency', async () => {
+  const options = await createOptions({ cwd });
+  const { issues, counters } = await main(options);
+
+  assert(!issues.files.has(join(cwd, 'metro.config.js')));
```
webpro: This is fine, but `counters.files === 0` so we know there are no unused files.

packages/knip/src/plugins/metro/index.ts
```diff
@@ -9,11 +9,15 @@ import type { MetroConfig } from './types.ts';
...
+const enablers = ['metro', '@react-native/metro-config', 'expo'];
 
 const isEnabled: IsPluginEnabled = options => hasDependency(options.dependencies, enablers);
 
-const config = ['metro.config.{js,cjs,json}', 'package.json'];
+const config = [
+  'metro.config.{js,cjs,mjs,ts,cts,mts,json}',
+  '.config/metro.{js,cjs,mjs,ts,cts,mts,json}',
+  'package.json',
+];
```
webpro, line 20: https://github.com/facebook/metro/blob/ac169d231fa56e88d9409e5997e38b9d0906b78d/packages/metro-config/src/loadConfig.js#L51-L58 👍

packages/knip/src/plugins/metro/index.ts
```diff
@@ -9,11 +9,15 @@ import type { MetroConfig } from './types.ts';
 
 const title = 'Metro';
 
-const enablers = ['metro', '@react-native/metro-config'];
+const enablers = ['metro', '@react-native/metro-config', 'expo'];
```
webpro, line 12: It's bit hard for me to predict the impact of this from this PR alone. Are there maybe public repo/examples that benefit from this, or other useful resources/docs?

packages/knip/test/plugins/expo4.test.ts
DaniFoldi: Good point, will fix

packages/knip/src/plugins/metro/index.ts
DaniFoldi: Hmm, for us, the reason for having this file is `react-native-svg-transformer`, which requires a custom transform being added to metro.

One approach is looking for projects which have a metro config and it contains an expo import https://github.com/search?q=path%3Ametro.config.ts%20%22expo%2Fmetro-config%22&type=code or look for package.json files that have expo and the transformer in their dependencies - this one is easy to add knip to: https://github.com/search?q=path%3Apackage.json+%22%5C%22expo%5C%22%22+%22%5C%22react-native-svg-transformer%5C%22%22+%22%5C%22knip%5C%22%22&type=code

From here, probably a manual knip run to see if it's reported as unused, or a config check to see if it's ignored. What do you think? Happy to do this if you think it's a good approach

webpro: yes that could work, the point is mostly to at least have some confidence this isn't a "local fix" for a single project; apologies the overhead here

having some related work merged in, maybe @wojtekmaj has an opinion?

wojtekmaj: This makes sense to me.

packages/knip/test/plugins/expo4.test.ts
DaniFoldi: Fixed in the latest commit

packages/knip/src/plugins/metro/index.ts
DaniFoldi: Hey @webpro, anything I can help with to get this over the line?

webpro: Thought you had some suggestions. Anyway I'll try to take a look soon. Any input and repro steps are welcome.

webpro:
Thank you! Rebased, tried it on some repos and results LGTM.

webpro:
:rocket: _This pull request is included in v6.1.0. See [Release 6.1.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.1.0) for release notes._

PR #1605, opened 2026-03-11
Add 'ng' to the list of IGNORED_GLOBAL_BINARIES

PatrykKuniczak:
I've added `ng` of Angular to the list of IGNORED_GLOBAL_BINARIES, to avoid warning for that.

### DIFF
--- a/packages/knip/src/constants.ts
+++ b/packages/knip/src/constants.ts
@@ -145,6 +145,7 @@ export const IGNORED_GLOBAL_BINARIES = new Set([
   'yarn',
   'yes',
   'zip',
+  'ng'
 ]);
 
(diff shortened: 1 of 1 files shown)


webpro:
Thanks!

webpro:
Going to revert this, last publish of `ng` is `7/27/2018`, but if Angular CLI is installed globally, add it to `ignoreBinaries`. Otherwise, this short-circuits the projects that do have it locally installed.

PatrykKuniczak:
@webpro Yeah, i've added it, but it seems meaningful to add it here, instead of in ignoreBinaries for each Angular project.
You can't install `ng` standalone, as external package, it's included in `@angular/cli`.

But anyway it's preffered way to install:
https://github.com/angular/angular-cli?tab=readme-ov-file#setting-up-a-project

PR #1613, opened 2026-03-14
fix: Use runtime FS case-sensitivity check in cspell and taskfile tests

Zamiell:
I cloned the knip repo and ran `pnpm test` and it just immediately fails.
The reason is because knip assumes that if you are on Windows, you are on a case insensitive file system.
But this is not true - it is best practice to [set your "repositories" directory to be case sensitive](https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity).

Thus, this PR replaces `process.platform === 'win32'` checks with a helper that tests actual filesystem case-sensitivity at runtime.

Claude helped me write this.


webpro:
Appreciate the PR, but should not be an issue after merging #1462.

PR #1625, opened 2026-03-18
feat: add vite-plus plugin

max-ae:
This PR adds a plugin for [Vite+](https://viteplus.dev) which bundles Vite, Vitest, tsdown, Oxfmt and Oxlint.
When moving to Vite+ by adding it to a package.json and removing the individual dependencies, Knip reports several new issues because it is not loading the configuration from the individual packages anymore.
This PR attempts to fix that by adding a dedicated plugin to the enablers of the individual plugins.

### DIFF
--- a/packages/knip/fixtures/plugins/vite-plus/package.json
+++ b/packages/knip/fixtures/plugins/vite-plus/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "@plugins/vite-plus",
+  "devDependencies": {
+    "vite-plus": "*"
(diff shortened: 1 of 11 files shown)


webpro:
No description and no clear value I could derive from the fixture.

max-ae:
@webpro sorry, the CLI seems to have eaten my PR description when submitting, updated again!

webpro:
Any chance you could extend the fixture? Currently it's not showing off much.

PR #1638, opened 2026-03-24
fix(nx): fall back to workspace cwd when run-commands target has no cwd

### LINKED ISSUE #1637: Nx plugin crashes when nx:run-commands target has no options.cwd
## Description

The Nx plugin crashes with `TypeError [ERR_INVALID_ARG_TYPE]` when a `project.json` contains `nx:run-commands` targets that don't specify `options.cwd`. This is the common case — most projects use `nx:run-commands` with just `options.command`.

## Reproduction

**knip version:** 6.0.4

`project.json`:
```json
{
  "name": "bff",
  "sourceRoot": "apps/bff/src",
  "projectType": "application",
  "targets": {
    "build": {
      "executor": "nx:run-commands",
      "options": {
        "command": "bun build apps/bff/src/index.ts --outdir apps/bff/dist --target bun"
      }
    },
    "serve": {
      "executor": "nx:run-commands",
      "options": {
        "command": "bun run apps/bff/src/index.ts"
      }
    }
  }
}
```

Run `knip` in a monorepo with this project.json.

## Error

```
TypeError [ERR_INVALID_ARG_TYPE]: The "from" argument must be of type string. Received undefined
    at Object.relative (node:path:1357:5)
    at relative (knip/dist/util/path.js:18:25)
    at getDependenciesFromScript (knip/dist/binaries/bash-parser.js:140:72)
    at knip/dist/binaries/index.js:6:47
    at Array.flatMap (<anonymous>)
    at getInputsFromScripts (knip/dist/binaries/index.js:6:29)
    at Object.getInputsFromScripts (knip/dist/WorkspaceWorker.js:199:90)
    at knip/dist/plugins/nx/index.js:49:24
    at Array.flatMap (<anonymous>)
    at Object.resolveConfig (knip/dist/plugins/nx/index.js:40:10)
```

## Root Cause

In `dist/plugins/nx/index.js:48`:

```js
const cwd = target.options?.cwd ? join(options.cwd, target.options.cwd) : undefined;
```

When `target.options.cwd` is not set (which is the default for `nx:run-commands`), `cwd` becomes `undefined`. This is passed through `getInputsFromScripts` → `getDependenciesFromScript` → and eventually into `path.relative(options.cwd, ...)` inside the bash parser's error handler at `bash-parser.js:140`, which crashes because `options.cwd` is `undefined`.

## Fix

The fallback should be `options.cwd` (the project root) instead of `undefined`:

```diff
- const cwd = target.options?.cwd ? join(options.cwd, target.options.cwd) : undefined;
+ const cwd = target.options?.cwd ? join(options.cwd, target.options.cwd) : options.cwd;
```

This ensures the bash parser always has a valid `cwd` to resolve paths against, falling back to the workspace root when no explicit `cwd` is configured on the target.

## Workaround

Using `bun patch` / `patch-package` to apply the one-line fix above.

liorp:
## Summary

Fixes #1637

When an `nx:run-commands` target does not specify `options.cwd` (which is the common case), the Nx plugin passes `cwd: undefined` to `getInputsFromScripts`. Because the spread in `WorkspaceWorker` (`{ ...baseOptions, ...options }`) includes undefined-valued properties, this overrides the valid `baseOptions.cwd`, causing a crash in the bash parser's error handler — `path.relative(undefined, ...)` throws `ERR_INVALID_ARG_TYPE`.

### Root cause

```js
// packages/knip/src/plugins/nx/index.ts:68
const cwd = target.options?.cwd ? join(options.cwd, target.options.cwd) : undefined;
//                                                                        ^^^^^^^^^ overrides baseOptions.cwd
```

### Fix

Fall back to `options.cwd` (the workspace root) instead of `undefined`:

```diff
- const cwd = target.options?.cwd ? join(options.cwd, target.options.cwd) : undefined;
+ const cwd = target.options?.cwd ? join(options.cwd, target.options.cwd) : options.cwd;
```

### Tests

Added `test/plugins/nx-run-commands-no-cwd.test.ts` with two test cases:

1. **No cwd on target** — verifies `getInputsFromScripts` receives the workspace `cwd` (not `undefined`)
2. **Explicit cwd on target** — verifies `cwd` is properly joined with the workspace root

Both tests are verified to fail without the fix and pass with it. Existing Nx plugin tests continue to pass.

### DIFF
--- a/packages/knip/fixtures/plugins/nx/libs/b/project.json
+++ b/packages/knip/fixtures/plugins/nx/libs/b/project.json
@@ -62,12 +62,6 @@
         "command": "tsc -p tsconfig.build.json",
         "cwd": "libs/b"
       }
-    },
-    "serve": {
-      "executor": "nx:run-commands",
-      "options": {
-        "command": "bun run apps/b/server.ts"
-      }
     }
   }
 }


webpro:
Why add a different type of test? Please see other (Nx) tests. Running `main()` is almost entirely end-to-end. Please keep it concise and consistent if at all possible.

liorp:
Removed the redundant `serve` target from the fixture — the existing fixture already has multiple `nx:run-commands` targets without `options.cwd` (`webpack`, `ls-project-root`, `generate-docs`, `lint`), so adding another one didn't exercise any new code path.

The fix itself is a one-line defensive correctness change: `cwd` should never be `undefined` when passed to `getInputsFromScripts`, since it gets spread over `baseOptions` in `WorkspaceWorker` and overrides the valid `cwd`. The specific crash (in the bash parser's error handler via `path.relative(undefined, ...)`) is hard to reproduce in an e2e test because `unbash` rarely throws parse errors, but the fix ensures correct behavior regardless of downstream code paths.

webpro:
You're right, let's just roll with it 🛼

liorp:
Thanks. Waiting on your approval 🙏

webpro:
:rocket: _This pull request is included in v6.0.5. See [Release 6.0.5](https://github.com/webpro-nl/knip/releases/tag/knip%406.0.5) for release notes._

webpro:
Thanks Lior!

PR #1645, opened 2026-03-26
Fix peer dependency resolution

controversial:
Solves an issue where `sass` was reported as a false-positive “unused dependency” when it’s used as an optional peer dependency of the new `@tsdown/css`, which is itself an optional peer dependency of `tsdown`—`knip` (intentionally) doesn’t normally resolve optional peer dependencies as “used” more than one level deep.

Disclaimer: I used Claude Code to author much of this PR but I believe the solution to be sound/sensical and tested, and it solves a problem I faced adopting `knip`. if this PR violates the [contributing guidelines](https://arc.net/l/quote/gkiefkex) feel free to close it.

### DIFF
--- a/packages/knip/src/plugins/tsdown/index.ts
+++ b/packages/knip/src/plugins/tsdown/index.ts
@@ -1,12 +1,5 @@
-import type {
-  IsLoadConfig,
-  IsPluginEnabled,
-  Plugin,
-  Resolve,
-  ResolveConfig,
-  ResolveFromAST,
-} from '../../types/config.ts';
-import { toDependency, toProductionEntry } from '../../util/input.ts';
+import type { IsLoadConfig, IsPluginEnabled, Plugin, ResolveConfig, ResolveFromAST } from '../../types/config.ts';
+import { toProductionEntry } from '../../util/input.ts';
 import { hasDependency } from '../../util/plugin.ts';
 import { getEntryFromAST } from './resolveFromAST.ts';
 import type { Entry, TsdownConfig } from './types.ts';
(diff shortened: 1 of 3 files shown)


packages/knip/src/plugins/tsdown/index.ts
```diff
@@ -47,6 +47,14 @@ const resolveFromAST: ResolveFromAST = program => {
   return [...entries].map(id => toProductionEntry(id, { allowIncludeExports: true }));
 };
 
+const resolve: Resolve = ({ manifest }) => {
+  const deps = { ...manifest.dependencies, ...manifest.devDependencies, ...manifest.peerDependencies };
+  if (!('@tsdown/css' in deps)) return [];
+  return ['postcss', 'postcss-import', 'postcss-modules', 'sass', 'sass-embedded']
```
controversial: I would have liked to implement this in a way that is more future-proof to changes in the `@tsdown/css` optional peer deps, but I don’t know a great way to do so without modifying the core resolution `isReferencedDependency` logic, whose current design I think is intentional.

I don’t think it would be appropriate to try to resolve `@tsdown/css` at runtime to determine its optional peer deps, because it might not be importable by `knip` for a variety of reasons (eg `pnpm` strict package isolation)

webpro:
Thanks for raising this! Dove in and looks like that `isReferencedDependency` logic was flawed perhaps. So I moved the solution to core. Any chance you could give this a shot in your project(s)?

controversial:
@webpro Sorry for the slow response here - confirmed this branch from pkg.pr.new solves the issue in my project!

webpro:
:rocket: _This pull request is included in v6.3.1. See [Release 6.3.1](https://github.com/webpro-nl/knip/releases/tag/knip%406.3.1) for release notes._

webpro:
Thanks for confirming!

PR #1647, opened 2026-03-27
feat: support YAML package manifests

al:
## Summary

- Add support for `package.yaml` and `package.yml` manifests in addition to `package.json`.
- Load the root manifest from JSON or YAML and improve config-related errors to reference the actual manifest file.
- Extend workspace manifest discovery/parsing to include YAML, with deterministic preference for `package.json` when both formats exist.
- Update dependency location peeking to recognise YAML dependency sections and safely match package names containing regex-special characters.

## Why

Some projects use YAML manifests instead of JSON. Knip previously assumed `package.json` only, which caused manifest discovery and dependency diagnostics to fail in those setups.

> [!NOTE]
> Support for `package.yaml` seems conspicuous by its absence, but I couldn't see any discussion in Issues or PRs. Perhaps it's against the philosophy or maybe I've missed a substantial blocker in my, frankly quite superficial, analysis of this?

## What Changed

- `packages/knip/src/util/fs.ts`
  - Added `findManifestFile()` for `package.json` / `package.yaml` / `package.yml`.
  - Added `loadManifest()` to parse JSON or YAML by extension.
- `packages/knip/src/util/create-options.ts`
  - Switched root manifest loading to `findManifestFile()` + `loadManifest()`.
  - Updated error messages when manifest/config isn’t found.
- `packages/knip/src/util/map-workspaces.ts`
  - Added YAML manifest glob patterns.
  - Parse workspace manifests as JSON or YAML.
  - Group matches by workspace directory and prefer `package.json` if both formats exist.
  - Fixed root manifest directory normalisation.
- `packages/knip/src/PackagePeeker.ts`
  - Added YAML section detection for `dependencies`, `devDependencies`, and `optionalPeerDependencies`.
  - Added escaped YAML key matching for package names (avoids regex false matches).
- Tests
  - Added `packages/knip/test/util/map-workspaces.test.ts`
    - root workspace mapping with `package.yaml`
    - preference of `package.json` over YAML in same workspace
  - Added `packages/knip/test/util/package-peeker.test.ts`
    - YAML dependency location detection
    - package names with regex-special characters

## Test Plan

- [x] `pnpm install`
- [x] `pnpm -C packages/knip run build`
- [x] `pnpm -C packages/knip run test:node`
- [x] `pnpm -C packages/knip exec tsx --test test/util/map-workspaces.test.ts test/util/package-peeker.test.ts`
- [x] Manual validation in a consumer project using YAML manifest.

### DIFF
--- a/packages/docs/src/content/docs/features/monorepos-and-workspaces.md
+++ b/packages/docs/src/content/docs/features/monorepos-and-workspaces.md
@@ -8,7 +8,8 @@ Workspaces are handled out-of-the-box by Knip.
 
 Workspaces are sometimes also referred to as package-based monorepos, or as
 packages in a monorepo. Knip uses the term workspace exclusively to indicate a
-directory that has a `package.json`.
+directory that has a package manifest (`package.json`, `package.yaml` or
+`package.yml`).
 
 ## Configuration
 
@@ -66,11 +67,15 @@ The `workspaces` in Knip configuration (4) not already defined in the root
 
 :::caution
 
-A workspace must have a `package.json` file.
+A workspace must have a package manifest file (`package.json`, `package.yaml`
+or `package.yml`).
+
+If multiple manifest files are present in the same workspace, Knip prefers
+`package.json`.
 
 :::
 
-For projects with only a root `package.json`, please see [integrated
+For projects with only a root package manifest, please see [integrated
 monorepos][2].
 
 ## Additional workspaces
--- a/packages/docs/src/content/docs/reference/faq.md
+++ b/packages/docs/src/content/docs/reference/faq.md
@@ -300,9 +300,10 @@ dependencies found in webpack configuration, handled by Knip's webpack plugin.
 
 ### What's the difference between workspaces, projects and programs?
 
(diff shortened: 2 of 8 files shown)


webpro:
> Some projects use YAML manifests instead of JSON.

Can you give examples? My first instinct says this isn't all that common..

al:
No, I guess you're right. AFAIK only PNPM supports package.yaml and I don't know what the adoption is.

If you won't accept built-in support, would you consider a PR that refactored the current implementation to extract the package handling logic to a ManifestAdapter, and exposed some kind of hook or option that allowed users to provide their own implementation?

webpro:
> AFAIK only PNPM supports package.yaml

TIL!

> If you won't accept built-in support, would you consider a PR that refactored the current implementation to extract the package handling logic to a ManifestAdapter, and exposed some kind of hook or option that allowed users to provide their own implementation?

This is an idea I'm not too fond of: exposing more interfaces eventually means more work for me (including maintenance, potential breaking changes, whatnot). There is already so many funky edges to keep track of.

PR #1650, opened 2026-03-30
Handle absolute Vitest test roots

filipw01:
This fixes Vitest configs that set `test.root` to an absolute path.

For example, this shape was not handled correctly:

```ts
import { join } from 'node:path';
import { fileURLToPath } from 'node:url';
import { defineConfig } from 'vitest/config';

const root = join(fileURLToPath(new URL('.', import.meta.url)), 'tests');

export default defineConfig({
  test: {
    root,
    include: ['*.test.ts'],
    setupFiles: ['./setup.ts'],
  },
});
```

Knip resolved `test.root` as if it were always relative to the config file directory, so an already-absolute root ended up being joined again and produced invalid paths.

This change uses `toAbsolute(...)` for Vitest `test.root` handling in:
- config dependency resolution
- test entry path resolution
- build lib entry resolution

It also adds a regression fixture covering an absolute `test.root`.

### DIFF
--- a/packages/knip/fixtures/plugins/vitest10/node_modules/vitest/dist/config.js
+++ b/packages/knip/fixtures/plugins/vitest10/node_modules/vitest/dist/config.js
@@ -1,5 +1,0 @@
-function defineConfig(config) {
-  return config;
-}
-
-export { defineConfig };


webpro:
Thanks Filip!

webpro:
:rocket: _This pull request is included in v6.1.1. See [Release 6.1.1](https://github.com/webpro-nl/knip/releases/tag/knip%406.1.1) for release notes._

PR #1657, opened 2026-03-31
Deduplicate module graph queue to avoid redundant file processing

MatissJanis:
This is a doozy..

We want to run knip v6 on a HUGE repository, but there was a problem: duplicate queue items piled on causing used files to be reported as unused.

The fix is relatively simple.. do not push duplicate items to the queue. But I also added a unit test to catch this if it regresses back.

Applying this patch allows us to run Knip v6 and catch the _expected_ amount of unused-files.


webpro:
Thanks for catching _and_ fixing this!

Looking into it, we might be able to get rid of the problem by using only the `visited` set. This is more of a structural fix by getting rid of `queue` altogether, also removing the need for mocking the `walkAndAnalyze` as well. Does that work for you?

packages/knip/test/queue-dedup.test.ts
```diff
@@ -1,32 +1,20 @@
...
   const options = await createOptions({ cwd });
   const { issues, counters } = await main(options);
 
-  ProjectPrincipal.prototype.walkAndAnalyze = orig;
-
   assert(!('leaf.ts' in issues.files));
   assert(!('shared-a.ts' in issues.files));
   assert(!('shared-b.ts' in issues.files));
 
-  assert.equal(capturedQueueLength, 5, `Queue should have 5 unique entries, got ${capturedQueueLength}`);
```
MatissJanis: The updated patch is now running knip for our codebase. I'll revert back with results in ~40m.

In the meantime: this assertion removal will make the unit test evergreen. It will pass on current main branch (we don't want that) and also on this feature branch (that is OK).

webpro: Gotcha. Yet the thing is: the problem has been fixed in a structural way. There is no more `queue` or something that is not a `Set` to leak from mocked function. The fact that we're only using a `Set` now is the fix. I didn't see a way to catch this in a failing test.

MatissJanis: That's fine! In that case I would recommend removing the unit test entirely. :)

webpro: Fair enough. Didn't want to remove most of your work and, like, remove the credits.

MatissJanis: Ah.. I don't mind as long as we get the problem solved & knip runnable in DataDog repos

MatissJanis:
LGTM! Works as expected in our codebase

webpro:
:rocket: _This pull request is included in v6.2.0. See [Release 6.2.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.2.0) for release notes._

PR #1660, opened 2026-04-02
Workers: improving unused file identification

MatissJanis:
Knip is working quite well in our mega repo, but it still finds a bunch of false-negative "unused files". One of those instances is: workers.

So with this PR I am improving how Knip identifies Web Worker files - to reduce the amount of false-negative signals.

### DIFF
--- a/packages/knip/src/typescript/visitors/calls.ts
+++ b/packages/knip/src/typescript/visitors/calls.ts
@@ -97,7 +97,14 @@ export function handleCallExpression(node: CallExpression, s: WalkState) {
   ) {
     const specifier = getStringValue(node.arguments[0])!;
     if (specifier) {
-      s.addImport(specifier, undefined, undefined, undefined, node.arguments[0].start, IMPORT_FLAGS.ENTRY);
+      s.addImport(
+        specifier,
+        undefined,
+        undefined,
+        undefined,
+        node.arguments[0].start,
+        IMPORT_FLAGS.ENTRY | IMPORT_FLAGS.OPTIONAL
+      );
       return;
     }
   }
@@ -159,7 +166,14 @@ export function handleNewExpression(node: NewExpression, s: WalkState) {
   ) {
     const specifier = getStringValue(node.arguments[0])!;
     if (specifier) {
-      s.addImport(specifier, undefined, undefined, undefined, node.arguments[0].start, IMPORT_FLAGS.ENTRY);
+      s.addImport(
+        specifier,
+        undefined,
+        undefined,
+        undefined,
+        node.arguments[0].start,
+        IMPORT_FLAGS.ENTRY | IMPORT_FLAGS.OPTIONAL
+      );
       return;
     }
   }


MatissJanis:
I suspect the 2x failing tests are just flaky and would work if re-ran.

webpro:
Thanks for the pull request. I think I get what you're after, but we should not try to resolve everything dynamically from static analysis. Let's start out more conservatively by including only safe patterns, then we don't need the post-processing either.

Example:

- Fixture: https://github.com/webpro-nl/knip/blob/f08681a2659de9d6c99c3f0e7950682069b60db1/packages/knip/fixtures/module-register/index.ts
- Visitor: https://github.com/webpro-nl/knip/blob/d8c016164fda981209186c205f86329b522e76ee/packages/knip/src/typescript/visitors/calls.ts#L70-L90

The idea in this example is to only accept `register()` call expressions that have exactly `import.meta.url` as second argument and skip everything else. And only if `node:module` is imported.

MatissJanis:
Ok, I have significantly simplified the logic now. No more following variables and catching the `path(__dirname, './name.js')` variant.

webpro:
Thanks for simplifying! A few remaining concerns — I think we need to tighten this further to match only patterns that are statically provable.

### `new Worker(stringLiteral)` / `new WorkerThread(stringLiteral)`

- `WorkerThread` isn't a standard API (Node.js `worker_threads` exports `Worker`, not `WorkerThread`) — matching it would be guessing.
- `new Worker(stringLiteral)` without a base-URL anchor is too loose — `Worker` could be any user-defined class, and even with a real `worker_threads` Worker, the path resolves from `cwd`, not the current file.
- The safe pattern `new Worker(new URL('./w.js', import.meta.url))` is **already handled** by the existing `new URL()` visitor — no new code needed.

### `fork(stringLiteral)`

- The `hasChildProcessImport` gate is good, but `fork('./x.js')` resolves from `cwd`, not from the calling file. Without a `__dirname` or `import.meta.url` anchor there's no static guarantee the specifier points to a project file.
- The real-world safe patterns (`fork(path.join(__dirname, './x.js'))`, `fork(new URL('./x.js', import.meta.url))`) are compound expressions, not string literals — so they wouldn't match `isStringLiteral(node.arguments[0])` anyway.

### `*.fork(stringLiteral)` (MemberExpression branch)

This branch isn't gated by `hasChildProcessImport`, so any `obj.fork('./path.js')` anywhere would match.

### Bottom line

Applying the same whitelisting principle as `register()` (which requires `import.meta.url` as second arg): none of the new handlers have a safe-enough gate. The one truly safe Worker pattern (`new URL` + `import.meta.url`) is already covered.

If there are specific real-world patterns from your mega-repo that we're missing, sharing those would help figure out what's safe to add.

MatissJanis:
Sure, here are some code samples of the patterns we observe:

```
// 1
  new Worker(path.join(__dirname, 'file.js'), { ... });

// 2
  const workerPath = path.resolve(__dirname, './worker.js');
  const worker = fork(workerPath, args, { env: process.env });

// 3
  import { Worker as WorkerThread } from 'node:worker_threads';
  this.workerMain = path.join(__dirname, 'worker_thread/worker.js');
  // Usage:
  const nodeWorker = new WorkerThread(this.workerMain, { ... });
  const childWorker = childProcess.fork(this.workerMain, { ... });

// 4
  const WORKER_SCRIPT = path.resolve(__dirname, 'workers/typecheck-worker.js');
  const child = child_process.fork(WORKER_SCRIPT, [], { ... });

// 5
  const workerPath = path.join(__dirname, 'hash-worker.js');
  const worker = new Worker(workerPath, { ... });

```

I would really love to help improve Knip upstream to reduce the amount of false-negative signals of "unused" files. Our other alternative is to either maintain a long list of manually defined entrypoints (yuck..) or a local patch of knip (also yuck..).

webpro:
Only strictly *inline* patterns (similar to the `import.meta.url` gating) could be considered, e.g.:

- `path.join(__dirname, ...literals)` / `path.resolve(__dirname, ...literals)`
  - As first arg of `new Worker(...)` gated on `node:worker_threads` import
  - As first arg of `fork(...)` / `spawn(...)` / `execFile(...)` gated on `node:child_process` import

That would cover the first pattern from your list.

MatissJanis:
Any recommendations what to do with the other patterns?

In DataDog we have a monorepo with >17m LOC. Manually finding all these cases has proven to be quite a counter-productive task.

webpro:
> Any recommendations what to do with the other patterns?

I'd argue these aren't quite false positives — the files genuinely aren't statically reachable. Patterns 2–5 all cross a variable boundary, and following bindings plus evaluating `path.join` across scope is interpretation, not static analysis. I'd rather not pull Knip toward resolving arbitrary dynamic patterns that way.

Closing this PR, while planning to add the inline patterns as I mentioned.

webpro:
> planning to add the inline patterns as I mentioned.

This has been introduced in https://github.com/webpro-nl/knip/releases/tag/knip%406.4.0

PR #1662, opened 2026-04-03
fix: unlisted binaries bug with execa

Zamiell:
Knip has a false positive with the following pattern:

```ts
const command = 'npm';
await $`${command} publish --access=public --ignore-scripts`;
```

With this code, Knip thinks that the "publish" binary is not listed.

This PR fixes the bug. (I didn't bother opening an issue in favor of just fixing it.)

### DIFF
--- a/packages/knip/fixtures/script-visitors-execa-variable-cmd/package.json
+++ b/packages/knip/fixtures/script-visitors-execa-variable-cmd/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "@fixtures/script-visitors-execa-variable-cmd",
+  "scripts": {
+    "start": "node ./script.mjs"
+  },
+  "devDependencies": {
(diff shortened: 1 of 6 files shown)


webpro:
Thanks for the pull request. However, this is expected behavior of a static analysis tool. Some patterns could perhaps be resolved relatively safely still, but currently we're not after that (not sure if we even should, really).

For reference, it's similar to #1660, see https://github.com/webpro-nl/knip/pull/1660#issuecomment-4178386348 specifically.

Zamiell:
@webpro Ok! However, I am worried about the vast number of false positives though in the new patch that introduced this template inspection behavior.

What about this: if Knip sees any variable inside of a execa template string, it just bails entirely.
That way, Knip would have have false negatives instead of false positives, which seems a bit better.
Would you consider that?

PR #1663, opened 2026-04-03
docs: flesh out windows explanation

Zamiell:
self-explanatory PR


webpro:
Like to keep this doc lean i.e. tbh it's noise for most contributors and the root cause is already there, enough for a contributor to search and resolve on their own platform.

PR #1664, opened 2026-04-03
chore: change license file to have more conventional casing

Zamiell:
The license file is supposed to be capitalized, as documented here: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository


webpro:
Thanks!

PR #1671, opened 2026-04-08
Add Panda CSS plugin

Faithfinder:
## Summary

- Adds a new `panda-css` plugin that activates when `@pandacss/dev` is in dependencies
- Detects `panda.config.{ts,js,mjs,cjs,mts,cts}` as config files
- Resolves string presets from the config as dependencies (e.g. `@pandacss/preset-panda`, `@park-ui/panda-preset`)
- Marks `postcss` as used since `@pandacss/dev` brings it transitively via `@pandacss/dev/postcss`

## Test plan

- [x] Added fixture at `fixtures/plugins/panda-css/` with panda config containing two presets
- [x] Test verifies both presets are detected as unlisted dependencies
- [x] Existing plugin tests (postcss, unocss, tailwind, babel) still pass

### DIFF
--- a/packages/knip/fixtures/plugins/panda-css/node_modules/@pandacss/dev/index.js
+++ b/packages/knip/fixtures/plugins/panda-css/node_modules/@pandacss/dev/index.js
@@ -1,0 +1,1 @@
+export const defineConfig = id => id;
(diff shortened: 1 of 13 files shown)


webpro:
Thanks!

Issue #1672, opened 2026-04-10
💡 Flag unused none js/ts files


pdeveltere:
### Suggest an idea for Knip

Hello knip team,

in our codebase, we fetch our big MySQL query templates with `imports`:
```
import exampleTemplate from './example.template' with { type: 'text' };

function async getUser() {
   return db().raw(exampleTemplate);
}
```

we would love to have a mechanism, which also detects unused templates. so when an import is removed, it's flagged by knip that the template file is unused.

I thought this would work by adding the following to our `knip.json`:
```
{
  "project": [
    "*.template"
  ],
  "include": [
    "files",
  ],
  "rules": {
    "files": "error",
  }
}
```

but this is not working at the moment.. would it be possible to add support for this? 🙏🏼

PR #1673, opened 2026-04-10
feat: add configurable foreignFileExtensions option

pdeveltere:
## Summary

This PR adds a new configuration option `foreignFileExtensions` that allows users to customize which file extensions should be treated as foreign files (non-source files that don't trigger unresolved import warnings).

## Problem

Previously, the list of foreign file extensions was hardcoded in `FOREIGN_FILE_EXTENSIONS` constant. Users couldn't customize this list for their specific project needs.

## Solution

- Added `foreignFileExtensions` configuration field to the schema with proper Zod validation
- Updated TypeScript types to include the new configuration option
- Modified `ConfigurationChief` to handle the new option with sensible defaults
- Updated `graph/build.ts` and `SourceFileManager.ts` to use configurable extensions
- Maintained full backward compatibility - existing projects work unchanged

## Usage

Users can now configure foreign file extensions in their `knip.json`:

```json
{
  "foreignFileExtensions": [".css", ".scss", ".png", ".jpg", ".custom"]
}
```

## Testing

- Added comprehensive tests for both custom and default configurations
- Verified backward compatibility with existing behavior
- All existing tests continue to pass
- Manual testing confirms expected behavior

## Backward Compatibility

✅ Existing projects work without any changes
✅ Default behavior unchanged
✅ No breaking changes to existing APIs

The default includes the same extensions as before: `.avif`, `.css`, `.eot`, `.gif`, `.html`, `.ico`, `.jpeg`, `.jpg`, `.less`, `.mp3`, `.png`, `.sass`, `.scss`, `.sh`, `.svg`, `.ttf`, `.webp`, `.woff`, `.woff2`, `.yaml`, `.yml`

### DIFF
--- a/packages/knip/fixtures/foreign-file-extensions-default/data.yaml
+++ b/packages/knip/fixtures/foreign-file-extensions-default/data.yaml
@@ -1,0 +1,1 @@
+yaml: content
--- a/packages/knip/fixtures/foreign-file-extensions-default/image.png
+++ b/packages/knip/fixtures/foreign-file-extensions-default/image.png
@@ -1,0 +1,1 @@
+PNG placeholder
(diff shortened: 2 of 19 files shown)


webpro:
Thanks for the pull request. Non-standard extensions shouldn't be/aren't added as demonstrated in #1672, because they're invalid JS/TS.

The solution is add a no-op compiler. Added a section to the docs: https://knip.dev/features/compilers#tracking-non-source-files

Also added config hints that should detect the issue early on.

PR #1681, opened 2026-04-13
fix(util): tolerate JSONC and array extends in findRootDirsBase

Hoffs:
## Problem

`findRootDirsBase` reads tsconfig.json with `JSON.parse(stripJsonComments(...))` — strict JSON with comments stripped, **not** JSONC.

`tsconfig.json` is officially [JSONC](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html): trailing commas are valid, accepted by `tsc`, IDEs, and by `parseTsconfig` from `get-tsconfig` (which the rest of this same loader uses). When a workspace tsconfig contains a trailing comma, `JSON.parse` throws, the error is silently swallowed by the surrounding `try/catch`, `findRootDirsBase` returns `undefined`, and the fallback `?? dir` resolves `rootDirs` against the workspace dir instead of the file that declared them.

The most visible symptom is **SvelteKit `./$types` imports being reported as unresolved** when the workspace tsconfig has a trailing comma — SvelteKit's generated `.svelte-kit/tsconfig.json` declares `rootDirs: ["..", "./types"]`, the user's tsconfig extends it, and with the wrong base dir the rootDirs-based virtual-sibling lookup for `$types.d.ts` fails.

## Fix

- Pass `{ trailingCommas: true }` to `stripJsonComments` — already a direct dep, no new deps.
- Also handle the array form of `compilerOptions.extends` (introduced in TS 5.0). The previous code did `join(dir, raw.extends)` on the raw value, which produces a nonsensical path when `extends` is an array.

## Test

New fixture `packages/knip/fixtures/tsconfig-rootdirs-jsonc/`:

- `tsconfig.json` — extends a base; contains a trailing comma in `compilerOptions` (valid JSONC).
- `shared/tsconfig.base.json` — declares `rootDirs: [".", "./types"]`.

Test asserts `rootDirs` are absolutized against `shared/` (where they were declared), not the workspace root.


webpro:
Thanks! Can you please make it use the central `load` or `loadJSON` function? That has `trailingCommas: true` as well.

Hoffs:
> Thanks! Can you please make it use the central `load` or `loadJSON` function? That has `trailingCommas: true` as well.

Good call! Ended up with `loadJSONC` as that uses stripping and `load` determines based on extension.

Hoffs:
sorrry, some changes got lost in selective hunk staging 😅

webpro:
Knip uses https://github.com/privatenumber/get-tsconfig for parsing tsconfig. Reading up on it, looks like it states that it supports `extends`. So I think it would be worth it trying to resolve this upstream (extend `resolveExtends` to also rewrite `rootDirs`/`typeRoots` (mapped) and `rootDir`), resulting in a much cleaner solution. Apologies for not spotting this right away. Any chance you could consider opening a PR against get-tsconfig?

Hoffs:
Introduced a PR in upstream package, ran tests with locally linked package, will need to wait for that one to be resolved then I guess.

Hoffs:
bumped the version, should be all clean

webpro:
Thanks Ignas!

This is great. The failing test reveals an interesting case, I think we still need the `loadJSONC` fallback. I'm happy to push a fix, but feel free to give it a spin from your end. Up to you!

Hoffs:
>  The failing test reveals an interesting case, I think we still need the `loadJSONC` fallback. I'm happy to push a fix, but feel free to give it a spin from your end. Up to you!

Seems like the fixtures were incomplete, so added a missing tsconfig. This matches the real-world astro module.

webpro:
Yeah I was a bit on the fence, if for some reason a transitive tsconfig is not found/errs then we could retry loading only the root one. Now it fails harder which is probably better.

Thanks a lot! You improved two projects at once ✨

webpro:

PR #1684, opened 2026-04-13
Add a plugin for react-email

xaqrox:
Adding a plugin to support [`react-email`](https://react.email/docs/introduction).

### DIFF
--- a/packages/knip/src/plugins/react-email/index.ts
+++ b/packages/knip/src/plugins/react-email/index.ts
@@ -1,5 +1,5 @@
-import parseArgs from 'minimist';
-import type { IsPluginEnabled, Plugin, Resolve } from '../../types/config.ts';
+import type { Args } from '../../types/args.ts';
+import type { IsPluginEnabled, Plugin } from '../../types/config.ts';
 import { toDependency, toEntry } from '../../util/input.ts';
 import { hasDependency } from '../../util/plugin.ts';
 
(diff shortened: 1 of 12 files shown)


packages/knip/src/plugins/react-email/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    const dir = parsed.dir.replace(/^\.\//, '');
+    if (dirs.has(dir)) continue;
+
+    dirs.add(dir);
+    inputs.push(toEntry(`${dir}/**/*.tsx`));
+  }
+
+  if (dirs.size === 0) inputs.push(toEntry('emails/**/*.tsx'));
+
+  return inputs;
```
webpro: There is no need for manual script/argument parsing, please implement `Plugin.args` (instead of `resolve`) for this: https://knip.dev/writing-a-plugin/argument-parsing (just added a note to [.agents/PLUGINS.md](https://github.com/webpro-nl/knip/blob/main/.agents/PLUGINS.md#script-parsing)).

packages/knip/src/plugins/react-email/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  enablers,
+  isEnabled,
+  resolve,
+};
+
+export default plugin;
+
+export const docs = {
+  note: 'Email templates that re-export a component as both a named and default export will trigger a `duplicates` issue. Add a [`@alias` JSDoc tag][1] to the default export to suppress it, or set `ignoreIssues` for the emails directory in your knip config.\n\n[1]: /reference/jsdoc-tsdoc-tags#alias',
+};
```
webpro: This is a generic feature of Knip, not specific to the plugin. Not sure if we should add it here. And if for some reason it should be here (e.g. if files are generated), it should also be reflected in the test assertions (i.e. not seeing `duplicates` counter)?

webpro (commented):
Thanks! Main issue I'm seeing is re. script parsing, any chance you could take a look?

packages/knip/src/plugins/react-email/index.ts
xaqrox: Nifty! I'll note that it was initially *very* unclear that "argument parsing" meant "arguments of scripts". Nor was it clear what the objects on that page represented. It might help to have an example in-context on the "Writing a Plugin" page to illustrate. Is there a plugin you'd recommend as a candidate for that example (for a future PR...)?

packages/knip/src/plugins/react-email/index.ts
xaqrox: Got it. These Email templates usually have both named and default exports of the same object... for some reason. In attempting to deal with that I manually added those files as entries, which doesn't cause the duplicates to be ignored. But it looks like adding entries via toEntry *does* ignore duplicates.

packages/knip/src/plugins/react-email/index.ts
webpro: Yeah you're still one of the first to use this `Plugin.args` API, thanks for the clean implementation still! I'll fix that up in the docs soon. Thanks for the feedback.

webpro:

PR #1695, opened 2026-04-21
fix(rslib): resolve entry points from rslib config

rpereira-anchor:
## The problem

The rslib plugin was a stub — `resolveConfig` always returned `[]` and the config glob was listed under `entry` (treating `rslib.config.ts` as a source entry itself) instead of `config` (a file to parse for entries). This meant knip never extracted source entry points from `rslib.config.ts`, so all source files in rslib-based packages were reported as unused files.

## The fix

Mirrors the rolldown and tsdown plugins:

- Change `entry` key to `config` so knip parses the file rather than treating it as a source entry
- Add `resolveFromAST` using `collectPropertyValues(program, 'entry')`, which walks the full AST and handles both the object-map form (`source.entry: { index: './src/index.ts' }`) and the array `defineConfig([...])` form

## Testing

Two new tests and an expanded fixture:
- Single config with multiple named entries (`source.entry: { index, browser }`)
- Array config form (`defineConfig([...])`) — verifies deduplication when two lib configs share an entry

Tested against a real monorepo (12 rslib packages): reduced false-positive "unused files" reports from 72 to 25.

All 256 plugin tests pass.

### DIFF
--- a/packages/knip/fixtures/plugins/rslib/rslib.config.ts
+++ b/packages/knip/fixtures/plugins/rslib/rslib.config.ts
@@ -1,3 +1,10 @@
 import { defineConfig } from '@rslib/core';
 
-export default defineConfig({});
+export default defineConfig({
+  source: {
+    entry: {
+      index: './src/index.ts',
+      browser: './src/browser.ts',
+    },
+  },
+});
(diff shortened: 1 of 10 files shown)


webpro:
Thanks Ricardo! 👌

webpro:
:rocket: _This pull request is included in v6.6.1. See [Release 6.6.1](https://github.com/webpro-nl/knip/releases/tag/knip%406.6.1) for release notes._

PR #1699, opened 2026-04-24
Ensure output is flushed on exit

joshkel:
The use of `process.exit` means that the Node.js runtime may immediately exit, without waiting for asynchronous `process.stdout.write` calls to finish. I noticed that, if I ran Knip and piped its output through `grep` or `less`, the output got truncated.


webpro:
Agreed, thanks Josh

webpro:
:rocket: _This pull request is included in v6.6.3. See [Release 6.6.3](https://github.com/webpro-nl/knip/releases/tag/knip%406.6.3) for release notes._

PR #1703, opened 2026-04-26
feat: add WXT plugin

sebastianbreguel:
## Summary

- Adds plugin for [WXT](https://wxt.dev/) (Web Extension Framework)
- Resolves `modules` array entries in `wxt.config.ts` as dependencies  
- Recognizes `entrypoints/**/*` as production entry files
- Prevents false positive "unused dependency" reports for WXT modules

## Context

Based on PR #1580 by @adelin-b, addressing @webpro's review feedback:
- Changed entry pattern to `entrypoints/**/*` to match all files (not just specific extensions)
- Replaced `reduce` with `for..of` for readability
- Fixed entry pattern processing by returning entries from `resolveConfig` (static `entry` array is skipped when plugin has `resolveConfig`)

## Test plan

- [x] Plugin test passes (`bun test test/plugins/wxt.test.ts`)
- [x] Correctly identifies unused dependencies not in `modules`
- [x] Does not flag modules listed in `wxt.config.ts` as unused
- [x] Entrypoint files are not flagged as unused

### DIFF
--- a/packages/knip/src/plugins/wxt/index.ts
+++ b/packages/knip/src/plugins/wxt/index.ts
@@ -1,0 +1,53 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.ts';
+import type { Input } from '../../util/input.ts';
+import { toDependency, toProductionEntry } from '../../util/input.ts';
+import { hasDependency } from '../../util/plugin.ts';
+import type { WxtConfig } from './types.ts';
+
+// https://wxt.dev/guide/essentials/config.html
+
+const title = 'WXT';
+
+const enablers = ['wxt'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['wxt.config.{js,mjs,ts}'];
+
+const production = ['entrypoints/**/*'];
+
+const setup = async () => {
+  if (globalThis && !('defineConfig' in globalThis)) {
+    Object.defineProperty(globalThis, 'defineConfig', {
+      value: (id: unknown) => id,
+      writable: true,
+      configurable: true,
+    });
+  }
+};
+
+const resolveConfig: ResolveConfig<WxtConfig> = async localConfig => {
+  const inputs: Input[] = [];
+
+  for (const pattern of production) {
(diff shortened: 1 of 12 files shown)


packages/knip/src/plugins/wxt/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config = ['wxt.config.{js,mjs,ts}'];
+
+const production = ['entrypoints/**/*'];
+
+const setup = async () => {
+  if (globalThis && !('defineConfig' in globalThis)) {
+    Object.defineProperty(globalThis, 'defineConfig', {
```
webpro: What needs this global function to be defined? At least in the fixture it's a regular import.

packages/knip/src/plugins/wxt/index.ts
```diff
@@ -0,0 +1,53 @@
+import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.ts';
+import type { Input } from '../../util/input.ts';
+import { toDependency, toProductionEntry } from '../../util/input.ts';
+import { hasDependency } from '../../util/plugin.ts';
+import type { WxtConfig } from './types.ts';
+
+// https://wxt.dev/guide/essentials/config.html
```
webpro: gives 404, how about this:

```suggestion
// https://wxt.dev/guide/essentials/entrypoints.html
```

webpro (commented):
Thank you! Two questions I have..

packages/knip/src/plugins/wxt/index.ts
sebastianbreguel: Fair point. The fixture imports defineConfig from wxt directly. If a user skips that import, their config breaks at runtime anyway, not Knip's job to patch it. Removed setup().

webpro:
Thanks Sebastian!

webpro:
:rocket: _This pull request is included in v6.8.0. See [Release 6.8.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.8.0) for release notes._

PR #1715, opened 2026-05-05
fix: type-only imports in monorepos

lishaduck:
In monorepos, Knip has begun to flag `type`-only `import`s as being unlisted. While I'm generally in favor of this change, Knip is clearly documented to have different behavior, so this patch should hopefully fix it.

Disclaimer: This patch was suggested by an LLM running in an agentic loop on the Flint repo. I typed the fix myself in hopes of convincing myself it was wrong, but from my limited reading of the code it seems reasonable. The repro was minimized from the Flint repository by an agent, and has been manually verified to be a) representive of the problem and b) fail previously & pass now.

### DIFF
--- a/packages/knip/fixtures/workspaces-type-only-dev-dependencies/knip.json
+++ b/packages/knip/fixtures/workspaces-type-only-dev-dependencies/knip.json
@@ -1,0 +1,8 @@
+{
+  "workspaces": {
+    "packages/*": {
+      "entry": "index.ts!",
+      "project": "index.ts!"
+    }
+  }
+}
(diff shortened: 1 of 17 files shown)


webpro:
Thanks for the catch and pull request! I've moved the condition and added coverage to improve handling of production types. If you're good with this I'm happy to merge.

lishaduck:
This seems to work for us, thank you!

webpro:
Thanks Eli!

webpro:
:rocket: _This pull request is included in v6.12.1. See [Release 6.12.1](https://github.com/webpro-nl/knip/releases/tag/knip%406.12.1) for release notes._

PR #1726, opened 2026-05-06
Fix Vercel config detection

### LINKED ISSUE #1725: vercel.ts is still reported as unused in workspace roots
## Reproduction

https://github.com/jakeleventhal/knip-vercel-ts-unused-repro

## Description

Issue #1717 was closed by #1720 and released in `knip@6.12.0`, but the fix only covers a Vercel config file in the root project. In a monorepo, `vercel.ts` files in workspace roots are still reported as unused files.

The reproduction now uses `knip@6.12.0` with a workspace app at `apps/web`:

```sh
npm install
npm run knip
```

Actual output:

```txt
Unused files (1)
apps/web/vercel.ts
```

Expected behavior: `apps/web/vercel.ts` should be treated as a Vercel project configuration file for the `apps/web` workspace, just like root-level `vercel.ts` is handled for the root project.

## Root cause

The Vercel plugin is marked `isRootOnly`, so it is skipped for non-root workspaces before its `isEnabled` check can see `vercel.ts` in the workspace directory.

## Validation

I have a branch that removes the root-only gate and adds a workspace fixture. The updated repro fails on published `knip@6.12.0` and passes when run with the branch source.

Related: #1717

jakeleventhal:
## Summary

- allow the Vercel plugin to enable from workspace roots instead of root-only projects
- add a workspace fixture covering `apps/web/vercel.ts`
- keep the existing root-level `vercel.ts` fixture passing

Fixes #1725.

## Validation

- `bun test packages/knip/test/plugins/vercel.test.ts`
- `node --test test/plugins/vercel.test.ts` from `packages/knip`
- `pnpm --dir packages/knip build`
- `node scripts/run-test.js --smoke --runtime node` from `packages/knip`
- `pnpm --dir packages/knip lint`
- published `knip@6.12.0` repro: `npm run knip` reports `apps/web/vercel.ts`
- fixed local source against updated repro: clean
- fixed local source against `rip-technologies` with the `**/vercel.ts` ignore removed: clean

### DIFF
--- a/packages/knip/fixtures/plugins/vercel-workspaces/apps/web/package.json
+++ b/packages/knip/fixtures/plugins/vercel-workspaces/apps/web/package.json
@@ -1,3 +1,6 @@
 {
-  "name": "@plugins/vercel-workspaces-web"
+  "name": "@plugins/vercel-workspaces-web",
+  "devDependencies": {
+    "@vercel/config": "*"
+  }
 }
--- a/packages/knip/fixtures/plugins/vercel-workspaces/apps/web/package.json
+++ b/packages/knip/fixtures/plugins/vercel-workspaces/apps/web/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "@plugins/vercel-workspaces-web"
+}
--- a/packages/knip/fixtures/plugins/vercel-workspaces/apps/web/src/index.ts
+++ b/packages/knip/fixtures/plugins/vercel-workspaces/apps/web/src/index.ts
@@ -1,0 +1,1 @@
+export const value = 1;
--- a/packages/knip/fixtures/plugins/vercel-workspaces/apps/web/vercel.ts
+++ b/packages/knip/fixtures/plugins/vercel-workspaces/apps/web/vercel.ts
@@ -1,0 +1,3 @@
+export const config = {
+  framework: 'nextjs',
+};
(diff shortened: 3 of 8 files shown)


webpro:
Thanks for the PR. How about using `@vercel/config` as an enabler: [https://vercel.com/docs/
project-configuration/vercel-ts#install-@vercel/config](https://vercel.com/docs/project-configuration/vercel-ts#install-@vercel/config). That's basically what makes the programmatic config interesting anyway, right?

The reason I set `rootOnly` is to prevent additional fs access which plugins ideally don't do. So if we can prevent reading all permutations in all workspaces (especially in repos that don't use this plugin), that would be nice. Reading a package name from a set is nearly free (like almost all plugins do).

Btw no need to open an issue if you're planning to open a PR for that issue anyway.

jakeleventhal:
@webpro if you do it based on the package, what about people who dont use the type annotation? or who use .json?

webpro:
The people who use json don't need the plugin anyway. `@vercel/config` exports more than just types.

jakeleventhal:
@webpro done

webpro:
Thanks!

webpro:
:rocket: _This pull request is included in v6.12.1. See [Release 6.12.1](https://github.com/webpro-nl/knip/releases/tag/knip%406.12.1) for release notes._

PR #1763, opened 2026-05-22
Load plugin configs from workspace cwd

jakeleventhal:
## Summary

- Load JavaScript/TypeScript plugin config files with the workspace/package cwd, while restoring the original process cwd afterward.
- Keep the cwd switch scoped to plugin config loading instead of changing the generic loader API/behavior for every `load()` caller.
- Add a GraphQL Codegen regression fixture whose `--config ./src/codegen.ts` file reads package-relative config and only resolves when evaluated from the package cwd.

## Root Cause

`graphql-codegen --config ./src/codegen.ts` is discovered from the package script, so the config should be evaluated the same way that package script runs: from the workspace/package cwd. When Knip loads the config from the repo-root cwd, relative paths inside the config resolve incorrectly and plugin dependencies referenced by the loaded config can be reported as unused.

## Validation

- `node --test packages/knip/test/plugins/graphql-codegen.test.ts`
- `pnpm exec oxfmt --check packages/knip/src/util/loader.ts packages/knip/src/util/plugin.ts packages/knip/test/plugins/graphql-codegen.test.ts packages/knip/fixtures/plugins/graphql-codegen-script-config-cwd-context/**/*.ts packages/knip/fixtures/plugins/graphql-codegen-script-config-cwd-context/**/*.json`
- `pnpm run --dir packages/knip lint`
- `pnpm run --dir packages/knip build`
- No-cache riptech validation with the GraphQL Codegen ignore removed and the original relative dotenv path left in place: `node /Users/jakeleventhal/Documents/Codex/2026-05-22/update-my-pr-based-on-this/knip/packages/knip/src/cli.ts --directory /Users/jakeleventhal/Developer/rip-technologies --no-progress --workspace @artelo/integrations`

### DIFF
--- a/packages/knip/fixtures/plugins/graphql-codegen-script-config-cwd-context/config/codegen-plugins.json
+++ b/packages/knip/fixtures/plugins/graphql-codegen-script-config-cwd-context/config/codegen-plugins.json
@@ -1,0 +1,5 @@
+[
+  "typescript",
+  "typescript-operations",
+  "typescript-graphql-request"
+]
--- a/packages/knip/fixtures/plugins/graphql-codegen-script-config-cwd-context/node_modules/@graphql-codegen/cli/package.json
+++ b/packages/knip/fixtures/plugins/graphql-codegen-script-config-cwd-context/node_modules/@graphql-codegen/cli/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "@graphql-codegen/cli",
+  "bin": {
+    "gql-gen": "index.js",
+    "graphql-codegen": "index.js",
+    "graphql-code-generator": "index.js",
+    "graphql-codegen-esm": "index.js"
+  }
+}
--- a/packages/knip/fixtures/plugins/graphql-codegen-script-config-cwd-context/package.json
+++ b/packages/knip/fixtures/plugins/graphql-codegen-script-config-cwd-context/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "@plugins/graphql-codegen-script-config-cwd-context",
+  "scripts": {
+    "codegen": "graphql-codegen --config ./src/codegen.ts"
+  },
+  "devDependencies": {
+    "@graphql-codegen/cli": "*",
+    "@graphql-codegen/typescript": "*",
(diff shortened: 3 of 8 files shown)


jakeleventhal:
Not sure if you have a better approach here

webpro:
AST resolver is fine (better), but ideally it's either/or (not both `resolveConfig` and `resolveFromAST`).

Btw documentation for users at https://knip.dev/reference/known-issues#exceptions-from-config-files

jakeleventhal:
@webpro 

Thanks. I reworked this to avoid combining `resolveConfig` and `resolveFromAST`.

Looking at the repro again, the issue is not that the config throws and Knip should recover with AST. It’s that Knip evaluates the `--config ./src/codegen.ts` file from the repo-root cwd, while the package script runs from the workspace cwd. Relative paths inside the config then resolve differently.

The updated PR keeps the existing `resolveConfig` path and changes plugin config loading to temporarily use the workspace cwd.

webpro:
While I understand it _should just work_ and _it's on Knip_, yet still: why not follow good practice and use `resolve` or `join` for absolute path? Not sure if we should add `chdir()` for _each_ `load()` invocation.

jakeleventhal:
@webpro check again

webpro:
Thanks! Much better, performance is no longer an argument against this. However, I'm still on the fence with this one, because with this type of change there's basically no going back once merged and it blocks potential future parallel work.

jakeleventhal:
@webpro what's the path forward here?

webpro:
Let's close it and not shoot ourselves in the foot. I'll close it by updating docs soon.

jakeleventhal:
@webpro then how can I solve cases like this: https://github.com/jakeleventhal/knip-1763-repro

webpro:
That link 404s. Probably I would answer:

> follow good practice and use `resolve` or `join` for absolute path

jakeleventhal:
err, sorry can you check again please @webpro

webpro:
:rocket: _This issue has been resolved in v6.16.0. See [Release 6.16.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.16.0) for release notes._

PR #1772, opened 2026-05-28
fix(vite): detect inline module script entry points in index.html

### LINKED ISSUE #1771: 💡 Vite plugin: detect entry points from inline module scripts in index.html
### Suggest an idea for Knip

The Vite plugin's `resolve` function detects entry points from `index.html` by matching `<script type="module" src="...">` tags. However, it does not detect inline module scripts.

**Currently detected:**

```html
<script type="module" src="/src/main.tsx"></script>
```

**Not detected:**

```html
<script type="module">
  import '/src/main.tsx'
</script>
```

The current regex in `packages/knip/src/plugins/vite/helpers.ts` requires both `type="module"` and a `src` attribute:

```ts
const moduleScriptPattern =
  /<script\b(?=[^>]*\btype\s*=\s*["']?module["']?)(?=[^>]*\bsrc\s*=\s*["']?([^"' >]+)["']?)[^>]*>/gi;
```

Inline module scripts are valid HTML and used in some project setups (e.g. conditional loading, pre-import setup, or framework scaffolds that don't use `src`). These entry points are silently missed, which can lead to false positives for unused files/dependencies.

**Proposal:** Extend `getModuleScriptSources` in the Vite plugin to also extract import specifiers from inline `<script type="module">` blocks (static `import` declarations).

Happy to submit a PR if this sounds reasonable.

lucas-spin:
Closes #1771

## Problem

The Vite plugin's `getModuleScriptSources` helper only detects `<script type="module" src="...">` tags — scripts that reference an external file via the `src` attribute. Inline module scripts like:

```html
<script type="module">
  import '/src/main.tsx'
</script>
```

are silently skipped, causing false positive "unused file" reports for projects using this pattern.

## Solution

Replaced the single-pass regex (which required both `type="module"` **and** `src` via positive lookaheads) with a two-pass approach:

1. Match all `<script type="module" ...>...</script>` blocks
2. For each block:
   - If `src` attribute is present → extract path (preserves existing behavior)
   - If no `src` → scan the inline body for import statements and extract file-path specifiers

Only relative/absolute paths (`/`, `./`, `../`) are picked up from inline scripts — bare package specifiers (e.g. `'vue'`) are ignored since they're resolved elsewhere.

## Regex notes

The `importSpecPattern` is intentionally simple and covers:
- Static: `import '/src/main.tsx'`, `import App from './App.vue'`
- Dynamic: `import('/src/lazy.ts')`

This aligns with the patterns already used in `src/compilers/compilers.ts` (`importMatcher`).

## Test

Added `vite5` fixture with an inline `<script type="module">` importing `/src/main.tsx`. The test asserts that `src/main.tsx` is correctly identified as an entry point (not flagged unused) while `src/unused.ts` is still reported.


webpro:
Thanks Lucas!

webpro:
:rocket: _This pull request is included in v6.15.0. See [Release 6.15.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.15.0) for release notes._

PR #1774, opened 2026-06-01
Add default ignore for .context artifacts (for Conductor)

jakeleventhal:
## What changed

- Adds `.context/**` as a built-in default ignore pattern so Conductor AI worktree artifacts are suppressed without user configuration.

- Keeps built-in ignores out of unused ignore-pattern configuration hints and preserves negated user ignore behavior by evaluating default ignores separately.

- Adds a regression fixture/test and documents the default in the configuration reference.
## Validation
- `bun test packages/knip/test/default-ignore-context.test.ts packages/knip/test/ignore-negated.test.ts packages/knip/test/ignore-patterns.test.ts`
- `corepack pnpm exec oxfmt --check packages/knip/src/constants.ts packages/knip/src/ConfigurationChief.ts packages/knip/src/IssueCollector.ts packages/knip/src/graph/build.ts packages/knip/test/default-ignore-context.test.ts packages/docs/src/content/docs/reference/configuration.md`
- `corepack pnpm exec oxlint packages/knip/src/constants.ts packages/knip/src/ConfigurationChief.ts packages/knip/src/IssueCollector.ts packages/knip/src/graph/build.ts packages/knip/test/default-ignore-context.test.ts`

- `corepack pnpm --dir packages/knip build`

- `corepack pnpm --dir packages/knip test`

### DIFF
--- a/packages/docs/src/content/docs/reference/configuration.md
+++ b/packages/docs/src/content/docs/reference/configuration.md
@@ -250,6 +250,9 @@ Avoid `ignore` patterns. There is almost always a better solution:
 - Use [production mode][10].
 - Other `ignore*` options.
 
+Knip ignores `.context/**` by default. This excludes AI worktree artifacts
+created by tools such as Conductor.
+
 **NOTE**: An exception to the rule: to _temporarily_ report only issues in files
 that match the negated `ignore` pattern:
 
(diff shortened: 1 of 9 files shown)


webpro:
Why not add `.context` to `.gitignore` (or `.git/info/exclude`)?

jakeleventhal:
@webpro turns out the problem actually is that `.context` is not specified in gitignore, rather from `.git/info/exclude` (I don't recall ever setting this, maybe conductor did).

I think the correct approach is that knip should respect `.git/info/exclude` perhaps?

webpro:
It should do that already:

https://github.com/webpro-nl/knip/blob/33a26fb0ad4d1fcd0b34575d8d55570004818fec/packages/knip/src/util/glob-core.ts#L152-L156

PR #1787, opened 2026-06-09
fix(capacitor): detect iOS platform with Swift Package Manager

### LINKED ISSUE #1785: Capacitor plugin: support Swift Package Manager iOS projects (false-positive unused @capacitor/ios)
The built-in Capacitor plugin in knip incorrectly reports @capacitor/ios as an unused dependency when a Capacitor project's iOS target is configured with Swift Package Manager (SPM) instead of CocoaPods.

Starting with Capacitor 6, iOS projects can use SPM as an alternative to CocoaPods. In that setup, the iOS project typically contains a Swift package manifest (for example, ios/App/CapApp-SPM/Package.swift) and does not include an ios/App/Podfile.

The issue appears to be that knip's Capacitor plugin only checks for the presence of a CocoaPods Podfile when determining whether the iOS platform is installed. As a result, SPM-based projects are treated as if the iOS platform is missing, causing @capacitor/ios to be flagged as unused even though it is required and actively used to build the iOS application.

Happy to put up a PR if it would be welcome. Thanks for making such a useful piece of software!

jthrilly:
Fixes #1785. 

Capacitor 6+ supports Swift Package Manager (SPM) for the iOS platform, which generates `ios/App/CapApp-SPM/Package.swift` instead of a CocoaPods `Podfile`. The plugin only checked for the `Podfile`, so SPM-based projects had `@capacitor/ios` falsely reported as an unused dependency. 

Note: I authored this with Claude, but manually verified and reviewed. If you'd like a different implementation, please let me know and I'd be happy to take another look.

### DIFF
--- a/packages/knip/fixtures/plugins/capacitor-spm/capacitor.config.ts
+++ b/packages/knip/fixtures/plugins/capacitor-spm/capacitor.config.ts
@@ -1,0 +1,9 @@
+import type { CapacitorConfig } from '@capacitor/cli';
+
+const config: CapacitorConfig = {
+  appId: 'com.company.name',
+  appName: 'name',
+  plugins: {},
+};
+
+export default config;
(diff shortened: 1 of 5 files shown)


webpro:
Thanks Joshua!

webpro:
:rocket: _This pull request is included in v6.17.0. See [Release 6.17.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.17.0) for release notes._

PR #1791, opened 2026-06-11
Don't disable configuration hints in workspace-scoped runs

### LINKED ISSUE #1790: 🐛 --workspace disables configuration hints and makes --treat-config-hints-as-errors a silent no-op
### Prerequisites

### Reproduction url

https://github.com/WooWan/knip-workspace-hint-reproduction

### Reproduction access

### Description of the issue

When `--workspace` is set, configuration hints are unconditionally disabled:

```ts
// packages/knip/src/util/create-options.ts
isDisableConfigHints: args['no-config-hints'] || isProduction || Boolean(workspace),
```

As a result, `--treat-config-hints-as-errors` silently does nothing in workspace-scoped runs
(`cli.ts` guards the exit code on `!isDisableConfigHints`, and reporters/`IssueCollector`
skip hint collection entirely).

## Reproduction

```sh
# services/knip-reproduction has a stale ignoreDependencies(react)
knip --treat-config-hints-as-errors                          # reports hint, exits 1 ✅
knip --workspace services/knip-reproduction --treat-config-hints-as-errors # no hint, exits 0 ❌
```

## Why this matters

In large monorepos, CI typically runs Knip per changed workspace. In that setup, stale
`ignoreDependencies` / `ignore` / `ignoreFiles` entries can never fail CI — even when they
belong to the selected workspace. This is surprising because `--workspace` is a scoping
option, while config hints are a hygiene signal.

## Suggested fix
Would it make sense to keep configuration hints enabled for --workspace runs, but filter
them to the selected workspace?

WooWan:
- fixes: #1790 

`--workspace` unconditionally disables configuration hints:

```ts
// src/util/create-options.ts
isDisableConfigHints: args['no-config-hints'] || isProduction || Boolean(workspace),
```

So `--treat-config-hints-as-errors` is a silent no-op in workspace-scoped runs — in monorepo CI that lints only changed workspaces, stale `ignoreDependencies` entries can never fail CI.

This PR filters hints to the selected workspaces instead of disabling them, mirroring how issues are filtered. Hints without a `workspaceName` are treated as root workspace hints.

`--no-config-hints` and production mode are unchanged. Tests added in `test/workspaces/config-hints.test.ts`.

### DIFF
--- a/packages/knip/fixtures/workspaces/config-hints/knip.json
+++ b/packages/knip/fixtures/workspaces/config-hints/knip.json
@@ -1,0 +1,11 @@
+{
+  "ignoreDependencies": ["unused-ignore-root"],
+  "workspaces": {
+    "packages/a": {
+      "ignoreDependencies": ["unused-ignore-a"]
+    },
+    "packages/b": {
+      "ignoreDependencies": ["unused-ignore-b"]
+    }
+  }
+}
--- a/packages/knip/fixtures/workspaces/config-hints/knip.json
+++ b/packages/knip/fixtures/workspaces/config-hints/knip.json
@@ -1,5 +1,5 @@
 {
-  "ignoreDependencies": ["unused-ignore-root"],
+  "ignoreDependencies": ["unused-ignore-root", "used-by-a"],
   "workspaces": {
     "packages/a": {
       "ignoreDependencies": ["unused-ignore-a"]
--- a/packages/knip/fixtures/workspaces/config-hints/package.json
+++ b/packages/knip/fixtures/workspaces/config-hints/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "@fixtures/workspaces-config-hints",
+  "version": "*",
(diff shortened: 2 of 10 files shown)


webpro:
Thanks for the PR. I deliberately gated this, because it might lead to confusion as ancestor + dependent workspaces of the selected ones are also included.

Made an adjustment to your branch to resolve conflict and `--workspace .` still returns no hints (guarding against false positives, unwanted results in CI). WDYT?

changwan-wu:
Thanks for the adjustment.

Since the common CI use case in monorepos is checking individual package/service workspaces, excluding root/top-level hints from workspace-scoped runs makes sense to me!

webpro:
Thanks for the feedback, glad you agree. Let's go 🚀

webpro:
:rocket: _This pull request is included in v6.17.2. See [Release 6.17.2](https://github.com/webpro-nl/knip/releases/tag/knip%406.17.2) for release notes._

PR #1805, opened 2026-06-19
fix: mark npx binary as optional to suppress false unlisted-binary reports

axelray-dev:
## Summary

When a script uses `npx some-tool`, knip reports `some-tool` as "Unlisted binary" even though it's being run on-demand via npx and shouldn't be expected in `package.json`.

The `bunx` resolver already marks binaries as optional (`toBinary(specifier, { optional: true })`), but the `npx` resolver did not.

## Fix

Add `{ optional: true }` to the `toBinary` call in the npx resolver when the specifier is the npx target (no `--package` flag). When `--package` is used, the positional arg is a command binary and remains non-optional.

This matches the existing `bunx` resolver behavior.

## Test plan

- Updated test expectations in `test/util/get-inputs-from-scripts.test.ts` to reflect that npx-resolved binaries are now optional
- All 30 tests in `get-inputs-from-scripts.test.ts` pass
- Cases with `--package` flag correctly leave the command binary non-optional

### DIFF
--- a/packages/knip/src/binaries/resolvers/npx.ts
+++ b/packages/knip/src/binaries/resolvers/npx.ts
@@ -20,7 +20,7 @@ export const resolve: BinaryResolver = (_binary, args, options) => {
   const restArgs = argsFrom(args, packageSpecifier);
 
   const isBinary = specifier && !packageSpecifier.includes('@') && !isInternal(specifier);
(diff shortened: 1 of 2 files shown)


webpro:
Thanks for the PR! Since npx has `-y` and `-n` flag and the fix is slightly more involved so I just went ahead with it in main, hope you don't mind.

Issue #1804, opened 2026-06-18
🐛 Unused devDependency while used for extending in the tsconfig


Netail:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/Netail/repro-knip/tree/main/repros/unused-devdependency-tsconfig-extends

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

devDependency marked as unused while used for extending in the tsconfig

webpro:
Looks like there's a typo: `"devDe"`

Netail:
> Looks like there's a typo: `"devDe"`

Oops, that's a typo in the repro. But did somehow experience the issue in our monorepo. Lemme get back to this ticket

Netail:
Oh it's because we make use of node's type-stripping, which doesn't require typescript to be installed. But that doesn't make use of the tsconfig anyway apparently

Netail:
However, we do use the tsconfig for some set up in the jest config. So the tsconfig file is technically used @webpro

PR #1806, opened 2026-06-19
fix: enable TypeScript plugin when a tsconfig.json is present (resolv…

ixexel661:
Summary
When a workspace's tsconfig.json extends a package by subpath (e.g. "extends": "@repo/typescript-config/base.json"), knip reports that package as an "Unused devDependency", even though the tsconfig genuinely uses it.

The TypeScript plugin is what credits the extended package as a used dependency (via resolveConfig), but it only runs when typescript (or @typescript/native-preview) is a declared dependency. If a workspace has a tsconfig.json but no direct typescript dependency, and no ancestor provides one, the plugin stays off, the extends chain is never parsed for dependencies, and the config package gets flagged. Note that knip's core already resolves the extends chain (via get-tsconfig) to merge compilerOptions, so it reads the package's file without ever crediting it.

Fix
Enable the TypeScript plugin when a tsconfig.json is present, not only when typescript is a dependency:

const isEnabled: IsPluginEnabled = ({ cwd, dependencies }) =>
  hasDependency(dependencies, enablers) || isFile(cwd, 'tsconfig.json');

This follows the existing pnpm and yarn plugins, which enable on lockfile presence (isFile(cwd, 'yarn.lock')).

Test plan
Added fixture fixtures/plugins/typescript-extends-package (a consumer that extends a workspace config package by subpath, with no typescript dependency) and test/plugins/typescript-extends-package.test.ts. It fails before the fix and passes after.
Updated the moonrepo and remix plugin test expectations. With the plugin now enabled, knip correctly surfaces an extended-but-undeclared @tsconfig/node20 (unlisted) and a compilerOptions.types entry vitest/globals (unresolved).
Full smoke suite shows no new regressions.

### DIFF
--- a/packages/knip/fixtures/plugins/typescript-extends-package/package.json
+++ b/packages/knip/fixtures/plugins/typescript-extends-package/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "@plugins/typescript-extends-package",
+  "private": true,
+  "workspaces": [
+    "packages/*"
+  ]
+}
--- a/packages/knip/fixtures/plugins/typescript-extends-package/packages/app/node_modules/@plugins/typescript-extends-package__config/base.json
+++ b/packages/knip/fixtures/plugins/typescript-extends-package/packages/app/node_modules/@plugins/typescript-extends-package__config/base.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "compilerOptions": {
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "strict": true
+  }
+}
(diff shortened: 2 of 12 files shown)


webpro:
Thanks for the PR. Plugins are enabled by having a dependency in the workspace or ancestor workspace so I don't think this changeset is necessary. Also, the #1804 repro was invalid to begin with.

webpro:
Out of curiosity still, what is a real-world scenario where `typescript` should **not** be a **listed** (dev) dependency but `tsconfig.json` exists that should be parsed by Knip?

Netail:
> Out of curiosity still, what is a real-world scenario where `typescript` should **not** be a **listed** (dev) dependency but `tsconfig.json` exists that should be parsed by Knip?

We import the tsconfig into our jest config due to a plugin requiring the tsconfig for some module mapping

PR #1812, opened 2026-06-24
Add circular dependencies as opt-in `cycles` issue type (resolve #1734)

### LINKED ISSUE #1734: 💡 Tracking circular dependencies
### Suggest an idea for Knip

👋 __webpro__ and folks!

At the moment, knip docs mention circular dependency tracking as something that's out of scope of knip.

https://knip.dev/reference/related-tooling#circular-dependencies

https://github.com/webpro-nl/knip/blob/b255195be57173284dd313da599e7e86718c9cdb/packages/docs/src/content/docs/reference/related-tooling.md?plain=1#L28-L32

I wonder if that's a strategic decision or a feature that can be eventually added at some point.

I use knip in a couple of projects and it's a really useful source of feedback. Since exports are already tracked, knip must have already built an import graph that could be used for searching cycles in it. The existing config system is a perfect fit for exceptions, if necessary.

I am currently using `dpdm` to fill in the gap and it works, but feels to occupy in the same linting niche as knip. Unfortunately, `dpdm` cannot be configured via `dpdm.config.ts` (https://github.com/acrazing/dpdm/issues/65) and there is no way to define a list of known exceptions.

WDYT of including circular dependency scanning into knip?

webpro:
_tl/dr;_ Try with `knip --cycles --reporter cycles`:

```
$ knip --cycles --reporter cycles

Circular dependencies (1)

src/i18n/index.ts:1:28
└── import ./middleware → src/i18n/middleware.ts:3:15
    └── re-export ../core/i18n/handler → src/core/i18n/handler.ts:5:10
        └── import ../../i18n → src/i18n/index.ts:1:28 ↩
```

Detect circular dependencies across the whole module graph, reported as the opt-in `cycles` issue type. Enable with `--cycles`, `--include cycles` or `{"include": ["cycles"]}`; defaults to the `warn` rule. Set `rules.cycles` to `error` to fail on them.

Detection is a single-pass back-edge DFS (`findAllCycles`) over runtime (non-type-only) edges, emitting one cycle per back-edge. Reuses the issue-type machinery: filters, rules, every reporter, counters, exit code.

In the default report each cycle is one row: the route from the anchor file, paths relative to the root. The dedicated `cycles` reporter prints a verbose multi-line tree per cycle.

Updated docs at https://circular-dependencies.knip.pages.dev/features/reporters#cycles


kachkaev:
👋 @webpro, great stuff! Excited to try this new feature shortly!

Just to make sure I've understood it:  would `{"include": ["cycles"]}` check _only_ circular deps, or everything by default plus the circular deps? I would like to keep all my knip checks in a single run, but `--include` and `--reporter cycles` suggest that we'll need two of them? If it's the latter, what's the reason for checking circurlar deps on their own?

webpro:
- `--include cycles` → include the `cycles` issue type
- `--cycles` → shorthand to include only `cycle` issue type (like `--files`)
- `--reporter cycles` → print only cycles, tree-style with full clickable paths + position

Zamiell:
Interesting feature! Historically, the ecosystem has used the [`import-x/no-cycle`](https://github.com/un-ts/eslint-plugin-import-x/blob/master/docs/rules/no-cycle.md) ESLint rule to detect this. Can you elaborate a bit on what the motivation was to have it in Knip?

Also, I think it would be nice to [add to the docs](https://circular-dependencies.knip.pages.dev/features/reporters#cycles) a discussion of the `no-cycle` rule. Specifically: in what situations would you want to use the ESLint rule over Knip and vice versa?

lishaduck:
Knip has already created a module-graph as part of, well, existing. eslint-plugin-import is slow, incompatible with ESLint's caching (though so is TSESLint), is heavy, and generally doesn't support as much of of the ecosystem (e.g., it's not going to include MDX in the module graph). (That's not meant as a criticism of eslint-plugin-import(-x), it's fundamentally a limitation of how ESLint processes files)

webpro:
Next to what Eli lays out, I can add that it was fun to build and it was already [one of the existing operations](https://github.com/webpro-nl/knip/blob/23b4a744330d0dddd2b1517f2f2d72c9d373a0b9/packages/knip/src/graph-explorer/explorer.ts#L37-L36) on the module graph (i.e. before this PR, available in the [Editor Extension](https://knip.dev/reference/integrations#circular-dependencies)). The heavy lifting is done building the module graph, such ops are then relatively cheap. By design such graph-based ops spanning multiple files fit Knip better than any file-based linter.

Integrated into Knip also means affordances like support for monorepos/workspaces (including workspace filtering), non-standard file support, etc. etc. as `cycles` would be just another issue type.

Btw we can add the ESLint rule to https://knip.dev/reference/related-tooling#circular-dependencies when we're at it.

kachkaev:
> `--reporter cycles` → print only cycles, tree-style with full clickable paths + position

Is there a way to have this reporter while running other checks too? I am trying to have only one run of knip in CI which does what it used to _plus_ checks for cycles. I'm a bit confused by what changes to make in `knip.config.ts`. It also seems like `includes: ["cycles"]` removes the default checks - not sure have to keep the defaults.

webpro:
> In the default report each cycle is one row: the route from the anchor file, paths relative to the root. The dedicated `cycles` reporter prints a verbose multi-line tree per cycle.

I think you can use multiple reporters if you want. Try it out.

webpro:

PR #1825, opened 2026-06-27
feat: add support for Lunaria

### LINKED ISSUE #1822: 💡 [plugin]: add support for Lunaria
### Suggest an idea for Knip

I really love Knip so far. The philosophy to have little config and automatically detect everything is implemented really smooth.

## Idea of this request

When I tried it in some of my Starlight repositories, I noticed that one plugin that would be very useful is [**Lunaria**](https://lunaria.dev/). Although the project is officially in beta, it is already used in many places, like [Starlight](https://starlight.astro.build/) and [Astro Docs](https://docs.astro.build/en/getting-started/) itself, and also other repos in the ecosystem, like [npmx](https://npmx.dev/).

## Implementation details

I think a plugin for Lunaria should be quite simple: The default config file is [`lunaria.config.json`](https://lunaria.dev/reference/configuration/), and with [`resolveConfig`](https://knip.dev/writing-a-plugin#6-resolveconfig) (I _think_) the following JSON fields should be added with [`toProductionEntry`](https://knip.dev/writing-a-plugin/inputs#toproductionentry) because they can reference relative files:

- [`files[].location`](https://lunaria.dev/reference/configuration/#files-required)
- [`renderer`](https://lunaria.dev/reference/configuration/#renderer) (optional field)
- [`dashboard.customCss`](https://lunaria.dev/reference/configuration/#dashboardcustomcss) (optional field)
- [`dashboard.favicon.inline`](https://lunaria.dev/reference/configuration/#dashboardfavicon) (optional field)

I'm a little unsure about these fields, as these are files that should be ignored, but I think if we just define the above references as `entry`s, we do not need to set them explicitly in `ignore`:

- [`outDir`](https://lunaria.dev/reference/configuration/#outdir) (optional field)
- [`cloneDir`](https://lunaria.dev/reference/configuration/#clonedir) (optional field)
- [`dashboard.basesToHide`](https://lunaria.dev/reference/configuration/#dashboardbasestohide) (optional field)

Since I am just discovering Knip and have no prior experience with this type of tool, I wanted to create this issue first to clarify if my understanding is correct. Please correct me if any parts of this plan for implementation could be done better with the _Knip_ way.

- [x] I am open to creating a PR that adds support for Lunaria.

trueberryless:
#### Description

This PR adds support for `@lunariajs/core` and `@lunariajs/starlight`.

- Closes #1822


packages/knip/src/plugins/lunaria/index.ts
```diff
@@ -0,0 +1,66 @@
...
+const title = 'Lunaria';
+
+const enablers = [/^@lunariajs\//];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) =>
+  hasDependency(dependencies, enablers);
+
+const config: string[] = [
+  'lunaria.config.json',
+  'astro.config.{js,cjs,mjs,ts,mts}',
```
webpro: I'm leaning towards grabbing the config path from the Astro plugin, because `resolveConfig` and `resolveFromAST` are executed on each found config file (it doesn't hurt but we don't really want to have `astro.config.ts` go through Lunaria's `resolveConfig` here). What do you think about moving the logic around `toConfig` to the Astro plugin?

packages/knip/src/plugins/lunaria/index.ts
```diff
@@ -0,0 +1,66 @@
...
+) => {
+  if (!localConfig) return [];
+  const files = localConfig.files?.map(f => f.location) ?? [];
+  const renderer = localConfig.renderer ? [localConfig.renderer] : [];
+  const customCss = localConfig.dashboard?.customCss ?? [];
+  const favicon = localConfig.dashboard?.favicon?.inline
+    ? [localConfig.dashboard.favicon.inline]
+    : [];
+
+  const baseDir = dirname(options.configFilePath);
```
webpro: Nitpick: there is `configFileDir`

packages/knip/src/plugins/lunaria/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    toProductionEntry(isAbsolute(id) ? id : join(baseDir, id))
+  );
+};
+
+const resolveFromAST: ResolveFromAST = (program: Program) => {
+  const configPaths = collectPropertyValues(program, 'configPath');
+  return [...configPaths].map(id => toConfig('lunaria', id));
+};
+
+const args = {
```
webpro: We could import `Args` for easier typing like so:

```suggestion
const args: Args = {
```

packages/knip/src/plugins/lunaria/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  localConfig,
+  options
+) => {
+  if (!localConfig) return [];
+  const files = localConfig.files?.map(f => f.location) ?? [];
+  const renderer = localConfig.renderer ? [localConfig.renderer] : [];
+  const customCss = localConfig.dashboard?.customCss ?? [];
+  const favicon = localConfig.dashboard?.favicon?.inline
+    ? [localConfig.dashboard.favicon.inline]
+    : [];
```
webpro, line 33: In most projects, the `.svg` and `.css` files won't be added to the module graph by Knip. This is fine, as those files aren't included/reported as being unused in the first place. Only if the user has Tailwind or custom .css compiler configured then this should be picked up (https://feat-bulk-suppressions.knip.pages.dev/features/compilers). Same for the `.svg` file: only if a user has `.svg` compiler configured (pretty rare), will this input be useful.

packages/knip/src/plugins/lunaria/index.ts
trueberryless: Oh, that is smart. I didn't consider this possibility. 🤯 

Will do it soon 👍

packages/knip/src/plugins/lunaria/index.ts
trueberryless: The reason I included those files is that I tested Knip on the `withastro/starlight` repo and e.g. the Lunaria CSS files were marked as unused:

image

I'm not totally sure why CSS files get reported in this case, because [`withastro/starlight` docs](https://github.com/withastro/starlight/blob/main/docs/package.json) does not use Tailwind and AFAIK we also don't have a CSS compiler 🤔 
But you are probably right that this shouldn't be Lunaria's concern in either case, so I'll remove the `favicon` and `customCSS` entries 👍

packages/knip/src/plugins/lunaria/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  ResolveFromAST,
+} from '../../types/config.ts';
+import type { Args } from '../../types/args.ts';
+import { toProductionEntry } from '../../util/input.ts';
+import { hasDependency } from '../../util/plugin.ts';
+import type { LunariaConfig } from './types.ts';
+
+const title = 'Lunaria';
+
+const enablers = [/^@lunariajs\//];
```
webpro: Could we make this more explicit? Unless there are still more packages we should include, this is faster/clearer:

```suggestion
const enablers = ["@lunariajs/core", "@lunariajs/starlight"];
```

packages/knip/test/plugins/lunaria.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+test('Find dependencies with the lunaria core scenario', async () => {
+  const cwd = resolve('fixtures/plugins/lunaria/core');
+  const options = await createOptions({ cwd });
+  const { counters } = await main(options);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    binaries: 1,
+    dependencies: 1,
```
webpro: FYI, the reason for  1 bin + 1 dep here is that the dependency name does not match the binary name and there is no `node_modules` with a `package.json` that provides the mapping. Not blocking.

webpro (commented):
Thanks for the PR! LGTM, just a nit left.

packages/knip/src/plugins/lunaria/index.ts
trueberryless: Thanks for the suggestion! I wanted to update it like that at one point, but forgot to do it then. Sorry!
Much clearer 🥳

trueberryless:
Thanks for the reviews!

I tested out the preview package and can confirm that the `renderer.ts` and references from within the renderer file are resolved correctly and do not appear anymore in the report:

image

image

Regarding the still reported `lunaria/styles.css` file. Do you have any recommendations for how to handle it? I found that another project in the `withastro/starlight` monorepo uses tailwind and lightningcss, but I wouldn't expect that projects interfere with each other like that 🤔

packages/knip/test/plugins/lunaria.test.ts
trueberryless: How would a `package.json` in the `node_modules` folder need to look like to provide the mapping? I have to admit that I am a little unfamiliar regarding this 😅

webpro:
Thanks for the extra test drives, way to go. Please run knip on itself and find that some types can be un-exported (blocks CI currently).

We should keep `customCss` inputs. Sorry if I wasn't clear, didn't mean to say they must be deleted, was just additional info.

packages/knip/test/plugins/lunaria.test.ts
webpro: A minimal `node_modules/@lunariajs/core/package.json`:

```json
{
  "name": "@lunariajs/core",
  "bin": {
    "lunaria": "./index.js"
  }
}
```

I think `index.js` does not even need to exist btw.

trueberryless: Thanks, this is much cleaner 🎉

trueberryless:
Alright, thanks for taking the time to explain all my question. The current implementation looks much better than my initial draft. I learned a lot about Knip along the way. Was a pleasure 🫶

webpro:
Likewise! I am grateful to receive a _human_ PR and go through this together. Will release shortly.

webpro:
:rocket: _This pull request is included in v6.22.0. See [Release 6.22.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.22.0) for release notes._

PR #1829, opened 2026-06-27
feat: add support for @astrojs/markdoc

trueberryless:
#### Description

This PR creates a new Knip plugin for the [`@astrojs/markdoc` package](https://npmx.dev/package/@astrojs/markdoc/v/2.0.1).
Astro's Markdoc integration offers a [configuration file](https://docs.astro.build/en/guides/integrations-guide/markdoc/#render-components), which references Astro components via the `render` property (in [`nodes`](https://docs.astro.build/en/guides/integrations-guide/markdoc/#custom-markdoc-nodes--elements) and [`tags`](https://docs.astro.build/en/guides/integrations-guide/markdoc/#use-astro-components-as-markdoc-tags)). The plugin resolves all those paths, so the Knip analysis graph can be constructed.

I decided to add the `astro` dep in the fixture as it is a peerDep from `@astrojs/markdoc`.
I added the config file as a default production entry, because of this tip:

image

However, I remember reading somewhere that config files should never be entries (not sure if I am misremembering), so I wanted to mention it here to hopefully make reviewing the PR easier.

---

This PR should also enable auto-detection for users of [`@astrojs/starlight-markdoc`](https://npmx.dev/package/@astrojs/starlight-markdoc), as that package has `@astrojs/markdoc` as a peerDep, so they need to install it either way 🙌


packages/knip/src/plugins/astro-markdoc/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const title = 'Astro Markdoc';
+
+const enablers = ['@astrojs/markdoc'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const config: string[] = ['markdoc.config.{js,ts,mjs,mts}'];
+
+const production: string[] = ['markdoc.config.{js,ts,mjs,mts}'];
```
webpro: `config` entries are auto-added as regular entries, so we can remove `production` here (btw should've been `entry` otherwise as config files are not production code).

packages/knip/src/plugins/astro-markdoc/index.ts
```diff
@@ -0,0 +1,55 @@
...
+const resolveConfig: ResolveConfig<MarkdocConfig> = async localConfig => {
+  if (!localConfig) return [];
+
+  const entries = [...Object.values(localConfig.nodes ?? {}), ...Object.values(localConfig.tags ?? {})];
+
+  const specifiers = entries.flatMap(item =>
+    item && typeof item === 'object' && 'render' in item ? extractSpecifiers(item.render as MarkdocRenderSpecifier) : []
+  );
+
+  return specifiers.map(id => toDeferResolveEntry(id));
```
webpro: Is `toDeferResolveEntry` needed here? If `toEntry` or `toProductionEntry` does the job then that'd be more efficient. Defer resolution if the specifier might not include the file extension or if it's an external package that needs to be found somewhere in `node_modules`.

webpro (commented):
Thanks for another plugin! 🥳 Just a few nits left.

packages/knip/src/plugins/astro-markdoc/index.ts
trueberryless: Thanks for the info, I didn't know 👍

packages/knip/src/plugins/astro-markdoc/index.ts
trueberryless: Good point. For the Astro Markdoc integration, the `.astro` extension must always be specified AFAIK, so `toProductionEntry` should work more efficiently here 👍

trueberryless:
I fixed our suggestions and also added a second parameter to the `component` method (alto mocked), because `@astrojs/markdoc` supports [named exports](https://docs.astro.build/en/guides/integrations-guide/markdoc/#use-astro-components-from-npm-packages-and-typescript-files).

webpro:
Thanks! 🚀

webpro:
:rocket: _This pull request is included in v6.23.0. See [Release 6.23.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.23.0) for release notes._

PR #1838, opened 2026-06-28
fix: remove format_on_save: true settings from Zed settings to respect user settings

trueberryless:
#### Description

Thanks to @mootari for mentioning on Discord that the `format_on_save` option should ideally not be overridden in our projects Zed settings, so users can experience it in the way the want.

Since we'll probably add pre-commit hooks in #1837 anyway, it doesn't really matter if the user always formats when saving the files 👍

### DIFF
--- a/.zed/settings.json
+++ b/.zed/settings.json
@@ -22,7 +22,6 @@
   },
   "languages": {
     "CSS": {
-      "format_on_save": "on",
       "prettier": {
         "allowed": false
       },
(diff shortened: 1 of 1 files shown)


webpro:
:rocket: _This pull request is included in v6.24.0. See [Release 6.24.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.24.0) for release notes._

PR #1841, opened 2026-06-29
feat: backfill `CHANGELOG.md`

trueberryless:
#### Description

This PR generates a `CHANGELOG.md` file from all past knip releases. I used a script for that, which I also committed here so the Changelog doesn't have to be reviewed. Please delete the script before merging.

Given that I would really encourage to continue automatically maintaining the `CHANGELOG.md` file, the question now arises for me which automated release tool fits best. I myself am unfamiliar with `release-it`, but since you maintain it, you can probably evaluate best if this tool is able to do that. Otherwise (also aligning with my feedback regarding the bus factor), I would suggest switching to something like [Changesets](https://changesets.dev/), which is suited best for monorepos (but you probably need to get used to writing changeset files). Or something like [Release Please](https://github.com/googleapis/release-please) would also work, but here we would need to enforce conventional commits. To make the decision easier, I wrote [this little comparison guide](https://deploy-preview-2126--changesets.netlify.app/guide/technical-decisions#what-distinguishes-this-from-other-versioning-tools) (not merged yet, so I link to the deployment preview).

#### Todo
- [ ] Delete backfill script

### DIFF
--- a/changelog-backfill.js
+++ b/changelog-backfill.js
@@ -1,0 +1,108 @@
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const REPO_OWNER = 'webpro-nl';
+const REPO_NAME = 'knip';
+const PACKAGE_NAME = '@knip/mcp';
+
+const TOKEN = process.env.GH_TOKEN || process.env.GITHUB_TOKEN;
+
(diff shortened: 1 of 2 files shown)


webpro:
I don't feel comfortable merging this, as there is not yet a mechanism in place to keep it up to date.

None of the release tools seem to replace what the [release-it script](https://github.com/webpro-nl/knip/blob/main/scripts/release.sh) does. A replacement should have benefits, not replace it for the sake of it.

trueberryless:
Fair enough! 

I'll let you think about how we can improve the bus factor and maybe release workflows in that sense as well.

PR #1847, opened 2026-07-01
feat: add support for `packageExtensions` in yarn and pnpm

### LINKED ISSUE #1779: 💡 Feature request: support Yarn's `packageExtensions`
### Suggest an idea for Knip

Very much an edge case, but it would be neat if knip could support Yarn's [`packageExtensions`](https://yarnpkg.com/configuration/yarnrc#packageExtensions).

For example, with the following in `.yarnrc.yml`:
```yaml
packageExtensions:
  # Cf. https://github.com/react-csv/react-csv/pull/387
  react-csv@*:
    peerDependencies:
      prop-types: "*"
```

then knip should recognize `prop-types` as a peerDependency of react-csv and not report it as an unused dependency.

trueberryless:
#### Description

This PR adds support for the `packageExtensions` setting for yarn and pnpm (only `peerDependencies`). While at it, I also refactored some of the existing code (extracting types, adding `enablers` to pnpm, stuff like that) and also consolidate the tests and fixtures a little bit.

- Closes #1779


packages/knip/test/plugins/pnpm.test.ts
```diff
@@ -1,15 +1,32 @@
...
+});
+
+test('Find dependencies with the pnpm plugin in pnpm-workspace.yaml', async () => {
+  const cwd = resolve('fixtures/plugins/pnpm-workspace');
   const options = await createOptions({ cwd });
-  const { counters } = await main(options);
+  const { issues, counters } = await main(options);
+
+  assert(Object.keys(issues.unlisted || {}).length === 0);
+  assert(Object.keys(issues.dependencies || {}).length === 0);
```
webpro: These assertions are redundant, e.g. `counters.unlisted` (`baseCounters` holds `0` for each) below is the exact same thing. Same in other test file.

packages/knip/src/plugins/pnpm/index.ts
```diff
@@ -1,22 +1,48 @@
...
+const enablers =
+  'This plugin is enabled when a `pnpm-lock.yaml` or `pnpm-workspace.yaml` file is found in the root directory, or when `pnpm@` is specified in the `packageManager` field of `package.json`.';
+
 const isEnabled: IsPluginEnabled = async ({ cwd, manifest }) =>
   manifest.packageManager?.startsWith('pnpm@') || isFile(cwd, 'pnpm-lock.yaml') || isFile(cwd, 'pnpm-workspace.yaml');
 
 const isRootOnly = true;
 
-const config: string[] = ['.pnpmfile.cjs'];
+const config: string[] = ['.pnpmfile.cjs', 'package.json', 'pnpm-workspace.yaml'];
```
webpro: The `resolveConfig` function is invoked with each `config` item.

Looks like think `.pnpmfile.cjs` should be an `entry` item. We had this wrong before this PR, but as there was no `resolveConfig` it behaved the same since `config` is always auto-added as `entry`.

packages/knip/src/plugins/pnpm/types.ts
```diff
@@ -0,0 +1,16 @@
+export type PnpmConfig = {
+  pnpm?: {
```
webpro: I see why we are doing this, but any chance we could extend and use [PackageJson](https://github.com/webpro-nl/knip/blob/cf997b2408cc0814c2d310d1e8c8680340153fa1/packages/knip/src/types/package-json.ts#L34)?

webpro (commented):
Thanks for the PR! Much appreciated. A few comments I have.

packages/knip/src/plugins/pnpm/types.ts
trueberryless: I refactored them in [`269d602` (this PR)](https://github.com/webpro-nl/knip/pull/1847/changes/269d6028b52cb75767a907c7731985a1cd4e79e6), but I am not totally sure if you meant it that way. Let me know what you think 👍

trueberryless:
The change of moving the `.pnpmfile.cjs` from `config` to `entry` (see [linked suggestion](https://github.com/webpro-nl/knip/pull/1847#discussion_r3510548893)) resulted in the test reporting `files: 1`. I adapted it in [`bdaebcd` (this PR)](https://github.com/webpro-nl/knip/pull/1847/commits/bdaebcdab567113d4141261f1d10597edc968d63), but tbh I am not that sure if this is correct. Shouldn't the file be included in Knip's graph if it's in `entry`? I think I am missing something - maybe a missing mapping again? 🤔

webpro:
Sorry forgot about that. If `resolveConfig` is implemented, we should also add basically `entry.map(toEntry)` to the returned inputs. The idea is that plugins would otherwise not override the defaults but only _add_ which would be "globby overhead" (and in some edge cases could even lead to incorrect results).

trueberryless:
> Sorry forgot about that. If `resolveConfig` is implemented, we should also add basically `entry.map(toEntry)` to the returned inputs. The idea is that plugins would otherwise not override the defaults but only _add_ which would be "globby overhead" (and in some edge cases could even lead to incorrect results).

Thank you very much. That makes a lot of sense!

webpro:
Thank you! Nice work.

webpro:

PR #1852, opened 2026-07-02
fix: derive package name from package.json under Yarn `nodeLinker: pnpm` store

### LINKED ISSUE #1851: Package names mis-derived as `.store` under Yarn `nodeLinker: pnpm`
### Prerequisites

- Latest version of Knip: reproduced on `6.24.0`.
- The mis-derivation is in `getPackageNameFromFilePath` (`packages/knip/src/util/modules.ts`), so it is deterministic and reproducible from the code path alone; steps below reproduce it in a real project.

### Description of the issue

Under **Yarn Berry with `nodeLinker: pnpm`**, Knip reports bogus unlisted dependencies literally named `.store`.

#### Root cause

`getPackageNameFromFilePath` derives the package name by taking the **last** `node_modules/` path segment:

```js
const lastPackageNameMatch = /(?<=node_modules\/)(@[^/]+\/[^/]+|[^/]+)/g;
```

That works for the pnpm CLI (which nests `node_modules/<pkg>/`), but Yarn's pnpm linker uses a different on-disk layout. Instead of `node_modules/<pkg>`, packages live in a content-addressed store:

```
node_modules/.store/<flattened-locator>/package/…
```

There is **no** nested `node_modules/<pkg>` segment, so the last segment after `node_modules/` is `.store`, and every package resolved through a store realpath is attributed to a phantom dependency named `.store`. In practice this typically surfaces via plugins that resolve config files through realpaths (e.g. Storybook/Babel addons resolved via `dirname(require.resolve('<addon>/package.json'))`), producing `.store` in the "Unlisted dependencies" report.

#### Why the folder name can't just be parsed

The store folder is `<flatIdent>-<protocol>-<reference>-<hash>`, where scoped names are flattened (`@scope/name` → `@scope-name`). This is ambiguous in two independent ways:

1. **Flattened scope is ambiguous.** `@remix-run-router` could be `@remix-run/router` or `@remix/run-router` — the `/` was replaced by `-` and there is no way to know where it was.
2. **Protocol markers collide with real name segments.** The protocol field (`npm` / `virtual` / `patch` / `file` / `http` / `portal`) is not distinguishable from a legitimate name part: `@dmsnell/diff-match-patch`, `write-file-atomic`, `@types/http-errors`, `is-npm`, etc.

So string-parsing the store folder name is unsound. The robust source of truth is the store package's own `package.json` `name` field, which is authoritative and unambiguous.

#### Reproduction

```sh
mkdir knip-yarn-pnpm && cd knip-yarn-pnpm
yarn init -y
yarn set version berry
printf 'nodeLinker: pnpm\n' >> .yarnrc.yml
# add any dep that gets resolved via realpath (e.g. a storybook/babel addon), then:
yarn add react
npx knip
```

Knip reports `.store` under "Unlisted dependencies" instead of the real package name(s).

#### Fix

I've opened a PR that reads the store package's `package.json` `name` when (and only when) the derived segment is `.store`, with a per-store-root cache so classic `node_modules` layouts never touch the filesystem. It includes a fixture + unit test covering the ambiguous scoped case (`@remix-run/router`), the protocol-collision case (`write-file-atomic`), a path that ends at `/package` with no trailing slash, and graceful fallback to `.store` when no `package.json` is present.

blowery:
Fixes #1851

### Problem

Under **Yarn Berry with `nodeLinker: pnpm`**, Knip reports bogus unlisted dependencies named `.store`.

`getPackageNameFromFilePath` derives the package name from the **last** `node_modules/` segment. That holds for the pnpm CLI (`node_modules/<pkg>/`), but Yarn's pnpm linker lays packages out in a content-addressed store instead:

```
node_modules/.store/<flattened-locator>/package/…
```

There is no nested `node_modules/<pkg>`, so the derived name is literally `.store`. It typically surfaces via plugins that resolve config files through store realpaths (e.g. Storybook/Babel addons via `dirname(require.resolve('<addon>/package.json'))`).

### Why not parse the folder name

The store folder is `<flatIdent>-<protocol>-<reference>-<hash>` with scopes flattened (`@scope/name` → `@scope-name`), which is ambiguous two ways:

1. Flattened scope: `@remix-run-router` could be `@remix-run/router` or `@remix/run-router`.
2. Protocol markers (`npm`/`virtual`/`patch`/`file`/`http`/`portal`) are indistinguishable from real name segments (`@dmsnell/diff-match-patch`, `write-file-atomic`, `@types/http-errors`, `is-npm`).

So the folder name can't be parsed back to a package name reliably. The store package's own `package.json` `name` field is authoritative, so this reads that instead.

### Change

In `getPackageNameFromFilePath`, when — and only when — the derived last segment is `.store`, resolve the name from the store package's `package.json`:

- Matches `…/node_modules/.store/<locator>/package` with a `(?:\/|$)` boundary so it handles both paths that continue past `/package/…` and paths that end exactly at `…/package` (e.g. addon dirs from `dirname(require.resolve('<addon>/package.json'))`).
- Caches per store root, so classic `node_modules` layouts never hit the filesystem and each store package is read at most once.
- Falls back to `.store` (prior behavior) if the `package.json` is missing/unreadable.

This mirrors the existing sync-read + `Map` cache pattern in `src/manifest/helpers.ts`.

### Tests

Added a fixture (`fixtures/yarn-pnpm-store`) laid out as a Yarn pnpm store and unit tests in `test/util/modules.test.ts` covering:

- ambiguous flattened scope → `@remix-run/router`
- protocol-collision unscoped name → `write-file-atomic`
- path ending at `/package` with no trailing slash
- graceful fallback to `.store` when no `package.json` exists
- unchanged behavior for classic `node_modules` paths (including nested)

`pnpm build` (tsgo), `oxlint`, `oxfmt --check`, and the module tests pass locally (node + bun). The approach is validated in a production repo using `nodeLinker: pnpm`.


webpro:
Thanks Ben!

webpro:

Issue #1492, opened 2026-01-28
🧩 Replace `minimist` with `util.parseArgs`


silverwind:
### Discuss anything related to Knip

Stable since Node.js v20: https://nodejs.org/api/util.html#utilparseargsconfig

webpro:
Would love to but

- Knip supports Node.js v18
- minimist !== util.parseArgs

Perhaps in next major.

silverwind:
> minimist !== util.parseArgs

If your arg parsing is not too fancy, it should not be too hard of a easy migration.

> Perhaps in next major.

Node 18 is already EOL, so time to bump imho :)

webpro:
Knip itself uses `parseArgs`, minimist is what all the plugins and package manager resolvers use.

webpro:
> Node 18 is already EOL, so time to bump imho :)

Not each and every project/company has the luxury at all times.

webpro:
Note to self or anyone who likes to work on this; diff include:

- minimist has multi-aliasing (`parseArgs` only `short: string`)
- minimist gives explit `parsed['--']` for everything behind double-dash
- minimist is lax & greedy, `parseArgs` will need some pre/post-processing

Probably justifies major bump as it might break unknown cases out there.

webpro:
:rocket: _This issue has been resolved in v6.16.0. See [Release 6.16.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.16.0) for release notes._

_Using Knip in a commercial project? Please consider [becoming a sponsor](https://knip.dev/sponsors)._

PR #1853, opened 2026-07-03
chore: drop support for Node 20

trueberryless:
#### Description

This PR updates drops support for Node 20 because it has reached [EOL on Mar 24, 2026](https://nodejs.org/en/about/eol). It does so by updating all `"20"` to at least `"22.12.0"` or for test fixtures, just use `*` so we do not need to care about it anymore.

I think as this is a breaking change, this needs to wait for the next major. I'm not sure how you handle major releases. Is there some time windows? Any plans? Or just when the time is right?

If you want to drop a major, I would be fine with it. I could also help writing the blog post for Knip 7 or a section if you want. Just let me know 👍


webpro:
Thanks for the pull request! Would be a breaking change.

Knip's aim is to be useful also for legacy projects. For instance, it [supports CommonJS](https://knip.dev/guides/working-with-commonjs) (while promoting a style that helps migration to ESM) and has plenty of knobs to cater the configuration to any project style/structure. Knip should be helpful in such situations/migrations. I feel this is in conflict with dropping support Node.js v20.

Similar case that pops up in my mind now: #1492.

In practice, not every project has the luxury to bump Node.js itself (in CI and dev env) in a timely manner. I don't really have a timeline in mind and also don't feel that compelled to do the major bump at this point — comes with more maintenance and work for us, for what benefit exactly?

Happy to hear your thoughts on any or all of this!

trueberryless:
This makes sense to me. I value it very much that you aim to support older Node versions so Knip can be used more easily in corporate environments.

At the same time, I wonder how much benefit it really has, given that not even Node itself supports these EOL versions anymore. To me, the answer lies in the definition of "end of life" as it already means that support is not longer guaranteed because that's just how the lifecycle of big products works. However, I want to clarify that I am by far no expert in this matter and I agree that as long as we are not blocked anywhere because we cannot use modern APIs, it really has no benefit to drop support.

So overall, I'm completely fine to just put this one on hold (maybe even close the PR in the meantime) until we have the next major coming up. I think it makes definitely sense to drop support though when the next major is ready feature wise or whatever in maybe 6-12 months because supporting it forever could really hold us back.

webpro:
We should probably do this within a few months or so. Just too bad we don't seem to have the actual usage data so this is a bit of theoretical exercise. But I don't think supporting Node.js 20 is holding us back at this point. You or anyone please holler if I'm wrong here. Let's keep it open to keep for visibility.

PR #1859, opened 2026-07-05
fix(next): resolve src/ entry patterns like Next.js does

dayongkr:
## Problem

The plugin marks files as production entries that Next.js never even picks up, so dead code sits there unreported.

Every entry pattern is added twice, for the root and for `src/`, but Next.js only ever uses one of the two locations per router directory — from the [src folder docs](https://nextjs.org/docs/app/api-reference/file-conventions/src-folder):

> `src/app` or `src/pages` will be ignored if `app` or `pages` are present in the root directory.

Example — a project routes from a root `pages/` directory and keeps the page implementations under `src/pages/`:

```
pages/home.tsx           route (entry)
src/pages/HomePage.tsx   implementation, imported from pages/home.tsx
src/pages/OldPage.tsx    imported by nothing: dead code
```

Next.js ignores `src/pages` here, but the plugin's `src/pages/**` entry pattern marks every file in it as a production entry, so `OldPage.tsx` is never reported as an unused file.

## Fix

When `next.config.*` is present, locate each router directory the way Next.js does ([find-pages-dir.ts](https://github.com/vercel/next.js/blob/00598045032a0e5b313de7b6ef0af60ed9390c2a/packages/next/src/lib/find-pages-dir.ts#L4-L13), unchanged since v12) and scope the `app/` and `pages/` entry patterns to the resolved location: root when the root directory exists, `src/` when only the src one does, both when neither is found.

`pages` and `app` are located independently, since mixing a root `pages/` with `src/app` is valid up to Next.js 15 (the same-parent restriction was added in 16.0).

Root-level file conventions (`middleware`, `proxy`, `instrumentation`, `instrumentation-client`) are out of scope and stay matched in both locations, as before.

## Tests

- `next-root-and-src` (new): root `app/` + `pages/` next to a `src/` folder — `src/app/page.tsx` and `src/pages/legacy.ts` are reported unused; root entries and `src/middleware.ts` are not.
- `next-mixed-routers` (new): root `pages/` combined with `src/app` — all router files stay entries.

### DIFF
--- a/packages/knip/fixtures/plugins/next-mixed-routers/next.config.js
+++ b/packages/knip/fixtures/plugins/next-mixed-routers/next.config.js
@@ -1,0 +1,1 @@
+module.exports = {};
--- a/packages/knip/fixtures/plugins/next-mixed-routers/node_modules/next/package.json
+++ b/packages/knip/fixtures/plugins/next-mixed-routers/node_modules/next/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "next",
+  "bin": "./cli.js"
+}
--- a/packages/knip/fixtures/plugins/next-mixed-routers/package.json
+++ b/packages/knip/fixtures/plugins/next-mixed-routers/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "@plugins/next-mixed-routers",
+  "scripts": {
+    "dev": "next dev"
+  },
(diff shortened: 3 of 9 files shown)


webpro:
Thank you @dayongkr! ✨

webpro:

PR #1860, opened 2026-07-05
Fix Bun test defaults

jakeleventhal:
## Summary

- enable the Bun plugin for Bun lockfile projects as well as direct `bun test` scripts
- parse `bun test` scripts with options before `test`, such as `bun --config=... test`
- use Bun default test globs when the plugin is enabled without an explicit test script, so colocated `*.test.ts` and `*.test.tsx` files are treated as entries
- ignore the package-manager `bun` dependency when the Bun plugin provides defaults

## Validation

- `bun test test/plugins/bun.test.ts test/plugins/bun2.test.ts test/plugins/bun3.test.ts test/plugins/bun4.test.ts test/plugins/bun5.test.ts test/plugins/bun6.test.ts test/plugins/bun7.test.ts`
- `pnpm build`
- `pnpm test --runtime bun --smoke`
- verified against rip-technologies with its `bun.entry` workaround removed via a temporary config: patched Knip no longer reports colocated test files as unused, and no longer reports the root `bun` devDependency as unused


webpro:
Thanks Jake! I think I'm just going to merge this one and then revisit both Node.js and Bun plugins and apply a more consistent approach as each has its own pros/cons. The fixtures should prevent regressions.

webpro:
Just pushed a bunch of refactors around this, feel free to review and/or try it out. Planning to release main tomorrow.

webpro:

PR #1862, opened 2026-07-05
Improve Serverless imported handler detection

jakeleventhal:
## Summary

Improves the Serverless Framework plugin so it can discover production entries from TypeScript configs that cannot be fully loaded but are still statically inspectable.

This adds an AST fallback that:

- reads `functions` object entries that reference imported local function configs
- resolves handler templates based on `__dirname`, such as `` `${handlerPath(__dirname)}/handler.default` ``
- marks local Serverless `${file(...)}` references and esbuild `inject` files as production entries
- keeps the existing loaded-config behavior for simple/literal configs

## Why

Projects can split Serverless function configs into imported modules. If those modules use path aliases or other runtime-only config, loading the full `serverless.ts` can fail or be incomplete, so Knip may miss Lambda handler entrypoints and report reachable files as unused.

## Validation

- `node --test packages/knip/test/plugins/serverless-framework.test.ts`
- `corepack pnpm --dir packages/knip fmt`
- `corepack pnpm --dir packages/knip lint`
- `corepack pnpm --dir packages/knip build`
- `corepack pnpm --dir packages/knip test`

### DIFF
--- a/packages/knip/src/plugins/serverless-framework/helpers.ts
+++ b/packages/knip/src/plugins/serverless-framework/helpers.ts
@@ -1,5 +1,5 @@
-import { toDependency, toProductionEntry } from '../../util/input.ts';
-import { isInternal, join } from '../../util/path.ts';
+import { toDeferResolve, toDeferResolveProductionEntry, toProductionEntry } from '../../util/input.ts';
+import { isInternal } from '../../util/path.ts';
 
 export const handlerToEntry = (handler: string) => {
   const dot = handler.lastIndexOf('.');
@@ -7,4 +7,4 @@ export const handlerToEntry = (handler: string) => {
 };
 
 export const pluginToInput = (plugin: string, dir: string) =>
-  isInternal(plugin) ? toProductionEntry(join(dir, plugin)) : toDependency(plugin);
+  isInternal(plugin) ? toDeferResolveProductionEntry(plugin, { dir }) : toDeferResolve(plugin, { dir });
--- a/packages/knip/src/plugins/serverless-framework/helpers.ts
+++ b/packages/knip/src/plugins/serverless-framework/helpers.ts
@@ -1,0 +1,10 @@
+import { toDependency, toProductionEntry } from '../../util/input.ts';
+import { isInternal, join } from '../../util/path.ts';
+
+export const handlerToEntry = (handler: string) => {
+  const dot = handler.lastIndexOf('.');
+  return toProductionEntry(`${handler.slice(0, dot)}.{js,ts}`);
+};
+
+export const pluginToInput = (plugin: string, dir: string) =>
+  isInternal(plugin) ? toProductionEntry(join(dir, plugin)) : toDependency(plugin);
--- a/packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
+++ b/packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
@@ -1,6 +1,15 @@
-import { Visitor, type Program } from 'oxc-parser';
+import type {
+  BindingIdentifier,
+  Expression,
+  IdentifierName,
+  IdentifierReference,
+  Node,
+  ObjectExpression,
+  Program,
+  TemplateLiteral,
+} from 'oxc-parser';
 import type { ResolveFromAST } from '../../types/config.ts';
-import { collectPropertyValues, getImportMap, getPropertyKey, getStringValues } from '../../typescript/ast-helpers.ts';
+import { getImportMap, getPropertyKey, getStringValues } from '../../typescript/ast-helpers.ts';
 import { _parseFile, getStringValue } from '../../typescript/ast-nodes.ts';
 import type { Input } from '../../util/input.ts';
 import { toDependency, toProductionEntry } from '../../util/input.ts';
@@ -10,12 +19,38 @@ import { handlerToEntry, pluginToInput } from './helpers.ts';
 
 const serverlessFileReference = /\$\{file\(([^):]+)\)(?::[^}]*)?\}/g;
 
-const unwrapExpression = (node: any): any => {
+type Identifier = BindingIdentifier | IdentifierName | IdentifierReference;
+
+const isIdentifier = (node: Node | null | undefined, name?: string): node is Identifier =>
+  node?.type === 'Identifier' && (name === undefined || node.name === name);
+
+const isNode = (value: unknown): value is Node =>
+  value !== null &&
+  typeof value === 'object' &&
+  'type' in value &&
+  typeof (value as { type?: unknown }).type === 'string';
+
+const walkNode = (node: Node, visit: (node: Node) => void, seen = new Set<Node>()) => {
+  if (seen.has(node)) return;
+  seen.add(node);
+  visit(node);
+
+  for (const [key, value] of Object.entries(node)) {
+    if (key === 'parent') continue;
+    if (Array.isArray(value)) {
+      for (const item of value) if (isNode(item)) walkNode(item, visit, seen);
+    } else if (isNode(value)) {
+      walkNode(value, visit, seen);
+    }
+  }
+};
+
+const unwrapExpression = (node: Expression): Expression => {
   if (
-    node?.type === 'ParenthesizedExpression' ||
-    node?.type === 'TSAsExpression' ||
-    node?.type === 'TSSatisfiesExpression' ||
-    node?.type === 'TSTypeAssertion'
+    node.type === 'ParenthesizedExpression' ||
+    node.type === 'TSAsExpression' ||
+    node.type === 'TSSatisfiesExpression' ||
+    node.type === 'TSTypeAssertion'
   ) {
     return unwrapExpression(node.expression);
   }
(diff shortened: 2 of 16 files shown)


packages/knip/test/plugins/serverless-framework.test.ts
```diff
@@ -30,3 +31,15 @@ test('Find dependencies from Serverless Framework TypeScript plugins', async ()
...
+test('Find imported TypeScript function handlers from Serverless Framework config', async () => {
+  const options = await createOptions({ cwd: typescriptImportedFunctionsCwd });
+  const { issues } = await main(options);
+
+  assert(!('scripts/cjs-shim.mjs' in issues.files));
+  assert(!('scripts/secrets.cjs' in issues.files));
+  assert(!('src/functions/process-new-file/handler.ts' in issues.files));
+  assert(!('src/functions/process-new-file/message.ts' in issues.files));
+  assert('src/functions/process-new-file/unused.ts' in issues.files);
+  assert.deepEqual(issues.devDependencies, {});
```
webpro: Please use `counters` like other tests do, then if there's e.g. `files: 1` we don't need any of the negated `issues.files` assertions. Same for empty `devDependencies` obj.

packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
```diff
@@ -0,0 +1,164 @@
...
+import { _parseFile, getStringValue } from '../../typescript/ast-nodes.ts';
+import type { Input } from '../../util/input.ts';
+import { toDependency, toProductionEntry } from '../../util/input.ts';
+import { dirname, join, relative, toAbsolute } from '../../util/path.ts';
+import { _resolveSync } from '../../util/resolve.ts';
+import { handlerToEntry, pluginToInput } from './helpers.ts';
+
+const serverlessFileReference = /\$\{file\(([^):]+)\)(?::[^}]*)?\}/g;
+
+const unwrapExpression = (node: any): any => {
```
webpro: Please avoid `any` and use actual oxc-parser types (see other plugins or AST helpers if you need inspiration).

packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
```diff
@@ -0,0 +1,164 @@
...
+  if (
+    node?.type === 'ParenthesizedExpression' ||
+    node?.type === 'TSAsExpression' ||
+    node?.type === 'TSSatisfiesExpression' ||
+    node?.type === 'TSTypeAssertion'
```
webpro: Once we have a type, we probably no longer need optional chaining.

packages/knip/src/plugins/serverless-framework/helpers.ts
```diff
@@ -0,0 +1,10 @@
+import { toDependency, toProductionEntry } from '../../util/input.ts';
+import { isInternal, join } from '../../util/path.ts';
+
+export const handlerToEntry = (handler: string) => {
+  const dot = handler.lastIndexOf('.');
+  return toProductionEntry(`${handler.slice(0, dot)}.{js,ts}`);
+};
+
+export const pluginToInput = (plugin: string, dir: string) =>
+  isInternal(plugin) ? toProductionEntry(join(dir, plugin)) : toDependency(plugin);
```
webpro: You might want to try [toDeferResolve](https://knip.dev/writing-a-plugin/inputs#todeferresolve) for simpler and more robust resolution (accepts `{ dir }` option).

packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
```diff
@@ -0,0 +1,164 @@
...
+const getHandlerInputsFromProgram = (
+  program: Program,
+  containingFilePath: string,
+  options: Parameters<ResolveFromAST>[1]
+) => {
+  const inputs: Input[] = [];
+  const visitor = new Visitor({
+    Property(node) {
+      if (getPropertyKey(node) === 'handler') inputs.push(...getHandlerInputs(node.value, containingFilePath, options));
+    },
```
webpro: This looks like too broad of a matcher (e.g. _any_ `handler` prop value might become a production entry? literally any `plugin` value a dependency?) Pease try to scope things to the serverless schema.

packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    const resolvedFilePath = toAbsolute(resolvedPath, options.cwd);
+    if (visited.has(resolvedFilePath)) return;
+
+    visited.add(resolvedFilePath);
+
+    const sourceText = options.readFile(resolvedFilePath);
+    if (!sourceText) return;
+
+    const importedProgram = _parseFile(resolvedFilePath, sourceText).program;
```
webpro: Please wrap reading and parsing within `resolveFromAST` in try/catch.

webpro (commented):
Thanks for the pull request! To be honest the ROI is a tad steep here and might need another round of review from my end, but here's a first review.

packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
jakeleventhal: Done in https://github.com/webpro-nl/knip/pull/1862/commits/03a2a710045797cdc636c52b583f88dadb6aa532 by scoping AST extraction to the exported Serverless config shape, then resolving handlers, plugins, file references, and esbuild inject values through the relevant schema properties instead of matching every property in the file.

packages/knip/src/plugins/serverless-framework/helpers.ts
jakeleventhal: Done in https://github.com/webpro-nl/knip/pull/1862/commits/03a2a710045797cdc636c52b583f88dadb6aa532 by changing plugin resolution to use `toDeferResolve` / `toDeferResolveProductionEntry` with the config directory.

packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
jakeleventhal: Done in https://github.com/webpro-nl/knip/pull/1862/commits/03a2a710045797cdc636c52b583f88dadb6aa532 by replacing the `any`-based AST helpers with concrete `oxc-parser` node types and local type guards.

packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
jakeleventhal: Done in https://github.com/webpro-nl/knip/pull/1862/commits/03a2a710045797cdc636c52b583f88dadb6aa532 by removing the optional chaining from typed template literal and expression handling where the OXC types guarantee the fields.

packages/knip/src/plugins/serverless-framework/resolveFromAST.ts
jakeleventhal: Done in https://github.com/webpro-nl/knip/pull/1862/commits/03a2a710045797cdc636c52b583f88dadb6aa532 by wrapping imported function config file reads and parsing in `try/catch`.

jakeleventhal (commented):
updated

jakeleventhal:
good for another look @webpro

webpro:
Let's go back to the "why" of this PR:

> Projects can split Serverless function configs into imported modules. If those modules use path aliases or other runtime-only config, loading the full `serverless.ts` can fail or be incomplete, so Knip may miss Lambda handler entrypoints and report reachable files as unused.

Overall, this would be the only plugin that always runs both `resolveConfig` and `resolveFromAST` function (basically fully redundant). This PR tries to solve a generic problem for one specific plugin. Also see https://knip.dev/reference/known-issues.

So I'd like to see first exactly the root motivation to open this pull request.

As it stands, I feel like only `resolveConfig` including a bit of `${file(...)}` parsing could get us pretty far?

jakeleventhal:
@webpro maybe you have a better sense for general direction?

What zooming out even further to the true cause/intent:

My `knip.json` looks like this:
```json
{
	"$schema": "./node_modules/knip/schema.json",
	"workspaces": {
		"apps/myapp/sls-api": {
			"entry": [
				"src/functions/*/handler.ts"
			]
		},
	}
}
```

In my repo, I have a file structure like this:
```
knip.json
apps/
  myapp/
    sls-api/
      serverless.ts
      src/functions/
        myhandler1/
          index.ts   → imports the handler.ts
          handler.ts
        myhandler2/
          index.ts   → imports the handler.ts
          handler.ts
        myhandler3/
          index.ts   → imports the handler.ts
          handler.ts
```

Here is what my `serverless.ts` file looks like:
```ts
import { myHandler1 } from "./src/functions/myhandler1";
import { myHandler2 } from "./src/functions/myhandler2";
import { myHandler3 } from "./src/functions/myhandler3";

...

export default {
  ...
  functions: {
    myHandler1,
    myHandler2,
    myHandler3,
  },
  ...
};
```

---

My goal is to have this be 0-config in `knip.json` and discover the handlers as entrypoints.

webpro:
Since `serverless.ts` is in the module graph (`config` files are auto-added as `entry`), then through regular import chain `./src/functions/myhandler1` will find `./src/functions/myhandler1/index.ts` which will import `./handler.ts` — these files should all be covered fine. This is without any config, provided `serverless` is listed as a dependency in `package.json`. What are we missing here? Can you please create a full repro that currently does not what you expect?

jakeleventhal:
Created a minimal repro here with generic names:

https://github.com/jakeleventhal/knip-serverless-handler-repro

I tightened this after the first pass so it mirrors the repo shape more closely:

- monorepo workspace at `apps/example/sls-api`
- `serverless.ts` exports `getServerlessConfiguration()`
- function configs are imported into `functions: { ... } satisfies Record<LambdaFunctionName, LambdaConfig>`
- function config modules use `~/*` path aliases
- handlers are referenced via `` `${handlerPath(__dirname)}/handler.default` ``

Run:

```sh
git clone https://github.com/jakeleventhal/knip-serverless-handler-repro
cd knip-serverless-handler-repro
npm install
npm run knip
```

Current Knip (`6.25.0`) reports:

```text
Unused files (2)
apps/example/sls-api/src/functions/image-resize/handler.ts
apps/example/sls-api/src/functions/queue-sync/handler.ts
```

The missing part is that the normal module graph only gets as far as the imported function config modules:

```ts
import { imageResize } from "./src/functions/image-resize";
```

Those `index.ts` files do not import their `handler.ts` files. They import supporting config through `~/*`, then expose Serverless handler strings:

```ts
handler: `${handlerPath(__dirname)}/handler.default`
```

So I think you're right that `serverless.ts -> ./src/functions/image-resize -> index.ts` is covered by regular imports, but the final `index.ts -> handler.ts` edge is only visible through Serverless handler-string resolution.

I also verified this PR's built CLI against the repro and it exits with no unused files:

```sh
knip --workspace apps/example/sls-api --include files
```

jakeleventhal:
Screenshot 2026-07-08 at 12 00 10 PM
This is a better approach in my repo that works. Closing

webpro:
OK, cool. Feel free to check out this branch I just pushed that does the `file` handling: https://github.com/webpro-nl/knip/blob/serverless-imported-handlers/packages/knip/src/plugins/serverless-framework/index.ts — happy to consider merging that if it helps. It seems to handle your repro well.

jakeleventhal:
@webpro my use case is solved so your call

PR #1857, opened 2026-07-04
chore(ci): add concurrency blocks

trueberryless:
#### Description

This PR adds concurrency blocks to every workflow. This has the effect of cancelling old workflows on the same branch/PR which are old because a new commit has already landed on it, and therefore only the status checks of the new commit is interesting, the old ones can be cancelled.

Discussed in #1849

PR #1863, opened 2026-07-05
chore(ci): remove redundant tag triggers

trueberryless:
#### Description

This PR removes all the "anti" tag triggers, which were added [in this commit](https://github.com/webpro-nl/knip/commit/66a3d31554da21b2b22fcf42b9bd1beeaf77415b) to not trigger CI on tag creations, but because the triggers already specify that they should only trigger on branches, this `!` trigger is redundant.

This is the last improvement I have planned after discussing about CI in #1849

### DIFF
--- a/.github/workflows/ci-bun.yml
+++ b/.github/workflows/ci-bun.yml
@@ -6,8 +6,6 @@ on:
   push:
     branches:
       - '**'
-    tags:
-      - '!**'
 
 # Automatically cancel in-progress actions on the same branch
 concurrency:
--- a/.github/workflows/ci-legacy-5.0.4.yml
+++ b/.github/workflows/ci-legacy-5.0.4.yml
@@ -8,8 +8,6 @@ on:
   push:
     branches:
       - '**'
-    tags:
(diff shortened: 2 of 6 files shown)


webpro:
Thanks! I vaguely remember trying to fix that with a release the pipelines runs twice. Does that make sense?

trueberryless:
> Thanks! I vaguely remember trying to fix that with a release the pipelines runs twice. Does that make sense?

I'm not that familiar with `release-it`, does it create some temporary branch when releasing? That could be a reason for why it runs twice. But I strongly believe that the current `tag` triggers do nothing because it only runs on branches, so it ignores tags already. 

I mean worst case is that we must revert this, but I highly doubt that this would be necessary...

webpro:
This is separate from release-it. It's just that we push a branch (i.e. `main`) and tag on release and I think it ran everything twice. I can try it with the next release.

webpro:

webpro:
Nice, redundant indeed! And now we would likely even still be covered by #1857.

PR #1864, opened 2026-07-05
Add eve (Vercel framework) plugin

jakeleventhal:
## What changed

Adds a Knip plugin for `eve`, enabled when the `eve` package is present. The plugin marks Eve's filesystem-discovered agent files as production entries, including root/flat agent layouts, authored runtime slots (`tools`, `skills`, `channels`, `connections`, `hooks`, `schedules`, sandbox files), and nested subagent slots.

The fixture intentionally includes an unused `agent/lib/unused.ts` file so the test verifies the plugin does not glob every TypeScript file under `agent/` as an entry.

## Why

Eve discovers agent capabilities from the filesystem rather than from a central import registry. Without a Knip plugin, real Eve apps have to add broad project-level entry globs such as `agent/**/*.ts`, which prevents Knip from reporting genuine unused helper files/exports under the agent tree.

## Validation

- `bun test test/plugins/eve.test.ts`
- `pnpm build`


packages/knip/src/plugins/eve/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const agentFileNames = '{agent,instructions,instrumentation,sandbox}';
+
+const production = [
+  `{,agent/}${agentFileNames}.${extensions}`,
+  `{,agent/}sandbox/sandbox.${extensions}`,
+  `{,agent/}{channels,connections,hooks,skills,tools,schedules}/**/*.${extensions}`,
+  `{,agent/}subagents/**/${agentFileNames}.${extensions}`,
+  `{,agent/}subagents/**/sandbox/sandbox.${extensions}`,
+  `{,agent/}subagents/**/{connections,hooks,skills,tools}/**/*.${extensions}`,
```
webpro, line 24: Subagent slots here are `{connections,hooks,skills,tools}`, but the root-level slots on line 19 also include `channels` and `schedules`. If Eve auto-discovers those inside subagents (e.g. `agent/subagents/x/schedules/daily.ts` or `.../channels/slack.ts`), they won't be marked as entries and will surface as false-positive unused files. If subagents genuinely can't own channels/schedules, could you confirm that against Eve's project-layout docs? The asymmetry isn't obvious from the code alone.

packages/knip/src/plugins/eve/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const title = 'eve';
+
+const enablers = ['eve'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const extensions = '{js,jsx,ts,tsx,mjs,cjs,mts,cts}';
+
+const agentFileNames = '{agent,instructions,instrumentation,sandbox}';
```
webpro: `instructions` and `instrumentation` aren't in the slot list from the PR description, and no fixture covers either (`instrumentation` looks like it may have carried over from the Next plugin's `rootOrSrc` line). If they aren't real Eve-discovered filenames, these globs silently broaden the entry set and can mask genuinely-unused `instructions.ts`/`instrumentation.ts` files — the exact false-negative this plugin is meant to prevent. Could you confirm they're real Eve conventions, and add fixture coverage for the untested slots (`skills`, `instructions`, `instrumentation`, and the subagent variants) either way?

packages/knip/src/plugins/eve/index.ts
jakeleventhal: Done in https://github.com/jakeleventhal/knip/commit/56b27db1340823216855cd6320d8937b92a1ec74. I confirmed Eve project-layout docs list channels and schedules as root-only, so the plugin keeps subagent slots limited to connections, hooks, skills, tools, and sandbox/instructions/agent files. The fixture now includes unsupported subagent channels/schedules files and asserts they remain reported as unused.

packages/knip/src/plugins/eve/index.ts
jakeleventhal: Done in https://github.com/jakeleventhal/knip/commit/56b27db1340823216855cd6320d8937b92a1ec74. Eve docs confirm instructions.ts and root instrumentation.ts are real conventions, with instrumentation root-only. I split root/subagent filename globs so subagent instrumentation no longer gets masked, and added fixture coverage for root skills/instructions/instrumentation plus subagent connections/hooks/instructions/sandbox/skills and unsupported subagent instrumentation.

webpro:
You're on a roll Jake, much appreciated!

webpro:

PR #1869, opened 2026-07-06
Add support for Prettier overrides

remcohaszing:
Prettier overrides may define plugins in addition to the top-level config.

### DIFF
--- a/packages/knip/src/plugins/prettier/index.ts
+++ b/packages/knip/src/plugins/prettier/index.ts
@@ -1,8 +1,8 @@
 import type { Args } from '../../types/args.ts';
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.ts';
-import { toDeferResolve, toDependency } from '../../util/input.ts';
+import { toDeferResolve, toDependency, type Input } from '../../util/input.ts';
 import { hasDependency } from '../../util/plugin.ts';
-import type { PrettierConfig } from './types.ts';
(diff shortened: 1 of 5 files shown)


packages/knip/src/plugins/prettier/index.ts
```diff
@@ -23,9 +23,28 @@ const config = [
...
-  return Array.isArray(config.plugins)
-    ? config.plugins.filter((plugin): plugin is string => typeof plugin === 'string').map(id => toDependency(id))
-    : [];
+  const result = new Set<Input>();
+
+  const processOptions = (options: PrettierOptions) => {
+    if (Array.isArray(options.plugins)) {
+      for (const plugin of options.plugins) {
+        if (typeof plugin === 'string') {
+          result.add(toDependency(plugin));
```
webpro, line 32: Bit of a minor thing, but `Input` is an object so using a `Set` does not dedupe here.

webpro (commented):
Thanks for the pull request! Happy to merge, just one non-blocking nit.

packages/knip/src/plugins/prettier/index.ts
remcohaszing: Right, that makes sense. Does deduplication even matter? It felt like the clean approach, but it probably doesn’t matter at all.

webpro: Eventually Knip itself dedupes. Anyway let's roll with it.

webpro:

PR #1854, opened 2026-07-03
docs: migrate to Starlight Blog plugin

trueberryless:
#### Description

This PR updates the docs to use the [Starlight Blog plugin](https://starlight-blog-docs.vercel.app/), which enhances how we can write blog posts, by allowing defined metadata, better overviews, automatically ordering the blog posts, ... (read more in the docs of the plugin).

PR #1870, opened 2026-07-06
feat: add support for Starlight plugins config dep detection

trueberryless:
#### Description

This PR adds support for [`plugins`](https://starlight.astro.build/reference/configuration/#plugins) in the Starlight Knip plugin.

- [x] I used AI to generate the code because I was playing around with it on the weekend for some time and couldn't figure out how to create a good test case. TIL there is the `isProduction` and `isStrict` option 🙌

### DIFF
--- a/packages/knip/src/plugins/starlight/index.ts
+++ b/packages/knip/src/plugins/starlight/index.ts
@@ -1,9 +1,9 @@
 import type { IsPluginEnabled, Plugin, ResolveFromAST } from '../../types/config.ts';
 import { findCallArg, getDefaultImportName, getImportMap, getPropertyValues } from '../../typescript/ast-helpers.ts';
-import { toProductionEntry } from '../../util/input.ts';
+import { toDependency, toProductionEntry } from '../../util/input.ts';
 import { hasDependency } from '../../util/plugin.ts';
 import { config } from '../astro/index.ts';
-import { getComponentPathsFromSourceFile } from './resolveFromAST.ts';
+import { getCalleeImportSources, getComponentPathsFromSourceFile } from './resolveFromAST.ts';
 
 // https://starlight.astro.build/reference/configuration/
 
@@ -26,6 +26,11 @@ const resolveFromAST: ResolveFromAST = program => {
       const customCssPaths = getPropertyValues(starlightConfig, 'customCss');
       for (const id of customCssPaths) {
         inputs.push(toProductionEntry(id));
+      }
+
+      const pluginSources = getCalleeImportSources(starlightConfig, 'plugins', importMap);
+      for (const source of pluginSources) {
+        inputs.push(toDependency(source));
       }
     }
   }
--- a/packages/knip/test/plugins/starlight.test.ts
+++ b/packages/knip/test/plugins/starlight.test.ts
@@ -30,3 +30,19 @@ test('Find custom CSS with the starlight plugin', async () => {
     files: 1,
   });
 });
+
+test('Find plugins with the starlight plugin (production strict)', async () => {
+  const cwd = resolve('fixtures/plugins/starlight/plugins');
+  const options = await createOptions({ cwd, isProduction: true, isStrict: true });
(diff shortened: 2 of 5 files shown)


packages/knip/src/plugins/starlight/index.ts
```diff
@@ -27,6 +27,11 @@ const resolveFromAST: ResolveFromAST = program => {
       for (const id of customCssPaths) {
         inputs.push(toProductionEntry(id));
       }
+
+      const pluginSources = getCalleeImportSources(starlightConfig, 'plugins', importMap);
+      for (const source of pluginSources) {
+        inputs.push(toDependency(source));
+      }
```
webpro, line 34: If the intention is to find dependencies, then the files being in the module graph (e.g. `astro.config.mjs` is a `config` → also `entry` → in module graph) is enough for Knip core to find the imports and see that `@astrojs/starlight` and `starlight-blog` are referenced. This is generic/core behavior that should not require a plugin.

packages/knip/test/plugins/starlight.test.ts
```diff
@@ -30,3 +30,19 @@ test('Find custom CSS with the starlight plugin', async () => {
...
+  const { issues, counters } = await main(options);
+
+  assert(
+    !issues.dependencies['package.json']?.['starlight-blog'],
+    'starlight-blog should be detected as used via starlight `plugins`'
+  );
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    dependencies: 2,
```
webpro, line 46: Odd that asserts `2`, I would expect only ~`1` (i.e. `astro`)~ `0`.

webpro (commented):
Thanks for the pull request! Though I wonder if we really need it?

trueberryless:
I think we need it. If you run the tests without the second commit, the tests will fail with `starlight-blog should be detected as used via starlight plugins`. And since #1854 shows that it is [failing in CI](https://github.com/webpro-nl/knip/actions/runs/28665859394/job/85017299881?pr=1854), there has to be something that is missing.

But tbh, I am not that familiar with all the parts of Knip yet. As far as I understand, the error is only happening when running with `production` mode. Maybe that is the issue for the failing CI. But I assumed we shouldn't change how we call Knip in docs 🤔

webpro:
Yes your instinct is right. The failure is in production mode. This means that unused `dependencies` are not used in production code; and thus should go into `devDependencies`.

webpro:
This is also why we have `entry` _and_ `production` patterns in plugins, `toProductionEntry`, `toProductionDependency`, etc. (https://knip.dev/writing-a-plugin/inputs). If we get this right in each plugin, then Knip can sort it all out.

trueberryless:
Oooooh, wait, does this mean "starlight-blog" should be a devDep? I have never thought about that possibility because in Starlight, all the plugins are normal dependencies.

Like e.g. the Blog plugin recommends installing as normal dep: https://starlight-blog-docs.vercel.app/getting-started/#installation

I wonder if we should still do something about this nonetheless. Because AFAIK all the Starlight plugin docs install normally, so Knip will report for all of the projects using a plugin that it is unused. Could we improve this somehow?

webpro:
The stance Knip has: if a package does not need to be _installed_ in production, it's not a production dependency. So e.g. a bundler or static website generator (plugin) is typically a dev dep.

Why should Knip change this behavior? It helps with dependency management and optimizing install speed etc. If one thinks it's too rigid then don't opt in to it. But Knip itself dogfoods like there's no tomorrow.

trueberryless:
> Why should Knip change this behavior? It helps with dependency management and optimizing install speed etc. If one thinks it's too rigid then don't opt in to it. But Knip itself dogfoods like there's no tomorrow.

I agree that Knip shouldn't change it. I was thinking about maybe adding some "hint" in the output if it is possible somehow, that those deps can be moved to dev deps. But I guess this is a general thing when running Production Mode, so closing this PR is the right call 👍

PR #1871, opened 2026-07-06
ci: add path filters

trueberryless:
#### Description

This PR adds path filters to all GitHub workflows to exit CI early when run is not necessary.

Here is some data of past runs from the last month and how much runs could have been skipped:

| Workflow | Runs sampled | Would skip | Skip % | Avg dur | Min saved |
| :--- | :--- | :--- | :--- | :--- | :--- |
| `ci.yml` | 84 | 13 | 15.5 % | 3.5 min | 50 |
| `ci-bun.yml` | 84 | 13 | 15.5 % | 7.5 min | 44 |
| `ci-legacy-5.0.4.yml` | 83 | 14 | 16.9 % | 12.5 min | 184 |
| `integration.yml` | 84 | 13 | 15.5 % | 3.1 min | 38 |
| Total/Avg | 335 | 53 | 15.8 % | 6.7 min | ~316 min / month |


webpro:
This is great! Path filters LGTM. What is the difference with setting e.g. `on.*.paths` at the root level of a workflow config?

trueberryless:
> This is great! Path filters LGTM. What is the difference with setting e.g. `on.*.paths` at the root level of a workflow config?

I knew/hoped you would ask that 😅

The difference is that with `on.*.paths` the workflow is not even triggered whereas here it reports a successful status. This is important if you add a branch ruleset which has those workflows marked as required before merging. Doing it with `on.*.paths` would lead to PRs, where the CI should not trigger, hanging because the requirements are not met.

Does that make sense?

webpro:
Thanks for clearing that up! The "if" is doing a lot of work here. Feels like we don't yet have any trait that warrant pulling in the extra complexity (actions + config). How about starting out with `on.*.paths` and expand on this idea as we go?

trueberryless:
> Thanks for clearing that up! The "if" is doing a lot of work here. Feels like we don't yet have any trait that warrant pulling in the extra complexity (actions + config). How about starting out with `on.*.paths` and expand on this idea as we go?

Good to know. I'll adapt it for now and if you enable auto-merge in the future, we just need to remember to change that 👍

webpro:
Thank you! 🙏 This looks great.

webpro:
:rocket: _This pull request is included in v6.26.0. See [Release 6.26.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.26.0) for release notes._

Issue #1877, opened 2026-07-12
💡 Support project-defined conditions beyond the built-in resolver fallbacks


popelenkow:
### Suggest an idea for Knip

Source-first monorepos may expose TypeScript sources to repository tooling through a project-defined condition while keeping compiled output for package consumers:

```json
{
  "exports": {
    ".": {
      "monorepo": "./src/index.ts",
      "types": "./dist/index.d.ts",
      "import": "./dist/index.js",
      "default": "./dist/index.js"
    }
  }
}
```

Knip's main resolver activates `require`, `import`, and `node`; its fallback activates `require`, `import`, and `browser`. Neither resolver activates `monorepo`, so both select a `dist` target instead of the available source.

I reproduced this with `knip@6.26.0`. A temporary backport of the proposed configuration and a working validation are available in [PR #667](https://github.com/musetric/musetric/pull/667).

The same project-defined condition can already be configured for other tools:

```ts
// Vite
conditions: ['monorepo', ...defaultClientConditions];

// eslint-import-resolver-typescript
conditionNames: ['monorepo', ...defaultConditionNames];
```

### Related issue

This is related to [issue #1873](https://github.com/webpro-nl/knip/issues/1873), which was recently resolved in Knip 6.26.0 by adding a browser-condition fallback. That fix handles the reported map where `node` points to unbuilt output and `default` points to source: the fallback omits `node`, so the source target can be selected.

However, changing from the `node` condition to the `browser` fallback cannot activate an arbitrary project-defined condition. The fix for #1873 therefore does not cover the configuration above.

### Prior art

- [Vite documents `resolve.conditions`](https://vite.dev/config/shared-options.html#resolve-conditions) as a configurable `string[]` and exports `defaultClientConditions` for users that want to extend its non-inherited default list.
- [`eslint-import-resolver-typescript` documents `conditionNames`](https://github.com/import-js/eslint-import-resolver-typescript#conditionnames)  as a resolver option and explicitly exports `defaultConditionNames` for reuse in custom configurations.

### Proposed configuration

Expose the `conditionNames` option of `oxc-resolver` as a top-level Knip
configuration option:

```json
{
  "conditions": ["monorepo", "require", "import", "node", "default"]
}
```

The configured array should replace Knip's built-in condition names for all module resolvers. When the option is omitted, existing resolver-specific node and browser behavior should remain unchanged. Package map `default` targets remain fallbacks according to conditional exports/imports semantics.

webpro:
[...]

Issue #1873, opened 2026-07-07
💡 Configurable resolver conditionNames


jasonsliu:
### Suggest an idea for Knip

Knip resolves modules with oxc-resolver and hardcodes the condition names ([resolve.ts#L19](https://github.com/webpro-nl/knip/blob/4249935adffe1b8eca9570fb325fa19cc8010584/packages/knip/src/util/resolve.ts#L19) and [resolve.ts#L53](https://github.com/webpro-nl/knip/blob/4249935adffe1b8eca9570fb325fa19cc8010584/packages/knip/src/util/resolve.ts#L53)):
```
conditionNames: ['require', 'import', 'node', 'default'],
```

In a repo that compiles with moduleResolution: "bundler", source is resolved through package.json imports/exports maps. Our repo serves built output to Node and source to everything else:
```
{ "imports": { "#internal/*": { "node": "./dist/*.js", "default": "./src/*.ts" } } }
```

Because node is in the hardcoded conditionNames and appears first, oxc-resolver selects ./dist/*.js. In a source-first setup that output doesn't exist, so every #internal/* subpath import is reported as unresolved.

Can we add a config option that overrides/prepends the hardcoded list?

webpro:
My read is that if we just remove `node` from the fallback resolver we don't regress anything — and we don't need custom configuration on the user's end. I'll push a fix and you could verify using the preview release.

jasonsliu:
Yep, this works. Thanks you!

webpro:
:rocket: _This issue has been resolved in v6.26.0. See [Release 6.26.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.26.0) for release notes._

_Using Knip in a commercial project? Please consider [becoming a sponsor](https://knip.dev/sponsors)._

PR #1878, opened 2026-07-12
Add conditions config for package export and import resolution

popelenkow:
Adds a top-level `conditions` option that replaces the condition names passed to Knip's module resolvers. Existing node and browser defaults are preserved when the option is omitted.

This follows up on #1873 and #1877, whose browser fallback handles built-in conditions but cannot activate project-defined conditions such as `source` or `monorepo`.

- Add the option to the configuration types and schemas.
- Create an isolated resolver per run and pass it through module, script, plugin, and watch paths.
- Document the option and cover custom-condition resolution with a fixture.


webpro:
Closing this with #1b0a2c0. See #1877.

PR #1881, opened 2026-07-13
Fix crash in Vue plugin's auto-import compiler on Vue 2.7

mfillon:
> Drafted with an AI coding agent (Claude Code); I reviewed the diff, reproduced the crash and confirmed the fix against a Vue 2.7-shaped compiler-sfc mock, and ran the QA checklist from `DEVELOPMENT.md` (`fmt`, `lint`, `knip`, `knip --strict`, `test`) before opening this.

## Root cause

Vue 2.7's `vue/compiler-sfc` has a different API than Vue 3's — it takes a single options object (parse({source, filename})) and returns the SFCDescriptor directly, whereas Vue 3's parse(source, options) returns {descriptor, errors}. Knip's `packages/knip/src/plugins/_vue/auto-import.ts:12` blindly returned the raw `vue/compiler-sfc` module assuming the Vue 3 shape, so on Vue 2.7 `const { descriptor } = sfc.parse(source, path)` destructured a nonexistent property, and the crash only surfaced when auto-import maps were non-empty (i.e. with `unplugin-vue-components`/`unplugin-auto-import` `generated .d.ts` files) — matching the report exactly.

## Fix

`packages/knip/src/plugins/_vue/auto-import.ts:12` now checks the resolved vue package's version; for Vue 2.x it wraps the call to use the options-object signature and normalizes the return value into `{ descriptor }`.

## Test

added `packages/knip/test/plugins/unplugin-vue-components-vue2.test.ts` with a fixture mocking Vue 2.7's real `compiler-sfc` API shape (single-options parse, unwrapped descriptor return). Verified it reproduces the exact reported `TypeError: Cannot read properties of undefined (reading 'script')` without the fix, and passes with it.

Also added a reference to `DEVELOPMENT.md` guidelines in `AGENTS.md`


webpro:
Thanks for the pull request and reproduction, it has been very useful! Closing this in favor of 2157f3913f77216d268b2255899b6fc58609f9bb. The fixture mock here adds template.ast to the Vue 2.7 descriptor, but the real Vue 2.7 SFC parser doesn't provide that AST. This patch prevents the crash, but template-only component and auto-import usage isn't detected against real Vue 2.7. The replacement uses the actual Vue 2.7 API shape, compiles the template when no descriptor AST is available, and covers both component tags and template expressions.

mfillon:
> Thanks for the pull request and reproduction, it has been very useful! Closing this in favor of [2157f39](https://github.com/webpro-nl/knip/commit/2157f3913f77216d268b2255899b6fc58609f9bb). The fixture mock here adds template.ast to the Vue 2.7 descriptor, but the real Vue 2.7 SFC parser doesn't provide that AST. This patch prevents the crash, but template-only component and auto-import usage isn't detected against real Vue 2.7. The replacement uses the actual Vue 2.7 API shape, compiles the template when no descriptor AST is available, and covers both component tags and template expressions.

Ok, thanks for the explanations and more robust fix!

webpro:
:rocket: _This issue has been resolved in v6.27.0. See [Release 6.27.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.27.0) for release notes._

PR #1887, opened 2026-07-16
feat: markdownlint-cli2

Zamiell:
This PR adds support for markdownlint-cli2, which is considered better than markdownlint-cli.

I had 5.6 Sol help me out with this.

### DIFF
--- a/packages/knip/src/plugins/markdownlint/index.ts
+++ b/packages/knip/src/plugins/markdownlint/index.ts
@@ -13,13 +13,16 @@ const enablers = ['markdownlint-cli', 'markdownlint-cli2'];
 
 const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
 
+const isBundled = (specifier: string) =>
+  specifier.startsWith('markdownlint/style/') || specifier === 'markdownlint-cli2-formatter-default';
+
 const config = ['.markdownlint-cli2.{jsonc,yaml,cjs,mjs}', '.markdownlint.{json,jsonc,yaml,yml,cjs,mjs}'];
 
 const resolveConfig: ResolveConfig<MarkdownlintConfig> = (config, options) => {
   const { manifest } = options;
   const dependencies: string[] = [];
   for (const extend of [config?.extends, config?.config?.extends]) {
-    if (extend && !extend.startsWith('markdownlint/style/')) dependencies.push(extend);
+    if (extend) dependencies.push(extend);
   }
   for (const customRule of config?.customRules ?? []) {
     if (typeof customRule === 'string') dependencies.push(customRule);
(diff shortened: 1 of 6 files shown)


webpro:
Thanks for the PR! Looks like the plugin creation script wasn't used. Please see at least both of [./AGENTS.md](https://github.com/webpro-nl/knip/blob/main/AGENTS.md) and [.agents/PLUGINS.md](https://github.com/webpro-nl/knip/blob/main/.agents/PLUGINS.md).

Zamiell:
I intentionally did not create a new plugin though so that we could keep all of the `markdownlint` stuff together. To be clear, do you want me to create a new plugin? It seems like potentially bad design, because updates to markdownlint would force changes to two separate plugin files.

packages/knip/src/plugins/markdownlint/index.ts
```diff
@@ -5,25 +5,41 @@ import { getArgumentValues } from './helpers.ts';
...
+  for (const modules of [config?.markdownItPlugins, config?.outputFormatters]) {
+    for (const module of modules ?? []) {
+      if (Array.isArray(module) && typeof module[0] === 'string') dependencies.push(module[0]);
+    }
+  }
   const scripts = manifest?.scripts
     ? Object.values(manifest.scripts).filter((script): script is string => typeof script === 'string')
     : [];
   const uses = scripts
     .filter(script => script.includes('markdownlint '))
```
webpro: Do we need to account for `markdownlint2` here as well?

webpro (commented):
Sorry I saw only the changed files and didn't catch you merged it into the existing one. One nit I have.

packages/knip/src/plugins/markdownlint/index.ts
Zamiell: https://github.com/Zamiell/knip/commit/d95d78f04d769de1607ba84bc60209ef4d30ef82

packages/knip/test/plugins/markdownlint-cli2.test.ts
```diff
@@ -0,0 +1,21 @@
...
+test('Find dependencies with the markdownlint-cli2 plugin', async () => {
+  const options = await createOptions({ cwd });
+  const { issues, counters } = await main(options);
+
+  assert(issues.binaries['package.json']['markdownlint-cli2']);
+
+  assert.deepEqual(counters, {
+    ...baseCounters,
+    binaries: 1,
+    devDependencies: 1,
```
webpro, line 19: FYI, the unused binary/devDep are caused by missing `node_modules` dir for knip to connect the dots ([hint](https://github.com/webpro-nl/knip/blob/02ca17a72c3f0afa7df32297d2dd2f5463ac6e00/packages/knip/test/plugins/_template.test.ts#L15)).

webpro:
Thanks for the fix! Any chance you could look into the failed ecosystem test?

Zamiell:
> Knip treats that as a direct  markdownlint  dependency, but it’s a built-in preset provided transitively by  markdownlint-cli2 . Hence the false-positive:
> Unlisted dependencies (1)
> .markdownlint.yml: markdownlint
> The resolver should ignore built-in  markdownlint/style/*  presets while still reporting external  extends  packages.

https://github.com/Zamiell/knip/commit/603a5c1f8df91372793f6e79acda8a01cf315785

webpro:
Thank you, James!

We can ignore the remaining failed test (another repo using Knip fixed something).

webpro:
:rocket: _This pull request is included in v6.28.0. See [Release 6.28.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.28.0) for release notes._

PR #1892, opened 2026-07-18
fix: keep gitignore negations scoped so sibling paths stay ignored

### LINKED ISSUE #1891: 🐛 Unrelated .gitignore negation un-ignores a sibling path globally
### Prerequisites

- [x] I'm using the latest version (6.27.0)

### Reproduction url

https://github.com/grahamtt/knip-gitignore-negation-repro

### Reproduction access

### Description of the issue

**Summary:** A `.gitignore` negation (`!…`) whose **basename** matches an ignored path un-ignores that path **globally** in knip, even when the negation targets a different directory. git scopes the negation correctly; knip does not, so knip analyzes files inside a git-ignored directory and reports them (e.g. as unused files).

**Reproduce** (`npm install && npx knip` in the linked repo):

```
Unused files (1)
build/reports/orphan.ts
```

`build/reports/orphan.ts` lives inside `build/`, which `.gitignore` ignores, so knip should not analyze it.

**The trigger is one unrelated line.** The `.gitignore` is:

```
node_modules/
build/
!foo/build/      <-- remove this line and the false positive disappears
```

`!foo/build/` un-ignores `foo/build/` — a non-existent, unrelated path. git still ignores the file either way:

```sh
git check-ignore build/reports/orphan.ts
# -> build/reports/orphan.ts   (git ignores it)
```

Delete the `!foo/build/` line, rerun `npx knip`, and there are no unused files.

**Scope of the trigger** (from bisecting):

- Leaks only when the negation's **basename** equals an ignored path's basename. `!foo/build/` leaks `build/`; `!completely/unrelated/path/` and `!keep.txt` do not.
- Independent of the negation's directory prefix or trailing-slash form: `!foo/build/`, `!x/build`, and ignoring as `build/**` all leak identically.
- `git check-ignore` disagrees with knip in every leaking case.

**Where it does *not* appear to be:** the pattern converter. Feeding the same `.gitignore` through the exported `parseAndConvertGitignorePatterns` + `expandIgnorePatterns`, composed into `picomatch(ignores, { ignore: unignores })`, returns `isIgnored("build/reports/orphan.ts") === true` **with and without** the negation. So the flat pattern set looks correct. The divergence seems to arise in the directory-walk / pruning in `src/util/glob-core.js`, where an unignore with a colliding basename appears to stop `build/` from being pruned as a directory (after which the descendant is walked and reported). That last part is a hypothesis; the black-box reproduction above is deterministic.

**Expected:** knip honors `.gitignore` with the same scoping git uses, so an unrelated negation does not cause files under an ignored directory to be analyzed.

### Versions

- knip 6.27.0
- node 24.16.0

xianjianlf2:
Closes #1891.

**Problem:** An unrelated `.gitignore` negation such as `!foo/build/` globally un-ignores sibling `build/` paths — files under e.g. `tools/build/` that should stay gitignored start showing up in results.

**Root cause:** glob-core drops ignore patterns that are "shadowed" by any unignore pattern. That drop is needed because tinyglobby cannot express unignores, and traversal must not prune directories that contain unignored paths (the yarn-berry `.gitignore` case depends on exactly this). But the glob results were never re-checked afterwards, so every path matched by the dropped pattern leaked through — not just the negated one.

**Fix:** Keeps the drop for traversal, and post-filters `glob()` results through the exact picomatch matcher that `getGitIgnoredHandler` already builds. The filter is gated on `unignores.size > 0`, so the common case (no negations) has no extra work.

**Testing:** New fixture `gitignore-scoped-negation` plus a regression test that was red before the fix and green after, under both bun and node. 16/16 related regression tests pass, full smoke suite 870/870, `tsgo`/oxlint/oxfmt clean.

### DIFF
--- a/packages/knip/fixtures/infra/gitignore-scoped-negation/.gitignore
+++ b/packages/knip/fixtures/infra/gitignore-scoped-negation/.gitignore
@@ -1,0 +1,2 @@
+build/
+!tools/build/
--- a/packages/knip/fixtures/infra/gitignore-scoped-negation/build/reports/orphan.ts
+++ b/packages/knip/fixtures/infra/gitignore-scoped-negation/build/reports/orphan.ts
@@ -1,0 +1,1 @@
+export const orphan = 1;
--- a/packages/knip/fixtures/infra/gitignore-scoped-negation/index.ts
+++ b/packages/knip/fixtures/infra/gitignore-scoped-negation/index.ts
@@ -1,0 +1,3 @@
+import { codegen } from './tools/build/codegen.ts';
+
+codegen();
--- a/packages/knip/fixtures/infra/gitignore-scoped-negation/package.json
+++ b/packages/knip/fixtures/infra/gitignore-scoped-negation/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "@fixtures/gitignore-scoped-negation"
+}
--- a/packages/knip/fixtures/infra/gitignore-scoped-negation/tools/build/codegen.ts
+++ b/packages/knip/fixtures/infra/gitignore-scoped-negation/tools/build/codegen.ts
@@ -1,0 +1,1 @@
+export const codegen = () => {};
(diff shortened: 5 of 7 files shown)


webpro:
Thanks for the pull request. Much appreciated but I have an unpushed fix that I prefer. Hope you don't mind.

Note to self: push any finished work.

webpro:
Closing this with 5c026ba (and then a5fa3b0 + 404a53e)

PR #1893, opened 2026-07-18
feat: support preprocessors in configuration

### LINKED ISSUE #1491: 💡 Preprocessor for VSCode Knip
### Suggest an idea for Knip

The VSCode Knip extension is reporting issues that I've excluded via a [preprocessor](https://knip.dev/features/reporters#preprocessors).
So running `knip --preprocessor ./knip-processor.ts` results in no issue, but the VSCode extension does.

Would it be possible to have the `preprocessor` be part of the config (preferably).
Or have a new option in VSCode extension settings that allows me to set a preprocessor path for my workspace.

Thanks!

morgan-coded:
`--preprocessor` only affected CLI reporting, so the editor path could return issues that the same preprocessor removed from CLI output. This adds root `preprocessor` and `preprocessorOptions` settings and runs one shared result stage for non-watch CLI output and session results, including refreshes. CLI values replace matching config values; config specifiers resolve from the file that supplied the setting, while CLI resolution stays unchanged. Sessions transform a copy and fail open with the error surfaced, while CLI errors stay loud; CLI watch is untouched, and preprocessor source edits require an editor or extension host restart or a configured path change.

Fixes #1491


webpro:
Not sure where to start here. I'm OK with AI PRs, but this is not convenient to review. Too many changes and redundancy in docs. In JSON config `preprocessorOptions` should be an object not a string. Excessive tests/fixtures. Don't overdo it.

morgan-coded:
Slimmed it down — the diff is now 45 files, +408/−56 (was 102, +795/−59). `preprocessorOptions` is an object in JSON config; it's stringified at the existing CLI/config merge point, so the `--preprocessor-options` flag and the reporter-options convention stay unchanged and preprocessors keep a single code path. Docs state each fact once in configuration.md, with cli.md and the features table reverted. Tests went from 23 cases to 14 and fixtures from 89 files to 34, collapsed onto the flat per-config fixture convention — kept one test per contract point, the Windows path regression, and the Bun skips. Happy to cut further if anything still reads as too much.

webpro:
Going to stop right here. I don't want to play 🏓 with some model until it doesn't feel like slop anymore. My reasoning is that still, looking at e.g. packages/knip/src/session/session.ts, packages/knip/src/util/create-options.ts and packages/knip/src/util/reporter.ts, we seem to be adding way too much logic just to execute some user-defined function. Which we were already doing in the CLI flow.

Issue #698, opened 2024-06-25
False negative when re-exporting


neelmraman:
Hi, thanks for fixing the first false negative case mentioned in https://github.com/webpro-nl/knip/issues/695.

[...]

PR #1895, opened 2026-07-20
Report unused re-exports when ignoreExportsUsedInFile is set

### LINKED ISSUE #1894: 🐛 --fix strips source export but leaves re-export when ignoreExportsUsedInFile is set
### Prerequisites

### Reproduction url

https://github.com/mlm20/knip-fix-dangling-reexport

### Reproduction access

### Description of the issue

With `ignoreExportsUsedInFile` enabled, `--fix` removes the `export` keyword at the declaration site but leaves the matching re-export in place. The result does not compile.

Running `knip --fix` on the reproduction:

```
Unused exported types (1)
Dead  type  schemas.ts:2:13  (removed)
```

`schemas.ts` after the fix:

```ts
export type Used = { a: number };
type Dead = { b: number };
```

`barrel.ts` is untouched:

```ts
export type { Used, Dead } from "./schemas";
```

```
barrel.ts(1,21): error TS2459: Module './schemas' declares 'Dead' locally, but it is not exported.
```

Removing `ignoreExportsUsedInFile` from `knip.json` avoids it. Knip then reports both sites and fixes both, giving `export type { Used,  } from "./schemas";`, which compiles.

Expected: `--fix` leaves the code in a working state in both cases, as documented in the auto-fix guide.

The same thing happens when only some names in a re-export statement are unused.

knip 6.27.0, TypeScript 5.9.3, Node 24.3.0.

mlm20:
Fixes #1894

With `ignoreExportsUsedInFile` set, re-exports were always treated as used in their own file and never reported. A re-export like `export type { Dead } from './schemas'` has no local binding, so it cannot actually be used in its own file. When the source export was unused, `--fix` removed the `export` keyword at the declaration but left the re-export behind, and the project no longer compiled.

This removes the blanket `isReExport` check, so re-exports get the same local reference check as other exports. The `import { X }` plus `export { X }` case keeps working, since usage of `X` in the file is picked up by that check (covered by the existing `re-export` fixture).

The expected count in `test/re-exports/ignore-exports-used-in-file.test.ts` goes from 1 to 2: the unused export is now also reported at the re-export site, same as without `ignoreExportsUsedInFile`. That test was added for #698 to make sure the export is reported at all, so this stays in line with its intent.

### DIFF
--- a/.github/workflows/snapshots/sanity.txt
+++ b/.github/workflows/snapshots/sanity.txt
@@ -1,0 +1,9 @@
+Unused exported types (8)
+packages/@repo/debug-proxy/src/createDebugProxy.ts: ProxyHeaders, SSEEvent
+packages/sanity/src/core/canvas/i18n/index.ts: CanvasLocaleResourceKeys
+packages/sanity/src/core/comments/i18n/index.ts: CommentsLocaleResourceKeys
+packages/sanity/src/core/form/inputs/PortableText/contexts/fullscreen/index.ts: FullscreenPTEContextValue
+packages/sanity/src/core/releases/i18n/index.ts: ReleasesLocaleResourceKeys
+packages/sanity/src/core/singleDocRelease/i18n/index.ts: SingleDocReleaseLocaleResourceKeys
+packages/sanity/src/core/tasks/i18n/index.ts: TasksLocaleResourceKeys
+packages/sanity/src/core/validation/index.ts: ValidationContext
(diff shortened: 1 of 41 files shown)


webpro:
The `item.isReExport` that's removed here was some kind of safety net. Removing it opens the door for certain false positives/negatives. So I'm trying to fix that up in my latest commit, can you please confirm it still behaves as expected?

mlm20:
Yep I can confirm that it work. Ran my original repro against d648f41, both Dead sites get reported now and `--fix` cleans up both. Thanks for tightening this!

webpro:
Thanks! These are real findings now: https://github.com/webpro-nl/knip/actions/runs/29754772364/job/88631556813?pr=1895

webpro:
:rocket: _This pull request is included in v6.28.0. See [Release 6.28.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.28.0) for release notes._

PR #1900, opened 2026-07-21
fix(glob): do not follow symbolic links in async glob

mogelbrod:
## Problem

In a large pnpm monorepo, one app workspace's `knip` run took ~106s cold. `--performance` attributed ~99% of the time to `defaultGlob`, and a CPU profile showed it dominated by `node:path` `normalizeString`/`resolve`/`relative` (~34%) plus `picomatch` (~12%) — i.e. filtering an enormous path set, not parsing or analysis.

## Root cause

The async `glob()` in `src/util/glob-core.ts` builds its tinyglobby options without `followSymbolicLinks`, so it inherits the default (`true`). The sync `glob()` in `src/util/glob.ts` already passes `followSymbolicLinks: false`.

With symlink-following on, relative patterns that reach outside `cwd` — such as the sibling-package workspace patterns knip generates for a monorepo app (`../../packages/*`, `../../packages/*/*`) — let the crawler descend through the symlinked `node_modules` into the pnpm store and walk the whole dependency tree. The `**/node_modules` ignore doesn't reliably prune this across `../` traversal.

Minimal repro from an app workspace, calling `tinyglobby` directly:

```js
await glob(['**/*'], { followSymbolicLinks: true })   // → ~133,000 paths
await glob(['**/*'], { followSymbolicLinks: false })  // → ~800 paths
```

## Fix

Set `followSymbolicLinks: false` in the async `glob()`, matching the sync variant. Knip already resolves workspace packages to their real paths (`realpathSync`) before globbing, so it never needs to follow symlinks to find source.

## Impact

In the affected repo, that workspace's `knip` run dropped from **~106s → ~7s** with **identical findings** (verified by diffing output before/after). Other workspaces were unaffected or slightly faster. `pnpm run ci` (oxlint + oxfmt --check) and the `tsgo` build pass locally.

Happy to add a fixture-based test if you'd like — a symlink fixture is a little awkward to commit, so I left it out of this draft.


webpro:
Thanks for the PR! I do like the idea and performance win. I'm a bit wary of the cases we're not capturing here, so I'm going to let it simmer for a bit trying not to miss blind spots. E.g. symlinked sources outside `node_modules`.

> Knip already resolves workspace packages to their real paths (`realpathSync`) before globbing, so it never needs to follow symlinks to find source.

That isn't true, `realpathSync` is applied only to individual files already enumerated by `tsc` and to installed dependency package roots (not to e.g. workspace roots).

> Happy to add a fixture-based test if you'd like — a symlink fixture is a little awkward to commit, so I left it out of this draft.

Why is it awkward? We did it before, please do.

Please rebase on top of main for green lights in CI.

mogelbrod:
> Thanks for the PR! I do like the idea and performance win. I'm a bit wary of the cases we're not capturing here, so I'm going to let it simmer for a bit trying not to miss blind spots. E.g. symlinked sources outside `node_modules`.

I asked Claude Fable to thoroughly review the possible unhandled cases, and got the following response:

<details>
<summary>Findings: case-by-case analysis of <code>followSymbolicLinks: false</code></summary>

> The breaking condition is precise: tinyglobby drops a path only when the symlink sits **strictly below the pattern's static base**, or **is the named file itself**. The crawl root (pattern base) is resolved by `readdir` regardless of the flag.

| Case | Before (`true`) | After (`false`) | Verdict |
|---|---|---|---|
| Symlinked file/dir below base, target inside project | duplicate lexical paths → false-positive unused files | enumerated once, at real path | improved |
| …target outside project, imported | false-positive unused files for every imported file | analyzed via module resolution, not reported | improved |
| …target outside project, never imported | reported unused (true positive) | silently missed | regression (narrow) |
| Entry/plugin-config file is itself a symlink (literal pattern) | analyzed at lexical path | dropped entirely | regression |
| Literal file behind a symlinked dir (`linkdir/file.ts`) | found | found | unchanged |
| Magic pattern with symlinked base (`linkdir/**/*.ts`) | found | found | unchanged |
| Symlinked workspace root | works (pre-existing false-positive quirk) | byte-identical output | unchanged |
| Dangling symlink | skipped | skipped | unchanged |
| `_dirGlob` (`onlyDirectories`) | symlinked dirs never returned | same | unchanged |
| `mapWorkspaces`, sync glob, fdir gitignore walk, `existsSync` script-entry checks | — | not touched by this PR | unchanged |

> **Symlinked sources outside `node_modules`**: following symlinks never produced correct results here. The module graph is realpath-based (TS resolves symlinks with `preserveSymlinks: false`, plus the `tryRealpath` calls), while the glob enumerates lexical paths — so every *imported* file enumerated through a symlink was reported as a false-positive unused file. With the change, targets inside the project tree are enumerated once at their real path (correct); targets outside the tree are still resolved and analyzed when imported, just no longer candidates for unused-files/exports reporting. The narrow loss: a never-imported file reachable only through a symlink is no longer reported unused — but before, that report came bundled with false positives for all its used siblings.
> 
> **Symlinked entry or plugin config files** (literal patterns like `index.ts` from `main`/`entry`, or config lookups in `WorkspaceWorker`): `glob(['link.ts'], { followSymbolicLinks: false })` returns `[]`, so these are now dropped. Verified with full runs on `index.ts -> impl.ts`: before, the entry was analyzed with one false positive (`impl.ts` reported unused); after, the entry is invisible and everything it references is reported unused (a `Refine entry pattern (no matches)` hint does fire). Nuance: before this PR this only "worked" for same-directory targets — a target in another directory had its relative imports resolved from the symlink's lexical directory, producing unresolved imports and false positives.
> 
> A possible follow-up that closes the entry/config gap while keeping the perf win: split out magic-less patterns (`picomatch.scan`) in `glob-core.ts` and check them with a direct `stat` (which follows symlinks) instead of sending them through the crawler — literal patterns then skip the crawl entirely.

---

</details>

What do you think about the "Entry/plugin-config file is itself a symlink" regression? Worth tackling in this PR?

> That isn't true, `realpathSync` is applied only to individual files already enumerated by `tsc` and to installed dependency package roots (not to e.g. workspace roots).

My bad - I've corrected the PR description. The module graph ends up realpath-based (TS module resolution resolves symlinks with `preserveSymlinks: false`, + the `tryRealpath` calls above), while the glob enumerates lexical paths.

> Why is it awkward? We did it before, please do.

Apologies, I had missed the `circular-symlinks` test. Added `fixtures/infra/glob-symlinks` with committed symlinks, using `circular-symlinks` as a template. Without the fix it fails with 3 false positive unused files.

The test is skipped in environments where the symlinks don't materialize (Windows without `core.symlinks`), since the bun workflow runs tests on `windows-latest`.

webpro:
Thanks, Victor! Let's roll with it 🛼

webpro:
:rocket: _This pull request is included in v6.29.0. See [Release 6.29.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.29.0) for release notes._

PR #1747, opened 2026-05-16
Detect dynamic imports in Svelte compiler

sjh9714:
## Summary

Improve the built-in Svelte compiler so it extracts dynamic imports from `<script>` blocks.

## What changed

- Added dynamic `import('...')` extraction for Svelte script blocks.
- Extended the Svelte compiler regression fixture with a type import and a dynamic component import.
- Kept import-like text in comments ignored by the existing comment stripping.

## Why

In #1670, the built-in Svelte compiler was suggested as the right place to preserve type imports while also detecting dynamic component imports. The existing compiler already handled static and type imports, but dynamic imports inside Svelte scripts were missed and could leave dynamically imported components reported as unused files.

## Testing

- `node --test test/svelte-script-jsdoc-import.test.ts`
- `node --test test/plugins/svelte.test.ts`
- `node --test test/plugins/sveltekit.test.ts`
- `pnpm run ci`
- `pnpm --dir packages/knip build`
- `pnpm --dir packages/knip test --runtime node --smoke`

Refs #1670

Issue #1670, opened 2026-04-08
🐛 Type imports in Svelte files are not detected when using svelte compiler


arnolicious:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/arnolicious/knip-compiler-types-repro

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

When using knip with the "real" svelte compiler, knip fails to detect types that are imported in svelte files.

When _not_ using the "real" svelte compiler, knip detects the types just fine, but then fails to detect svelte specific features, such as dynamic component imports.

I assume the solution is to somehow configure / patch the compile step to also include types, however I wouldn't really know where to start.

Unfortunately in this state, this makes it very difficult to use knip in a large sveltekit project

webpro:
The basic idea is to set `compilers: { svelte: input => output } )` where `input` is the Svelte source file string, and `output` is the TypeScript source file string containing the imports and exports. Perhaps an approach could be to copy functions/regexes from Knip source code and adjust to your needs, e.g. also include dynamic component imports.

Happy to help and/or review a pull request with improved svelte "compiler" in Knip.

webpro:
Also see https://knip.dev/features/compilers#svelte

arnolicious:
Sorry, just to make sure I understand correctly, you'd suggest adjusting the built-in knip svelte compiler to also include things like dynamic imports, rather than using the real svelte compiler like described in https://knip.dev/features/compilers#svelte ?

Or would there be a way to extend the compile step, to use the real compiler, and additionally handle keeping the type imports/exports?

I'm not very familiar with how knip works under the hood, so sorry for misunderstanding how things might work

webpro:
> Sorry, just to make sure I understand correctly, you'd suggest adjusting the built-in knip svelte compiler to also include things like dynamic imports [..]

Yes.

By "copy" I meant you could suit yourself by configuring your own compiler. If you set `compilers.svelte` then that function overrides the built-in one. Feel free to base that on Knip source code.

[...]

PR #1908, opened 2026-07-26
fix: detect dynamic imports in Svelte component templates

### LINKED ISSUE #1907: 🐛 Built-in Svelte compiler misses dynamic imports in template markup (false "unused files")
### Prerequisites

### Reproduction url

https://github.com/firatciftci/knip-svelte-template-import-repro

### Reproduction access

### Description of the issue

A `.svelte` component imported only from template markup via a dynamic `import()` outside any `<script>` block is reported as an unused file. In the reproduction, `src/Lazy.svelte` is used here in `src/App.svelte`:

```svelte
{#await import('./Lazy.svelte') then { default: Lazy }}
  <Lazy />
{/await}
```

Yet, `npx knip` reports:

```
Unused files (1)
src/Lazy.svelte
```

On a real app, one `{#await import(...)}` at a route root makes the lazily-loaded component and everything it transitively imports look unused.

### Expected output

`src/Lazy.svelte` is used, so it should not be reported.

### Notes

The built-in Svelte compiler extracts imports via `importsWithinScripts`, which only scans `<script>` blocks. Dynamic imports written directly in template markup are never seen.

This is the template-markup counterpart to #1747 (associated bug report is #1670). That PR taught the built-in compiler to detect dynamic imports written inside `<script>` and to keep type-only imports, but imports in the template is a gap that still remains. The documented `compilers.svelte = (s) => compile(s).js.code` override does pick up template dynamics, but it strips type-only imports (and per the docs keeps component-prop `export`s, reporting them as unused).

Proposed fix is to extend the built-in compiler to also collect dynamic `import()` expressions from template markup outside `<script>`.

### Environment

knip 6.29.0, Svelte 5/SvelteKit 2.

firatciftci:
Closes #1907.

### Problem

#1747 (associated bug report is #1670) taught the built-in Svelte compiler to detect dynamic imports inside `<script>` while preserving type-only imports. Dynamic imports written directly in template markup are still missed, because `importsWithinScripts` only scans `<script>` blocks:

```svelte
{#await import('./Modal.svelte') then { default: Modal }}
  <Modal />
{/await}
```

Components reachable only through a lazy template import and everything they transitively import are therefore reported as unused files.

### Fix

Keep scanning `<script>` and additionally collect dynamic `import()` expressions from the template. We strip `<script>` blocks first, so script-level imports found by `importsWithinScripts` are not duplicated, and normalize each hit to `import("<path>")`. As discussed in #1670, this stays consistent with the direction of #1747 and Knip's design goal for these compilers and avoids the documented downsides of the real `svelte/compiler`.

### Test

New fixture `fixtures/plugins/svelte-template-import` with a case in `test/plugins/svelte.test.ts` asserting `Lazy.svelte` is not reported unused. This sits alongside the existing built-in-compiler svelte test.


webpro:
Thanks! I like where this is going, but seems to cover only the happy path. Please add more cases and error-prone syntax to the fixture to harden the "compiler" (i.e. regex).

firatciftci:
@webpro Thanks! Pushed a follow-up. Beyond the happy path the fixture now covers double quotes, whitespace/multiline, event-handler imports, import attributes, and multiple-per-line, plus a static and a type-only import. The matcher ignores trailing import attributes and no longer trips on `foo.import(`/`reimport(`, and `<style>`/HTML comments are stripped so commented-out imports do not count.

webpro:
Since this expands the built-in compiler into Svelte template expressions, I extended its regex coverage beyond the initial happy path while keeping it intentionally lightweight.

The updated version only considers bounded Svelte expression regions, ignores basic strings and comments, supports quoted and static template-literal specifiers as well as comments around `import(...)`, and preserves the original import expression. I also moved the Svelte-specific logic into `compilers/svelte.ts` and added fixtures for these cases.

This remains a heuristic built-in compiler, not a Svelte or JavaScript parser. Projects using more complex syntax can override it with `svelte/compiler`, as described in the compiler documentation.

webpro:
Thanks Firat!

webpro:
:rocket: _This pull request is included in v6.30.0. See [Release 6.30.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.30.0) for release notes._

Issue #248, opened 2023-09-18
Ensure that `@types/*` used in emitted declarations are in `dependencies`


alecmev:
I'm not sure if this is in scope of Knip, but it feels like it could be. In a nutshell, if I have a module like this:

```ts
// bar/src/index.ts
import type { Foo } from "foo";
export const bar = (): Foo => ...
```

And `foo` comes with a corresponding `@types/foo`, then I need to make sure that it's actually in `dependencies`. Otherwise `bar/dist/index.d.ts` will be broken in any consumer of `bar` that doesn't have `@types/foo` in its dependency tree.

The complication I'm seeing is determining whether a `@types/*` actually ends up being used in `dist`. And putting all `@types/*` in `dependencies` doesn't make sense, obviously. Does TypeScript have an API of any sort for checking this?

More context in [this Stack Overflow question](https://stackoverflow.com/q/45176661/242684).

webpro:
Yes, I think Knip can help here. Related to #241.

webpro:
Currently Knip just expects `@types/*` packages to be in `devDependencies` (with `--production --strict`).

So yeah. the issue is that `@types/pkg` sometimes need to be in `dependencies`.

More refs:

- Ask in TS Discord: https://discord.com/channels/508357248330760243/640177429775777792/1156692609763389562
- https://github.com/microsoft/types-publisher/issues/81

To be continued

webpro:
Any chance you could whip up a repro? One or more cases would be very helpful. Running in strict production mode should "fail" now.

alecmev:
Spent half an hour trying to understand why I can't get my repro to fail, when I realized that Knip actually never complained :joy: Knip is fine with `@types/*` being anywhere, or missing. [This repro](https://github.com/alecmev/knip-repro-248) currently passes, but needs to fail.

alecmev:
Updated the repro to also include a case without `@types/*` (it shouldn't report `@microsoft/tsdoc` as unused).

danvk:
@alecmev you can often take advantage of duck typing to sever `@types` dependencies, i.e. define an `interface` with just your needs in your public API that's compatible with the other package's type. Just doing my part to fight dependency bloat :)

If you know examples of popular npm packages that bundle TS types and put `@types` in prod dependencies, I'd be interested to hear about them. Packages seem to be all over the place in how they handle the issue of transitive `@types` deps.

webpro:
[...]

PR #1912, opened 2026-07-27
Analyze published type dependencies (resolve #1909)

### LINKED ISSUE #1909: 💡 Option to treat type-only imports in production code as production dependencies
### Suggest an idea for Knip

**Problem**

Strict mode reports type-only imports listed in `dependencies`, expecting them in `devDependencies`. That is a sensible default for applications, but for published libraries, it conflicts with a placement that is sometimes required: when a type-only import ends up referenced by the emitted `.d.ts` files, the package must stay in `dependencies`, or the declarations break in consumer projects that do not hoist it - we learned this from real consumer breakage under Yarn PnP.

In other words, for a library, the correct placement rule is the opposite of what strict mode assumes today: a package imported (even type-only) by *production* code is a *production* dependency, because its types become part of the public API surface.

**Prior art**

Issue #248 asked to analyze the emitted `.d.ts` files and verify that `@types/*` packages referenced there are listed in `dependencies`. It was closed because analyzing emitted output is out of scope, and this limitation is now documented on the known-issues page. This proposal stays within that boundary: it needs no analysis of emitted output at all. Everything required is already visible to Knip in the source - which imports are type-only and whether they occur in production files. The only change is in how strict mode classifies those imports.

**Current workaround and why it falls short**

Production-scoped ignores work:

```ts
ignoreDependencies: [ '@types/hast!', 'hast-util-to-mdast!' ]
```

but every new type-only production dependency in any workspace needs a new manual entry, the ignore also disables the unused detection for that package in production mode, and the config expresses "ignore this" rather than "our project keeps public-API types in `dependencies`".

**Proposal**

An opt-in option — global or per workspace — under which strict mode treats type-only imports in production files as production dependencies. For example:

```json
{ "typesInDependencies": true }
```

Concretely, given a production source file with:

```ts
import type { Root } from 'hast';
```

- **Today (strict mode):** `@types/hast` in `dependencies` is reported, because type-only imports are expected in `devDependencies`.
- **With the option enabled:** `@types/hast` in `dependencies` satisfies the import and nothing is reported. Unused detection is unaffected — if the import is removed, the package is still reported as unused.
- **Optionally**, the rule could go further and report the inverse misplacement: a type-only import in a production file whose package is only in `devDependencies`, since that placement breaks consumers of the published library. This would give library authors the enforcement #248 asked for, without analyzing emitted declarations.

Type-only imports that appear only in test/dev files would keep the current behavior and still belong in `devDependencies`.

webpro:
Resolves #1909. Implements the emitted-declaration analysis originally requested in #248.

Type-only imports that end up in a package's emitted declaration files are part of its public API. Consumers resolve them from their own dependency tree, so the referenced packages must be in `dependencies`, while strict mode used to expect all type-only imports in `devDependencies`.

Knip now analyzes the published type surface of non-private workspaces. Entry declarations are taken from `types`/`typings`, the declaration next to `main`, the `index.d.ts` fallback, `exports` (with `types` condition precedence, wildcard patterns and `null` subpath exclusions) and `typesVersions`. The declaration graph is walked from there, including triple-slash references and relative and self-name imports.

External packages referenced by this graph count as production usage:

- In strict mode they are reported at their declaration file location when only in `devDependencies` or missing.
- Untyped packages map to their installed or declared `@types/*` counterpart, so the report names the package consumers actually need.
- Resolution follows the consumer view: `.ts` and `.js` specifiers in emitted declarations resolve to the matching `.d.ts` via extension aliases, and the author's tsconfig `paths` do not apply (the source graph still uses them).

If declaration output has not been built, Knip does not guess from source imports. The walk is bounded to declarations reachable from the published entry points; on this repository it adds about 35ms.


lishaduck:
Hey, I was planning on giving this a shot, someday! What's up with everybody taking my issues? (jk)

Whoooooowhooooo! Thanks Lars! This is🔥 🔥  You've made my day and I just woke up! (Ok I should go so I can get a chance to use it but I just had to show some appreciation here.)

webpro:
Thanks Eli. That is all much appreciated, and it means a lot coming from a true power user!

webpro:
:rocket: _This pull request is included in v6.30.0. See [Release 6.30.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.30.0) for release notes._

PR #1914, opened 2026-07-29
feat: Add basic support for Marko 6 and 5

caseycarroll:
Adds a marko plugin to knip to analyze dependencies in `.marko` files. 

Marko has an [auto tag discovery feature](https://markojs.com/docs/reference/custom-tag) which allows authors to use custom tags without explicitly importing them. To support this, this plugin will look for Marko in the dependencies then look for `marko.json` files for each project dependency in node_modules. It will then find tag names in the locations specified in marko.json and check that at least one of the tag names is used in the project. 

The `marko.json` parsing is intentionally limited. It only looks at `exports` and `tags-dir` because other mechanisms like `taglib-imports` will be deprecated. If a Marko dependency exposes tags with a deprecated property, the plugin will categorize the dependency as used without searching for usages in the project code. 

This plugin also naively covers marko's "concise syntax" since it's not widely used. A follow up PR could add stronger concise syntax support, if the community needs it. https://markojs.com/docs/reference/concise-syntax 

## Additional Changes
- Appears the OpenClaw plugin was not accounted for in the docs list, so it's been added to `plugins.md` as well as marko. 

## AI Disclosure
I authored the PR description in my own words and reviewed the code manually, but did use a model to implement the plugin and fixtures.


packages/docs/src/content/docs/reference/plugins.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Plugins (177)
+title: Plugins (179)
```
webpro: Please undo changes in this file, we can auto-generate separately (and they're always re-generated for website deploy).

packages/knip/src/plugins/marko/compiler.ts
```diff
@@ -0,0 +1,89 @@
...
+const commentMatcher = /<!--[\s\S]*?-->|\/\*[\s\S]*?\*\/|^[ \t]*\/\/.*$/gm;
+const importMatcher =
+  /^[ \t]*(?:(?:server|client)[ \t]+)?(import\s+(?:(?:type\s+)?[^"'`;]+?\s+from\s+)?(["'])([^"'\r\n]+)\2)[ \t]*;?/gm;
+const dynamicImportMatcher =
+  /^[ \t]*(?:(?:const|let|var)[ \t]+[$\w]+[ \t]*=[ \t]*|(?:await[ \t]+)?)(import\([ \t]*(["'])([^"'\r\n]+)\2[ \t]*(?:,[^)\r\n]*)?\))[ \t]*;?/gm;
+const exportMatcher =
+  /^[ \t]*(export\s+(?:type\s+)?(?:\*(?:\s+as\s+[$\w]+)?|\{[^}]*\})\s+from\s+(["'])([^"'\r\n]+)\2)[ \t]*;?/gm;
+const conciseStyleMatcher = /^[ \t]*style(?:\.([\w.-]+))?[^\n{]*\{([\s\S]*?)^[ \t]*\}/gm;
+const htmlTagMatcher = /<([a-z][\w.-]*)(?=[\s/|>])/g;
+const conciseTagMatcher = /^[ \t]*([a-z][\w.-]*)(?=[ \t/|]|--|$)/gm;
```
webpro, line 10: The import/export/dynamic-import matchers reimplement what `oxc-parser` does out of the box. How about this: instead of reconstructing statements with regexes, the compiler could reduce the source to parseable JS (and let the parser do its job):

1. Strip comments and style blocks (still compiling styles separately as it does now), lines starting with `<`, and `import ... from "<tag>"` lines.
2. Neutralize the Marko-only syntax so a hard error can't land before the imports: `server`/`client`/`static`/`$` prefixes, the anonymous `class {` (give it a name), and the `export` modifier on local declarations (`export interface Input` must not surface as an unused export).
3. Return the synthetic imports (marko, taglib dependencies, compiled styles) followed by the remainder verbatim. Synthetic lines go first: a parse error in markup leftovers truncates the module record from that point on, so the leftovers must come last.

No parsing in the compiler: the pipeline picks up static, dynamic and type imports and re-exports in any shape, quoted attribute values are inert to the tokenizer, and dynamic imports inside the class block get picked up too.

To me this feels sturdier and easier to maintain. _Even though_ Marko syntax might change, using regexes for what `oxc-parser` does better feels a bit cumbersome.

WDYT?

webpro (commented):
Thanks for the pull request! Would be cool to add Marko plugin. I do have a note.

packages/knip/src/plugins/marko/compiler.ts
caseycarroll: Agreed this feels cumbersome! I like your suggested approach. Will tinker with that and follow up soon.

packages/docs/src/content/docs/reference/plugins.md
caseycarroll: Great! Removed.

packages/knip/src/plugins/marko/compiler.ts
caseycarroll: Okay, refactored to have a tag dependency pass, a style pass, then a stage where I sanitize the `.marko` template so parseable JS is fed into oxc-parser. 

I've added regex's to account for different marko syntax things like `class {}`.

packages/knip/src/plugins/marko/index.ts
```diff
@@ -0,0 +1,93 @@
...
+const tagFilePatterns = [
+  '**/*.marko',
+  `**/{component,component-browser}.${scriptExtensions}`,
+  `**/*.{component,component-browser}.${scriptExtensions}`,
+  `**/style.${styleExtensions}`,
+  `**/*.style.${styleExtensions}`,
+];
+
+const tagProduction = tagFilePatterns.map(pattern => `**/{${tagDiscoveryDirs.join(',')}}/${pattern}`);
+export const production = tagProduction;
```
webpro:
```suggestion
const production = tagProduction;
```

It's in the default plugin export obj.

packages/knip/src/plugins/marko/index.ts
```diff
@@ -0,0 +1,93 @@
...
+explicit module imports, re-exports and style imports without loading the Marko compiler.
+
+For installed tag libraries, tag names are discovered from the directory configured by
+\`marko.json#exports\` or \`marko.json#tags-dir\` and matched against literal tags in project templates.
+Tag libraries without either field are conservatively considered used. \`taglib-imports\` is not resolved.
+
+Local tag templates and their conventionally associated component and style files in \`components\`
+(Marko 5) or \`tags\` (Marko 6) are conservatively treated as production entries. Knip does not
+report unused auto-discovered local tags inside those directories. Modern Marko 5
+\`marko-tag.json\` entry points and compiler hooks are also resolved.`;
```
webpro: Please remove `docs` and `note` entirely. They look like implementation details, not documenting surprising behavior with answers / workarounds users need to know about.

packages/knip/src/plugins/marko/compiler.ts
```diff
@@ -0,0 +1,103 @@
...
+      'import "marko";',
+      ...collectTagDependencies(source, tagDependencies, fallbackDependencies),
+      ...collectStyles(source, path),
+      sanitize(source),
+    ].join('\n');
+  };
+
+const compiler = createCompiler();
+
+export default compiler;
```
webpro:
```suggestion
```

webpro (commented):
Thanks for the updates! Almost good to go.

_Optional:_ a fixture with a `marko.json` that has neither `exports` nor `tags-dir` would pin the `taglib-imports` fallback ("conservatively considered used") behavior.

packages/knip/src/plugins/marko/index.ts
caseycarroll: resolved

packages/knip/src/plugins/marko/index.ts
caseycarroll: Removed the docs. I should have cross-validated with other plugins here, my bad!

packages/knip/src/plugins/marko/compiler.ts
caseycarroll: fixed

webpro:
:rocket: _This pull request is included in v6.31.0. See [Release 6.31.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.31.0) for release notes._

webpro:
Thanks again, Casey — nice one.

PR #1920, opened 2026-08-02
feat: support configuring preprocessors in knip.json

### LINKED ISSUE #1491: 💡 Preprocessor for VSCode Knip

bytedoe:
Resolves #1491

This PR bridges the gap between the Knip CLI and the VS Code extension by allowing preprocessors to be defined in `knip.json` config, ensuring they are respected by the Language Server.


packages/language-server/src/server.js
```diff
@@ -284,6 +297,35 @@ export class LanguageServer {
     this.start();
   }
 
+  /** @param {Issues} issues */
+  async #preprocessIssues(issues) {
+    if (this.preprocessor.length === 0) return issues;
+    const run = this.knip?.runPreprocessors ?? runPreprocessors;
+    /** @type {ReporterOptions} */
+    const data = {
```
webpro: Please do not re-create `ReporterOptions`, should be the actual object.

packages/language-server/src/server.js
```diff
@@ -258,14 +267,18 @@ export class LanguageServer {
       this.connection.console.log('Creating options');
       const options = await knip.createOptions({ cwd: this.cwd, isSession: true, args: { config: configFilePath } });
       this.rules = options.rules;
+      this.preprocessor = options.preprocessor ?? [];
+      this.preprocessorOptions = options.preprocessorOptions ?? {};
+      this.knip = knip;
```
webpro: Please avoid storing the dynamically loaded Knip module on the server instance just to recover `runPreprocessors` later.

packages/language-server/src/server.js
```diff
@@ -284,6 +297,35 @@ export class LanguageServer {
...
+      isDisableTagHints: true,
+      isTreatConfigHintsAsErrors: false,
+      isTreatTagHintsAsErrors: false,
+      options: '',
+      preprocessorOptions: this.preprocessorOptions,
+      selectedWorkspaces: undefined,
+      configFilePath: undefined,
+    };
+    const result = await run(this.preprocessor, data);
+    return result.issues;
```
webpro: This returns raw session results even though diagnostics are preprocessed separately. Applying preprocessing at the session result boundary would keep diagnostics, refreshes, and result requests consistent.

packages/knip/fixtures/config-preprocessor/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@fixtures/config-preprocessor",
+  "knip": {
+    "entry": ["preprocessor.js", "preprocessor-with-options.js"],
```
webpro: Nitpick: ideally we wouldn't need to add the processor files to `entry`. Blurs the intent/expectations. Might be a pre-existing issue, though.

webpro (commented):
Thanks for the pull request! It has potential, hopefully we can straighten the edges.

packages/language-server/src/server.js
bytedoe: Sorry for the delay! I’ve pushed an update with a new approach.

webpro:
Thank you! Took me a while to get it to an acceptable level of ROI/overhead.

webpro:

PR #1935, opened 2026-08-10
Fix Supported `lint-staged` Configs

joealden:
`lint-staged` doesn't use `lilconfig` (anymore?), and it's config file support list is different from it.

Derived from https://github.com/lint-staged/lint-staged/blob/main/lib/configFiles.js

Aside: I don't think `lilconfig` supports TS (https://github.com/antonk52/lilconfig/issues/61) - so I'm not sure `toLilconfig` should?


webpro:
Thanks! Good point re. lilconfig/TS, we can look into that separately.

webpro:
:rocket: _This pull request is included in v6.32.2. See [Release 6.32.2](https://github.com/webpro-nl/knip/releases/tag/knip%406.32.2) for release notes._

PR #1942, opened 2026-08-14
Fix nested gitignore negation inside ignored dot-directories (#1913)

### LINKED ISSUE #1913: 🐛 Nested .gitignore negation inside a dot-prefixed ignored directory un-ignores it (residual #1891 case on 6.29.0)
### Prerequisites

- [x] I'm using the latest version (6.29.0)

### Reproduction

Eight files, no dependencies beyond knip itself. Paste this into an empty directory:

```sh
git init -q .
printf '{"name":"repro","version":"1.0.0","main":"index.js"}\n' > package.json
printf 'export const used = 1;\n'      > index.js
printf 'node_modules\n.cache/\n'       > .gitignore

mkdir -p .cache/.bun/stubs
printf 'export const leaked = 1;\n'    > .cache/.bun/leak.js
printf '!*\n'                          > .cache/.bun/stubs/.gitignore   # <-- the trigger

npx knip
```

Result:

```
Unused files (1)
.cache/.bun/leak.js
```

git disagrees:

```sh
git check-ignore -v .cache/.bun/leak.js
# .gitignore:2:.cache/	.cache/.bun/leak.js
```

Delete `.cache/.bun/stubs/.gitignore` and rerun — the finding disappears. Nothing else changes.

I'm happy to push this to a public repo if you'd prefer a clone-able reproduction; it seemed shorter to inline it.

### Description of the issue

A `.gitignore` negation in a file **inside** a git-ignored directory cancels the ignore of that directory, and knip then analyzes and reports files under it.

This is the same ancestor-shadowing class as #1891 (fixed in 6.28.0), but it survives on 6.29.0 when the ignored path is dot-prefixed. Three conditions have to coincide, and I isolated each with a one-variable control:

**1. The crawl descends into the ignored directory (readdir order).**
`findAndParseGitignores` learns the root `.gitignore` patterns only when the fdir crawl *reaches that file*, but `exclude` decides whether to descend at the moment a directory is first encountered. A directory whose name sorts before `.gitignore` — anything dot-prefixed — is therefore queued while only `GLOBAL_IGNORE_PATTERNS` is loaded. `--debug` confirms the descent:

```
[*] Parsed gitignore files
{ gitignoreFiles: [ '.git/info/exclude', '.gitignore', '.cache/.bun/stubs/.gitignore' ] }
```

Descendants don't get caught on the way down either: the walk matcher holds the un-expanded `**/.cache`, which matches `.cache` but not `.cache/.bun`.

**2. The nested negation deletes the parent's ignore pattern.**
`!*` becomes the un-ignore `.cache/.bun/stubs/*`. The shadow pass walks each un-ignore's ancestor paths — which includes bare `.cache` — and deletes any ignore pattern matching one. `picomatch('**/.cache')('.cache')` is `true`, so `**/.cache` is removed from `cachedGitIgnores`, and the project glob no longer prunes the directory.

**3. `reconcileGitignoredPaths` misses dot-prefixed subtrees.**
The reconcile pass should still filter the result, and for most paths it does. But picomatch's `**` does not cross a dot-prefixed segment without `{ dot: true }`:

```js
const m = picomatch(['**/.cache', '**/.cache/**']);
m('.cache/leak.js')             // true
m('.cache/pnpm/x.js')           // true
m('.cache/.bun/install/x.js')   // false   <-- escapes the safety net
```

Controls, each changing exactly one variable:

| ignored dir | file lives under | result |
|---|---|---|
| `.cache/` | `.cache/.bun/` | **reported as unused** |
| `cache/` | `cache/.bun/` | clean — sorts after `.gitignore`, never descends (condition 1 absent) |
| `.cache/` | `.cache/bun/` | clean — reconcile catches it (condition 3 absent) |

### Why this isn't just a synthetic case

I hit it on a real project. Replit places the workspace's XDG cache root **inside** the repo as `.cache/`, and Bun's install cache sits at `.cache/.bun/` — dot-prefixed twice, satisfying conditions 1 and 3. The trigger file is shipped inside the published **tailwindcss** tarball, at `stubs/.gitignore`, containing exactly `!*`. So any project that has tailwindcss in a bun cache under a dot-prefixed ignored directory hits this.

`.cache/` was in `.gitignore` and `git status` was clean, but knip reported **12,087 phantom unused files** out of 12,147 — and precisely the `.cache/.bun/**` subtree, while `.cache/pnpm`, `.cache/ms-playwright` and `.cache/typescript` stayed correctly ignored, exactly as condition 3 predicts.

The same repo is clean on macOS, because there the directory doesn't exist. Being readdir-order dependent, this presents as "works on my machine / broken in CI (or vice versa)", which is what made it expensive to track down.

### Suggested direction

Not a patch, just where I'd look: parsing the root (and any ancestor) `.gitignore` *before* starting the crawl, rather than discovering it during the crawl, would remove condition 1 and make the behavior order-independent. Whether the reconcile matcher should use `{ dot: true }` is a separate call — it would close condition 3 but not the underlying descent.

### Versions

- knip 6.29.0
- node 24.x
- macOS 15 (APFS) — also observed on Replit's Linux workspace

DSeaStar:
## Summary

- `findAndParseGitignores` crawled for `.gitignore` files before loading the repo-root one. On filesystems where a dot-prefixed directory is seen first, the walker entered that tree with only the global ignore patterns, then a nested `!*` un-ignored files git itself never sees (`git check-ignore` still reports them ignored).
- Load the root `.gitignore` up front (same as ancestor ignores and `.git/info/exclude`) and skip a file if `addFile` has already seen it.
- Fixture matches the case from #1913: root ignore of a dot-directory, a sibling file that used to leak, a nested `!*`, and a file that `!*` would match if that rule were reachable.

## Test plan

- [x] `bun test test/infra/gitignore-nested-negation.test.ts` (fails without the preload; passes with it)
- [x] Existing gitignore tests still pass (`gitignore`, `gitignore-star-negation`, `gitignore-negated-basename`, `gitignore-mixed-patterns`, `find-and-parse-gitignores`)

Fixes #1913


webpro:
:rocket: _This pull request is included in v6.32.3. See [Release 6.32.3](https://github.com/webpro-nl/knip/releases/tag/knip%406.32.3) for release notes._

PR #1947, opened 2026-08-15
Exit with code 2 when a plugin config file fails to load

WooWan:
When a plugin config file (e.g. `vite.config.ts`) fails to load, Knip logs the error and continues, but the exit code previously ignored it.

Before:

| Scenario | Exit |
| --- | ---: |
| load error, no other findings | 0 |
| load error, with findings | 1 |
| load error, `--no-exit-code` | 0 |

After: every case above exits `2`.

Analysis still runs to completion and findings are still reported. `--no-exit-code` continues to suppress lint findings, but not config load errors, consistent with how Knip already handles failures in its own configuration.

The existing CLI test was only green because its fixture also had unrelated findings. A new `load-error-only` fixture verifies that a config load error alone exits `2`.

Docs are updated accordingly.


WooWan:
Although this is a bug fix, changing the exit code could cause existing services to suddenly start failing after a minor or patch upgrade. It may therefore be safer to ship this in the next major release, but I’ll leave that decision to the maintainers. 🙇

packages/knip/src/WorkspaceWorker.ts
```diff
@@ -475,6 +476,7 @@ export class WorkspaceWorker {
               } catch (error) {
                 if (!(error instanceof Error)) throw error;
                 hasLoadConfigError = true;
+                if (isLoaderError(error)) hasConfigLoadErrors = true;
```
webpro: P1 This narrows to `LoaderError`, but the `try` block also covers `plugin.resolveConfig` and `filterTransitiveDependencies`, and both `logError` calls run for anything they throw. A plugin throwing a plain `Error` prints the same two lines, discards the cache entry, and still exits 0.

`hasLoadConfigError` on the line above already carries this fact for the cache. Set the outer flag from that one and drop the `isLoaderError` import. The cache already treats every error here as a failed load; the exit code should agree.

packages/knip/src/WorkspaceWorker.ts
```diff
@@ -556,7 +558,7 @@ export class WorkspaceWorker {
 
     debugLogArray(wsName, 'Plugin dependencies', () => compact(inputs.map(input => toDebugString(input, rootCwd))));
 
-    return inputs;
+    return { inputs, hasConfigLoadErrors };
```
webpro: P1 Returning a boolean here changes the return type and ripples through `build`, `run` and `cli`. `IssueCollector` already carries run-level facts of exactly this kind in `tagHints` and `configurationHints`, `collector` is in scope in `build`, and `WorkspaceWorker` already takes callbacks such as `handleInput`. One more callback keeps `runPlugins(): Input[]` and leaves `build` and `run` untouched.

It also puts the fact in `results`, within reach of `main()` and the reporters. As submitted, the `github-actions` reporter has nothing to annotate, so the job turns red with no summary entry explaining why. That annotation can be a separate change.

packages/knip/src/cli.ts
```diff
@@ -105,6 +105,7 @@ const main = async () => {
     if (
       !args['no-exit-code'] &&
       (totalErrorCount > Number(args['max-issues'] ?? 0) ||
+        hasConfigLoadErrors ||
```
webpro: P1 This sits inside the `!args['no-exit-code']` guard and behind `totalErrorCount`, so the flag suppresses the load failure and a findings exit can claim the code ahead of it. Both halves matter: incomplete input then reports as a clean run for anyone who has opted out of exit codes, and otherwise reports as ordinary findings.

Set it before this branch instead, after the reporters have run, reading the flag from `results` once it moves into `IssueCollector`:

```ts
if (hasConfigLoadErrors) {
  process.exitCode = 2;
  return;
}
```

packages/knip/fixtures/plugin-config/load-error-only/knip.json
```diff
@@ -0,0 +1,8 @@
+{
```
webpro, line 1: P3 This turns the fixture into the clean-report case and drops the case it covered before, a load failure alongside genuine issues. Both are worth keeping: a copy under `fixtures/plugin-config/load-error-only` carrying this file leaves the original fixture and its assertions intact.

webpro (changes requested):
Thanks for the pull request, and for tracking down why the existing test was green.

The gap is real. With the `knip.json` from this PR in place, `main` reports the failure and still exits 0:

```
$ knip --no-progress
ERROR: Error loading vite.config.ts (Cannot find module './missing-bootstrap-output.js')
ERROR: Please fix or visit https://knip.dev/reference/known-issues
$ echo $?
0
```

Without it the exit code comes from the unused `vite` devDependency, the unlisted binary and the unresolved import, so the CLI test never proved anything about the load failure. Your reading is correct.

On the release question from your comment: exiting non-zero here has been the intended behavior since `0914bd383` added that test, so this is a bug fix rather than a breaking change and does not need to wait for a major.

The code should be 2 rather than 1 though, and `--no-exit-code` should not suppress it. A `vite.config.ts` that fails to load is not a finding about the user's code. It means the analysis input was incomplete, so every remaining result may be wrong. Knip already treats that class this way: a `knip.config.ts` that fails to load exits 2 even with `--no-exit-code`, because the error escapes to the catch block in `cli.ts`, which sets the code regardless of the flag. A plugin config file is the same failure, only non-fatal, and `reference/cli.md` already describes 2 as "Knip did not run successfully due to bad input or internal error". Anyone who cannot fix the file keeps the documented escape hatch of overriding the plugin's `config`, which skips the load attempt entirely.

That moves the assertions in `load-error.test.ts`: 1 becomes 2 in the CLI test and in the cache test, and the `--no-exit-code` case expects 2 rather than 0, with a name to match.

P2 The exit code table in `reference/cli.md` and the "Exceptions from config files" section in `reference/known-issues.md` both need the new behavior: a plugin config file that fails to load exits 2, and `--no-exit-code` does not suppress it.

The branch also conflicts with `main` in `cli.ts`, which now reads `options.maxIssues` instead of parsing the argument inline. Merging `main` into the branch updates the pull request without a force push.

packages/knip/src/reporters/util/configuration-hints.ts
```diff
@@ -108,7 +108,7 @@ const UNCONFIGURED_MIN_FILES = 20;
 const UNCONFIGURED_MIN_RATIO = 0.2;
 
 export const finalizeConfigurationHints = (
-  results: Results,
+  results: Pick<Results, 'counters' | 'issues' | 'includedWorkspaceDirs' | 'configurationHints'>,
```
WooWan, line 111: Results now requires hasConfigLoadErrors, which isn't in ReporterOptions. Narrowed this parameter to the four fields it uses. It can be extended when GitHub Actions annotations are added.

packages/knip/src/WorkspaceWorker.ts
WooWan: Thanks for the detailed feedback! Your approach makes more sense. Applied in [4765a10](https://github.com/webpro-nl/knip/pull/1947/commits/4765a10769949395962aacce473b28c22ea0dbd4).

packages/knip/src/WorkspaceWorker.ts
WooWan: Applied in [4765a10](https://github.com/webpro-nl/knip/pull/1947/commits/4765a10769949395962aacce473b28c22ea0dbd4)

packages/knip/src/cli.ts
WooWan: Applied in [4765a10](https://github.com/webpro-nl/knip/pull/1947/commits/4765a10769949395962aacce473b28c22ea0dbd4).

packages/knip/fixtures/plugin-config/load-error-only/knip.json
WooWan: Applied it in [fec9d99](https://github.com/webpro-nl/knip/pull/1947/commits/fec9d99dc997af891281e45cd7f10220b3cbf92f).

WooWan:
Added docs in [81e0474](https://github.com/webpro-nl/knip/pull/1947/commits/81e04748de7ed3163f3a7fc18d614594e0872db1). English isn't my first language, so feel free to tweak the wording as needed.

packages/knip/src/reporters/util/configuration-hints.ts
webpro: P2 This narrowing works around a gap that is better closed directly. `hasConfigLoadErrors` reaches `Results` but not `ReporterOptions`, which `cli.ts` assembles by hand, so reporters and preprocessors still cannot see it. Reaching them was half the reason for moving the fact into `IssueCollector`.

`finalizeConfigurationHints` is also exported from `knip/session` and consumed by `packages/mcp-server/src/tools.js`, so this changes a published signature as a side effect of the type gap.

Adding `hasConfigLoadErrors` to `ReporterOptions` and passing it through `initialData` closes the gap and makes this narrowing unnecessary.

packages/knip/src/WorkspaceWorker.ts
```diff
@@ -475,6 +479,7 @@ export class WorkspaceWorker {
               } catch (error) {
                 if (!(error instanceof Error)) throw error;
                 hasLoadConfigError = true;
+                this.handleConfigLoadError();
```
webpro, line 482: P2 Nothing pins this line. Restoring `if (isLoaderError(error))` in front of it leaves all 1096 tests passing, and that guard is the behavior the fix exists to remove.

A config file that loads and then throws inside `resolveConfig` covers it. A `vite.config.ts` of `export default { test: { projects: 42 } }` is enough:

```
ERROR: Error loading vite.config.ts (config.test.projects is not iterable)
```

That exits 2 on this branch and 0 with the guard back.

packages/knip/fixtures/plugin-config/load-error-only/knip.json
```diff
@@ -0,0 +1,8 @@
...
+  "ignoreDependencies": ["vite"],
+  "ignoreBinaries": ["vite"],
+  "ignoreIssues": {
+    "vite.config.ts": ["unresolved"],
+    "missing-bootstrap-output.js": ["files"]
```
webpro: P3 This entry does nothing here. It carries the cache recovery test in the sibling fixture, which writes `missing-bootstrap-output.js` partway through; nothing creates that file in this fixture. Removing the line leaves the output identical.

webpro (changes requested):
Every finding from the previous round is addressed, and the behavior holds up under direct checks rather than just the diff. A `vite.config.ts` that loads and then throws inside `resolveConfig` now exits 2, where the `isLoaderError` narrowing let it exit 0, and the same fixture with the load failure removed still exits 0. Thanks for the quick turnaround.

Two smaller things and a nit, all inline.

packages/knip/fixtures/plugin-config/load-error-only/knip.json
WooWan: Thanks, applied it [302da6a](https://github.com/webpro-nl/knip/pull/1947/commits/302da6a0380e99109536d506fa3b17b0f5483fc7)

packages/knip/src/reporters/util/configuration-hints.ts
WooWan: Updated ReporterOptions to include hasConfigLoadErrors and restored the Results parameter type. After merging main, toReporterOptions() passes the flag through to reporters and preprocessors, applied it in [4c75139](https://github.com/webpro-nl/knip/pull/1947/commits/4c751395f4d7af39e03a41b8332284cb74756c76)

packages/knip/src/WorkspaceWorker.ts
WooWan: Added a regression test using projects: 42 that expects exit code 2, including with --no-exit-code. Confirmed that restoring the isLoaderError guard makes both cases fail. [f241e7c](https://github.com/webpro-nl/knip/pull/1947/commits/f241e7cd418c22d0a8ee5a42aa2c59e8db977e5f), [f37936a](https://github.com/webpro-nl/knip/pull/1947/commits/f37936a59b3ecf766a38a3ef6764751d489fcf65)

webpro:
Thank you! 🌱

webpro:
:rocket: _This pull request is included in v6.35.1. See [Release 6.35.1](https://github.com/webpro-nl/knip/releases/tag/knip%406.35.1) for release notes._

PR #1938, opened 2026-08-12
fix(tsup): match config files with .cts and .mts extensions

codeAnqiang-ma:
Fixes #1937

## What

`packages/knip/src/plugins/tsup/index.ts` matched five of the seven config
file names tsup accepts:

```diff
-const config = ['tsup.config.{js,ts,cjs,mjs,json}', 'package.json'];
+const config = ['tsup.config.{js,ts,cjs,mjs,cts,mts,json}', 'package.json'];
```

## Why

tsup resolves its config against a fixed file list in
[`src/load.ts`](https://github.com/egoist/tsup/blob/dev/src/load.ts) — the same
file this plugin links in its comment:

```js
files: [
  'tsup.config.ts',
  'tsup.config.cts',
  'tsup.config.mts',
  'tsup.config.js',
  'tsup.config.cjs',
  'tsup.config.mjs',
  'tsup.config.json',
  'package.json',
]
```

A project on `tsup.config.cts` (e.g. a CJS config inside a `"type": "module"`
package) or `tsup.config.mts` therefore got a false positive on the config file
itself, and nothing referenced from its `entry` was marked as a production
entry either. Same class of gap as #1919 (playwright), #1933 (oxfmt), #1934
(oxlint).

## Reproduction

Extended `fixtures/plugins/tsup` with one config per missing extension, each
pointing at its own entry file. Before the change, running knip in that
fixture:

```
Unused files (4)
entry-6.ts
entry-7.ts
tsup.config.cts
tsup.config.mts
```

After the change the same command reports nothing, which is what the updated
counters in `test/plugins/tsup.test.ts` assert (`processed`/`total` 8 → 12).
The updated test fails without the one-line fix and passes with it.

## Tests

Ran from `packages/knip`:

- `bun test test/plugins/tsup.test.ts` — pass (also passes with `node --test`)
- `bun test test/plugins/` — 368/368 pass
- `pnpm test --runtime bun --smoke` — 945 pass, 5 fail; the same 5
  (`infra/cache-glob-*`, `plugin-config/load-error`) fail identically on a
  clean `main` checkout on this machine, so they are unrelated to this change
- root `pnpm run ci` (oxlint + oxfmt --check + installed-check) — pass

## Note

Written with AI assistance; I reproduced the false positive locally and
reviewed every change. Happy to adjust the fixture shape if you prefer a
smaller diff.

PR #1919, opened 2026-08-01
fix(playwright): match config files with .cjs, .cts and .mts extensions

davidpavlovschi:
## What

`packages/knip/src/plugins/playwright/index.ts` matched three of the six config
file extensions Playwright accepts:

```diff
-const config = ['playwright.config.{js,ts,mjs}'];
+const config = ['playwright.config.{js,cjs,mjs,ts,cts,mts}'];
```

## Why

Playwright resolves `playwright.config` against a fixed extension list in
`packages/playwright/src/common/configLoader.ts`
(https://github.com/microsoft/playwright/blob/main/packages/playwright/src/common/configLoader.ts):

```js
for (const ext of ['.ts', '.js', '.mts', '.mjs', '.cts', '.cjs']) {
  const configFile = resolveConfig(path.resolve(directory, 'playwright.config' + ext));
  ...
}
```

A project on `playwright.config.cjs`, `playwright.config.cts` or
`playwright.config.mts` therefore got a false positive on the config file
itself, and nothing referenced from it (reporters, `globalSetup`,
`globalTeardown`, `testDir`/`testMatch` entries) was resolved either.

## Reproduction

Extended `fixtures/plugins/playwright` with one config per missing extension,
each pointing at its own custom reporter. Before the change, running knip in
that fixture:

```
Unused files (6)
commonjs-reporter.cjs
commonjs-typescript-reporter.cts
esmodule-reporter.mts
playwright.config.cjs
playwright.config.cts
playwright.config.mts
```

After the change the same command reports nothing, which is what the updated
counters in `test/plugins/playwright.test.ts` assert (`processed`/`total` 3 → 9).

## Tests

Ran from `packages/knip` on Node 26.5.0 (no `bun` on this machine):

- `node --test test/plugins/playwright.test.ts test/plugins/playwright2.test.ts test/plugins/playwright-ct.test.ts test/plugins/playwright-ct2.test.ts` — 4/4 pass
- `node --test $(ls test/plugins/*.test.ts)` — 361/361 pass
- root `oxlint` — exit 0; root `oxfmt --check` — all files correctly formatted

## Note

Written with AI assistance. The reproduction above and the test runs are the
evidence behind it; happy to adjust the fixture shape or drop it to a single
extension if you prefer a smaller diff.

PR #1948, opened 2026-08-15
fix(webdriver-io): match all supported config file extensions

solomonaustin:
The WebdriverIO plugin matches `wdio.conf.{js,ts}`, but WebdriverIO resolves its config file across six extensions:

```ts
// packages/create-wdio/src/constants.ts
export const SUPPORTED_CONFIG_FILE_EXTENSION = ['js', 'ts', 'mjs', 'mts', 'cjs', 'cts']
```

`canAccessConfigPath` tries each of these against the config path, and `wdio-cli`'s `run` handler loads `tsx` for `.ts`, `.tsx`, `.mts` and `.cts` configs.

So `wdio.conf.{cjs,mjs,cts,mts}` are all valid, but Knip left them unanalyzed — reporting the config file itself as unused and the dependencies it references (framework, runner, reporters) as unused too. `.mts` is the extension [WebdriverIO recommends](https://webdriver.io/docs/typescript/) for ESM + TypeScript setups.

Extending the existing fixture reproduces it before the fix — 4 unused files and 4 unused dependencies, exactly the four unmatched configs and the reporters only they reference.

Each fixture config references a distinct reporter (`spec`, `dot`, `junit`, `allure`, `json`) rather than sharing one, so that a regression in any single extension surfaces as an unused dependency instead of being masked by the other configs.

Same shape as #1919 (playwright), #1933/#1934 (oxfmt/oxlint) and #1938 (tsup).

### Testing

- `node --test test/plugins/webdriver-io.test.ts` — passes (fails as described above without the plugin change)
- `node --test test/plugins/*.test.ts` — 368/368 pass
- `oxlint` and `oxfmt --check` clean

### DIFF
--- a/packages/knip/fixtures/plugins/webdriver-io/package.json
+++ b/packages/knip/fixtures/plugins/webdriver-io/package.json
@@ -7,6 +7,11 @@
     "@wdio/cli": "*",
     "@wdio/mocha-framework": "*",
     "@wdio/concise-reporter": "*",
-    "@wdio/browser-runner": "*"
+    "@wdio/browser-runner": "*",
+    "@wdio/spec-reporter": "*",
+    "@wdio/dot-reporter": "*",
+    "@wdio/junit-reporter": "*",
+    "@wdio/allure-reporter": "*",
+    "@wdio/json-reporter": "*"
   }
 }
--- a/packages/knip/fixtures/plugins/webdriver-io/wdio.conf.cjs
+++ b/packages/knip/fixtures/plugins/webdriver-io/wdio.conf.cjs
@@ -1,0 +1,3 @@
+exports.config = {
+  reporters: ['spec'],
+};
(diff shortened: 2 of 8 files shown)


solomonaustin:
@webpro whenever you get a chance — no rush. Flagging mainly because CI hasn't run yet: the workflow is pending approval since this is my first PR here, so the green/red signal you'd normally review against isn't there.

Locally: `node --test test/plugins/*.test.ts` is 368/368, and `oxlint` / `oxfmt --check` are clean.

Happy to adjust anything — in particular, if six config files in one fixture is more than you want, I can trim it to just `.mts`/`.cts` and keep the distinct-reporter trick that makes each extension independently verifiable.

I also have `cypress` lined up as the same class of fix (`cypress.config.{js,ts,mjs,cjs}` while Cypress supports `.mts`/`.cts` — it has explicit `cypress.config.mts` tests and `shouldLoadConfigAsEsm` handles it). Let me know if you'd prefer those batched into one PR per plugin, or a single sweep across the plugins with incomplete extension globs.

webpro:
:rocket: _This pull request is included in v6.32.3. See [Release 6.32.3](https://github.com/webpro-nl/knip/releases/tag/knip%406.32.3) for release notes._

PR #1958, opened 2026-08-19
fix: count enum element access as a whole-object read

### LINKED ISSUE #1956: 🐛 Enum element access does not count as using the enum's members
### Prerequisites

### Reproduction url

https://github.com/brammitch/knip-unused-enum-repro

### Reproduction access

### Description of the issue

# Enum element access does not count as using the enum's members

## Summary

`Object.values(SomeEnum)` suppresses `Unused exported enum members` for that enum entirely. Element access — `SomeEnum[expr]` — does not, and every member is reported unused.

Both are whole-object reads, and in neither case can the individual members read be known statically. knip already has the "the whole object is read, so don't report its members" behavior; it just doesn't apply it to element access.

Separately, element access with a **numeric literal** is statically resolvable and is also missed: `StatusCodeLiteral[3000]` reverse-maps to `warning = 3000`, and `warning` is still reported unused. Deleting `warning` would break that expression.

I did see #1890 that may have been similar to this, but the linked reproduction is a 404 now, so I could not verify.

## Reproduction

```
npm install
npx knip
```

Four identical enums, one per access pattern, so a single run shows all four outcomes.

## Actual vs expected

| Enum                | Access pattern in `src/utils/`             | Members reported | Expected |
| ------------------- | ------------------------------------------ | ---------------- | -------- |
| `StatusCodeValues`  | `Object.values(StatusCodeValues)`          | **0**            | 0        |
| `StatusCodeDynamic` | `StatusCodeDynamic[Number.parseInt(code)]` | **5**            | 0        |
| `StatusCodeLiteral` | `StatusCodeLiteral[3000]`                  | **5**            | 0, or 4  |
| `StatusCodeDirect`  | `StatusCodeDirect.normal`                  | **4**            | 4        |

`StatusCodeDirect` is a control: only `normal` is referenced, the other four are correctly reported, and this is the behavior working as intended.

`StatusCodeValues` and `StatusCodeDynamic` are the comparison. The two access patterns carry the same amount of static information about which members are read — none — and are treated differently.

For `StatusCodeLiteral`, either answer would be defensible: 0 if element access is treated as a whole-object read like `Object.values`, or 4 if the literal index is resolved to `warning`. Five is wrong under both readings.

## Why this matters

The documented workaround is a `@knipignore` tag plus `tags: ["-knipignore"]`. That works, but knip then reports the tag itself as an unused tag on every run, so there is no configuration in which the output is clean (I'm going to open a second issue for this). Enums keyed by codes from an external API are a common shape, and each one costs either a permanent false positive or a permanent wrong hint.

addielaruee:
## Problem

Element access does not count as using an enum's members. `SomeEnum[expr]` reports every member as unused, while `Object.values(SomeEnum)` correctly suppresses the same reports. Both are whole-object reads and neither can statically resolve which members are read, so they should be treated the same.

From the reproduction in #1956:

| Access pattern | Members reported (before) | After |
| --- | --- | --- |
| `Object.values(StatusCodeValues)` | 0 | 0 |
| `StatusCodeDynamic[Number.parseInt(code)]` | 5 | 0 |
| `StatusCodeLiteral[3000]` | 5 | 0 |
| `StatusCodeDirect.normal` | 4 | 4 |

## Fix

In `handleMemberExpression`, computed element access on a non-namespace enum import is now treated as a whole-object read: the enum is marked `enumerated`, the same way `Object.keys/values/entries` already do in the call visitor. Dot access (`Enum.member`) and string-literal access (`Enum['member']`) are unchanged and still resolve to the specific member, so genuinely unused members are still reported.

Numeric-literal access such as `Enum[3000]` is also treated as a whole-object read, so its count goes to 0, which the issue lists as an acceptable outcome. Statically reverse-mapping the literal to its member is left out to keep the change minimal and predictable.

## Tests

Added `fixtures/types/enum-members-element-access` and `test/types/enum-members-element-access.test.ts` covering dynamic and numeric-literal element access. The existing `enum-members` and `enum-members-enumerated` tests still pass, confirming dot-access members are still reported and the whole-object behavior stays consistent.

Fixes #1956

### DIFF
--- a/packages/knip/src/typescript/visitors/members.ts
+++ b/packages/knip/src/typescript/visitors/members.ts
@@ -1,8 +1,18 @@
 import type { MemberExpression, JSXMemberExpression } from 'oxc-parser';
 import { OPAQUE } from '../../constants.ts';
+import type { ImportMaps } from '../../types/module-graph.ts';
 import { addValue } from '../../util/module-graph.ts';
 import { getStringValue, isStringLiteral } from '../ast-nodes.ts';
 import { isShadowed, type WalkState } from './walk.ts';
+
+function markWholeObjectRead(internalImport: ImportMaps, id: string) {
+  internalImport.refs.add(id);
+  (internalImport.enumerated ??= new Set()).add(id);
+}
+
+function isNumericKey(value: string | undefined) {
+  return value != null && value !== '' && !Number.isNaN(Number(value));
+}
 
 export function handleMemberExpression(node: MemberExpression, s: WalkState) {
   if (node.object.type === 'MemberExpression' && node.object.object.type === 'MemberExpression') {
@@ -20,9 +30,17 @@ export function handleMemberExpression(node: MemberExpression, s: WalkState) {
         if (node.computed === false && node.property.type === 'Identifier') {
           memberName = node.property.name;
         } else if (node.computed && isStringLiteral(node.property)) {
-          memberName = getStringValue(node.property);
+          const value = getStringValue(node.property);
+          if (!_import.isNamespace && isNumericKey(value)) {
+            markWholeObjectRead(internalImport, _import.importedName);
+            return;
+          }
+          memberName = value;
         } else if (node.computed && _import.isNamespace) {
           addValue(internalImport.import, OPAQUE, s.filePath);
+          return;
+        } else if (node.computed) {
+          markWholeObjectRead(internalImport, _import.importedName);
           return;
         }
         if (memberName) {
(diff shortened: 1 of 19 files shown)


packages/knip/src/typescript/visitors/members.ts
```diff
@@ -24,6 +24,10 @@ export function handleMemberExpression(node: MemberExpression, s: WalkState) {
         } else if (node.computed && _import.isNamespace) {
           addValue(internalImport.import, OPAQUE, s.filePath);
           return;
+        } else if (node.computed) {
+          internalImport.refs.add(localName);
+          (internalImport.enumerated ??= new Set()).add(localName);
```
webpro: P1: `enumerated` uses the local import name, but analysis expects the exported identifier on the defining file. Consequently, both import aliases and barrel re-exports still report all members. Each focused negative control produced exactly five false positives, with unsafe fixes attached.

packages/knip/src/typescript/visitors/members.ts
```diff
@@ -24,6 +24,10 @@ export function handleMemberExpression(node: MemberExpression, s: WalkState) {
         } else if (node.computed && _import.isNamespace) {
           addValue(internalImport.import, OPAQUE, s.filePath);
           return;
+        } else if (node.computed) {
```
webpro, line 42: P1: collection only handles a directly imported identifier. These equivalent accesses remain broken:

- `namespace.Status[key]`
- `const Alias = Status; Alias[key]`
- dynamic access to an enum in its defining file

The all-members-used fact needs a canonical, alias/re-export-aware graph owner rather than another syntax-specific branch.

webpro, line 42: P2: string literals bypass the new fallback. `Status[2000]` now suppresses all reports, while equivalent `Status['2000']` still reports every member. TypeScript accepts this syntax, and numeric enums provide reverse value-to-name mappings. Either precise or conservative handling is defensible; reporting every member is not. [TypeScript enum documentation](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings).

webpro (changes requested):
PR #1958 needs changes.

The product decision is sound: non-statically-resolvable `Enum[expr]` means every member is potentially used. Named string access remains precise; conservative treatment of numeric literals is acceptable. The implementation currently realizes that policy only for direct, unaliased named imports.

webpro:
Thanks for the PR! Happy to merge, but please look into the findings. You can either ignore the ecosystem tests failures or rebase on top of main.

addielaruee:
Thanks for the detailed review. I reworked this so the whole-object read is recorded against the exported identifier on the defining file and resolved through the graph, instead of the local binding via a syntax branch. Rebased on `main`.

**P1 (canonical, alias/re-export-aware owner).** `markWholeObjectRead` now keys the `enumerated`/`refs` mark on `_import.importedName`, and the analyze-time check goes through a new `explorer.isEnumerated` walk that follows re-exports and aliased re-exports the same way `isReferenced` does. Element access now resolves for:

- import aliases: `import { Status as S }; S[key]`
- local aliases: `const Alias = Status; Alias[key]`
- namespace access: `import * as ns; ns.Status[key]`
- pass-through, aliased and star re-exports through a barrel

Each of these is covered by `fixtures/types/enum-members-element-access-resolution`, where every enum is consumed through exactly one pattern so a regression in any single one surfaces.

The remaining sub-case you listed, dynamic access to an enum in its own defining file, I could not reproduce as a false positive: an exported enum used only via `Status[key]` in the same file reports clean today, both before and after this change. If you have a repro where it does report, I will add it.

**P2 (string literals).** A numeric key cannot name a member and at most resolves a numeric enum's reverse mapping, so `Status[2000]` and `Status['2000']` are now both treated as whole-object reads. A named string key still resolves precisely: `Status['warning']` reports the other members. `fixtures/types/enum-members-element-access-string-key` asserts both halves in one run.

Verified: `tsc`, the enum and namespace suites, and a sweep of the exports/imports/entry/fix directories all pass; `oxlint` and `oxfmt --check` are clean.

webpro:
Thanks again!

webpro:

PR #1962, opened 2026-08-20
Don't report a tag as unused when it suppresses member issues (resolve #1957)

### LINKED ISSUE #1957: 🐛 `tags` exemption is reported back as an "Unused tag" hint
### Prerequisites

### Reproduction url

https://github.com/brammitch/knip-unused-tag-repro

### Reproduction access

### Description of the issue

# `tags` exemption is reported back as an "Unused tag" hint

## Summary

A `@knipignore` tag on an enum declaration, exempted via `tags: ["-knipignore"]`, correctly suppresses the `Unused exported enum members` findings for that enum — and is then reported as an **unused tag**. It is not unused: removing it brings all five findings straight back.

This leaves no clean state. Keeping the tag means living with a permanently wrong hint; removing it means living with five false-positive findings. In a CI setup that treats knip output as actionable, the hint reads as a regression on every run.

I came across this issue while trying to suppress the warnings I was experiencing for #1956. I thought it best to open up a second issue as they are distinct problems.

## Reproduction

```
npm install
npx knip
```

`src/models/status.model.ts` declares an enum whose members are reached by reverse numeric index (`StatusCode[Number.parseInt(code)]`) from `src/utils/status.utils.ts`, using a value that would come from an API at runtime. No static analysis can see the individual members being used, which is what the `@knipignore` tag is for.

## Actual vs expected

| State                                        | knip output                          | exit |
| -------------------------------------------- | ------------------------------------ | ---- |
| **Baseline** — tag present, `tags` configured | `Unused tag ... → @knipignore` (1)   | 0    |
| **A** — `@knipignore` removed, `tags` kept    | `Unused exported enum members` (5)   | 1    |
| **B** — `@knipignore` kept, `tags` removed    | `Unused exported enum members` (5)   | 1    |

Controls A and B both demonstrate that the tag in the baseline is doing real work. Expected in the baseline is **no output at all**: the tag suppressed five findings, so it is not unused.

To run the controls, delete the ` * @knipignore` line from `src/models/status.model.ts` (A) or the `tags: ["-knipignore"]` line from `knip.config.ts` (B).

## Note on file layout

The enum must be **consumed from a different file than the entry file** to reproduce. With the dynamic index access in the entry file itself, knip resolves the members as used and reports nothing in any of the three states — the tag really is unused there, and the hint is correct.

thribhuvan003:
Resolves #1957.

A tag that excludes an export also skips the enum/namespace member analysis, since that block sits behind the `continue` on `isIgnored`. The hint only checked whether the export itself was referenced, so `@knipignore` on a referenced enum was reported as unused while it was still suppressing five member issues. Removing the tag brought those five back, so neither keeping nor removing it gave a clean run.

This skips the hint when the export has members the tag gates.

Added `fixtures/tags-hints/tag-suppresses-enum-members` with a test for both sides: no hint when the tag is excluded, and the same five members reported when it isn't. The second one fails without the source change.

One thing worth your call: the guard is deliberately blunt, so a tag on an export whose members are all referenced no longer gets a hint either. Narrowing that needs the member walk to run before the hint is decided — happy to do it that way if you prefer.

### DIFF
--- a/packages/knip/fixtures/tags-hints/tag-suppresses-enum-members/index.ts
+++ b/packages/knip/fixtures/tags-hints/tag-suppresses-enum-members/index.ts
@@ -1,0 +1,3 @@
+import { labelFor } from './utils';
+
+console.log(labelFor('3000'));
--- a/packages/knip/fixtures/tags-hints/tag-suppresses-enum-members/package.json
+++ b/packages/knip/fixtures/tags-hints/tag-suppresses-enum-members/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "@fixtures/tag-suppresses-enum-members"
+}
--- a/packages/knip/fixtures/tags-hints/tag-suppresses-enum-members/status.ts
+++ b/packages/knip/fixtures/tags-hints/tag-suppresses-enum-members/status.ts
@@ -1,0 +1,12 @@
+/**
+ * Members are looked up by reverse numeric index, so they have no static references.
+ *
+ * @knipignore
+ */
+export enum StatusCode {
+  normal = 1000,
+  unknown = 2000,
+  warning = 3000,
+  critical = 4000,
+  nostatus = 5000,
+}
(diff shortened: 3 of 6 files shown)


webpro:
Thanks for digging into this. I think #1958 is the better direction: it fixes the underlying computed enum access analysis, so the tag from #1957 becomes genuinely unused and can be removed without bringing the member reports back. This change instead suppresses tag hints for every member-bearing export, including cases where the tag is actually stale. Let's resolve #1957 through #1958 rather than merge this approach.

PR #1971, opened 2026-08-26
fix: resolve pnpm catalogs above project root

### LINKED ISSUE #1963: 🐛 Unresolved catalog references reported when knip runs from a workspace subdirectory (regression in 6.30.0)
### Prerequisites

### Reproduction url

https://github.com/clemens/knip-catalog-issue-repro

### Reproduction access

### Description of the issue

Note: Reproduction and issue description created with Claude Code, but I've tested it manually as well to confirm it's not just hallucinated. It might be related to https://github.com/webpro-nl/knip/issues/1905, but looks pretty much like the inverse (callilng knip from within a workspace that uses catalog references to the root).

## Summary

Since **6.30.0**, running knip with its cwd inside a workspace package — rather than at the
pnpm workspace root — reports every `catalog:` specifier as an unresolved catalog reference
and exits 1. **6.29.0 is clean** on the same fixture.

The catalog entries do resolve correctly when knip runs from the workspace root, so this
looks like the catalog lookup becoming relative to the detected project root instead of
walking up to the `pnpm-workspace.yaml` that defines the catalog.

## Reproduction

Four files, no install required (`pnpm dlx` is enough):

**`pnpm-workspace.yaml`**
```yaml
packages:
  - pkg

catalog:
  zod: 4.4.3
```

**`package.json`**
```json
{
  "name": "repro-root",
  "private": true
}
```

**`pkg/package.json`**
```json
{
  "name": "pkg",
  "dependencies": {
    "zod": "catalog:"
  }
}
```

**`pkg/index.ts`**
```ts
import { z } from "zod";

export const schema = z.string();
```

### Run from `pkg/`

```console
$ cd pkg

$ pnpm dlx knip@6.29.0 --no-progress
$ echo $?
0

$ pnpm dlx knip@6.30.0 --no-progress
Unresolved catalog references (1)
zod  default  package.json:4:6
$ echo $?
1

$ pnpm dlx knip@6.32.2 --no-progress
Unresolved catalog references (1)
zod  default  package.json:4:6
$ echo $?
1
```

### Run from the workspace root — clean on 6.32.2

```console
$ pnpm dlx knip@6.32.2 --no-progress
$ echo $?
0
```

## Expected vs actual

**Expected:** `zod` resolves against the catalog in the `pnpm-workspace.yaml` above the
package, as it does on 6.29.0 and as it does from the workspace root on 6.32.2.

**Actual:** `Unresolved catalog references (1)` and exit 1.

## Scope of the regression

Verified on 6.32.2, all from `pkg/`:

| Variant | Result |
|---|---|
| Top-level `catalog:` + `"zod": "catalog:"` | ❌ `zod  default  package.json:4:6` |
| `catalogs.default:` + `"zod": "catalog:"` | ❌ `zod  default  package.json:4:6` |
| `catalogs.val:` + `"zod": "catalog:val"` | ❌ `zod  val  package.json:4:6` |

So it affects the default and named catalogs alike, and both declaration forms — it is not
specific to the bare `catalog:` shorthand.

Invocation matters:

| Invocation | Result on 6.32.2 |
|---|---|
| cwd = workspace root | ✅ clean |
| cwd = root, `--workspace pkg` | ✅ clean |
| cwd = root, `--directory pkg` | ❌ reports unresolved |
| cwd = `pkg/` | ❌ reports unresolved |

`--directory pkg` behaving like `cd pkg` is consistent with the cwd being what triggers it.

## Bisect

- `6.29.0` — clean
- `6.30.0` — first bad version
- `6.31.0`, `6.32.0`, `6.32.1`, `6.32.2` — all bad

## Possible cause

6.30.0 includes *"Track catalog references across scoped runs"* (#1905) and *"Include
workspace dependencies in scoped runs"* (#1901). #1905 was the inverse false positive —
catalog entries wrongly reported **unused** under `--workspace .` — so this may well be a
side effect of that fix. I could not find an existing issue for this direction. Apologies
if it is already covered by something I missed.

## Why this matters / current workaround

This pattern shows up in monorepos where each package runs knip as its own project with
its own config file (`pkg/knip.config.ts`), invoked per package (`cd pkg && knip`) so each
one is a separately reportable CI gate. There is no root config in that layout, so the
catalog definitions always sit above knip's project root.

The available workaround — invoke from the root with `--workspace <dir>` — means
consolidating the per-package configs into a single root config, which is a larger change
than a version bump. For now the practical option is pinning to 6.29.0.

## Environment

- knip: 6.29.0 (good) → 6.30.0–6.32.2 (bad)
- pnpm: 11.21.0
- Node: 24.19.0
- OS: Linux 6.12 (x64)

yangfan-yf-yf:
Fixes #1963.

When Knip runs from a package directory, it now uses the nearest parent `pnpm-workspace.yaml` as the catalog source without treating that file as the current project's workspace definition. Local pnpm, Yarn, and `package.json` catalog sources continue to take precedence. Parent catalog entries validate references, while unused entries outside the current project are not reported.

Regression coverage includes the child-directory invocation, an unrelated unused parent entry, local source precedence, Windows drive roots, and UNC share roots.

Validation:

- `node --test test/dependencies/catalog.*.test.ts test/util/catalog.test.ts test/util/fs.test.ts` (13 passed)
- `pnpm build`
- `pnpm run ci`
- direct CLI checks for cwd, `--directory`, and `--workspace` variants

### DIFF
--- a/packages/knip/fixtures/dependencies/catalog-subdirectory/package.json
+++ b/packages/knip/fixtures/dependencies/catalog-subdirectory/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "@fixtures/catalog-subdirectory",
+  "private": true
+}
--- a/packages/knip/fixtures/dependencies/catalog-subdirectory/pkg/index.ts
+++ b/packages/knip/fixtures/dependencies/catalog-subdirectory/pkg/index.ts
@@ -1,0 +1,3 @@
+import { z } from 'zod';
+
+export const schema = z.string();
--- a/packages/knip/fixtures/dependencies/catalog-subdirectory/pkg/package.json
+++ b/packages/knip/fixtures/dependencies/catalog-subdirectory/pkg/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "@fixtures/catalog-subdirectory__pkg",
+  "dependencies": {
+    "zod": "catalog:"
+  }
+}
--- a/packages/knip/fixtures/dependencies/catalog-subdirectory/pnpm-workspace.yaml
+++ b/packages/knip/fixtures/dependencies/catalog-subdirectory/pnpm-workspace.yaml
@@ -1,0 +1,6 @@
+packages:
+  - 'pkg'
+
+catalog:
+  lodash: 4.17.23
+  zod: 4.4.3
--- a/packages/knip/src/CatalogCounselor.ts
+++ b/packages/knip/src/CatalogCounselor.ts
@@ -7,13 +7,14 @@ import type { Issue } from './types/issues.ts';
 import type { Catalog, Catalogs, WorkspacePackage } from './types/package-json.ts';
 import { type CatalogReference, extractCatalogReferences, parseCatalog } from './util/catalog.ts';
 import type { MainOptions } from './util/create-options.ts';
-import { extname } from './util/path.ts';
+import { dirname, extname, isSameDir } from './util/path.ts';
 import { YamlCatalogPeeker } from './YamlCatalogPeeker.ts';
 
 export type CatalogContainer = { filePath: string; catalog?: Catalog; catalogs?: Catalogs };
 
 export class CatalogCounselor {
   private filePath: string;
+  private isExternal: boolean;
   private entries = new Set<string>();
   private referencedEntries = new Set<string>();
   private referenceIssues: Issue[] = [];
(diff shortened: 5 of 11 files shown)


webpro:
Thanks for the pull request. Running Knip from a workspace dir isn't officially supported. Adding this extra logic seems like overhead for all users (and more to maintain as well). That's why I'm closing this one.

If it's a regression from v6.30.0 as the linked issue states then a solution could be simpler as well? Note that it's not actually a regression as the use case isn't officially supported.

PR #1975, opened 2026-08-26
fix(cli): validate numeric options instead of passing NaN through

devYRPauli:
## Problem

`--max-issues` and `--max-show-issues` are declared as strings in `cli-arguments.ts` and read with `Number()` at three call sites, with no validation in between.

`Number("abc")` is `NaN`, and every comparison against `NaN` is false. So this line never fires when the option is malformed:

```ts
totalErrorCount > Number(args['max-issues'] ?? 0)
```

`knip --max-issues=abc` exits 0 regardless of how many issues were found. The exit-code gate stops gating and reports success.

The realistic way to hit this is a CI variable that holds a non-numeric value. An unset variable gives `--max-issues=`, and `Number('')` is 0, so that case already failed closed.

```
--max-issues="abc"   Number() = NaN   (100 > NaN) = false
--max-issues="10"    Number() = 10    (100 > 10)  = true
```

Negative and fractional values are accepted today too, and neither is meaningful for a maximum.

## Change

One helper in `cli-arguments.ts`, called once in `createOptions`:

```ts
export const parseNumericOption = (value: string | undefined, option: string) => {
  if (value === undefined) return undefined;
  if (!/^\d+$/.test(value)) {
    throw new ConfigurationError(`Option --${option} expects a non-negative integer, got: ${value}`);
  }
  return Number(value);
};
```

`createOptions` returns `maxIssues` and `maxShowIssues`. `cli.ts` reads those values and does not parse. `createOptions` runs before `run()`, so an invalid value throws before the analysis starts.

The regex states the contract directly. It also rejects a whitespace-only value, which `Number('  ') === 0` accepted before.

It throws `ConfigurationError`, so the handler already in `cli.ts` prints the message and the `Run knip --help` hint.

## Tests

`packages/knip/test/util/cli-arguments.test.ts` holds seven unit tests. They cover valid values, an absent value, and rejection of non-numeric, negative, fractional, empty and whitespace-only input. One test asserts the message names the offending option.

`packages/knip/test/cli/cli-max-issues.test.ts` spawns the real CLI against a fixture with `--max-issues=abc` and asserts exit code 2. It uses the `exec` helper, so it runs `dist/cli.js` on Node and `src/cli.ts` on Bun. It sits in `test/cli/`, which the smoke set skips.

Result on Node 24.18.0:

```
node --test test/util/cli-arguments.test.ts test/cli/cli-max-issues.test.ts
pass 8, fail 0

pnpm test --runtime node --smoke
tests 971, pass 969, fail 2
```

The two smoke failures are the moonrepo and Nx plugin tests. They fail the same way on `main` without this branch.

Reverting the regex leaves the whitespace test failing and the other seven passing. `oxlint`, `oxfmt --check`, `check-line-endings.sh` and `tsc` pass.

### DIFF
--- a/packages/knip/src/util/create-options.ts
+++ b/packages/knip/src/util/create-options.ts
@@ -6,7 +6,7 @@ import type { IssueType } from '../types/issues.ts';
 import type { Options } from '../types/options.ts';
 import type { PackageJson } from '../types/package-json.ts';
 import { getCatalogContainer } from './catalog.ts';
-import type { ParsedCLIArgs } from './cli-arguments.ts';
+import { parseNumericOption, type ParsedCLIArgs } from './cli-arguments.ts';
 import { ConfigurationError } from './errors.ts';
 import { findFile, loadJSON } from './fs.ts';
 import {
@@ -193,7 +193,7 @@ export const createOptions = async (options: CreateOptions) => {
     isTreatTagHintsAsErrors: args['treat-tag-hints-as-errors'] ?? parsedConfig.treatTagHintsAsErrors ?? false,
     isUseTscFiles: options.isUseTscFiles ?? args['use-tsconfig-files'] ?? (options.isSession && !configFilePath),
     isWatch: args.watch ?? options.isWatch ?? false,
-    maxShowIssues: args['max-show-issues'] ? Number(args['max-show-issues']) : undefined,
+    maxShowIssues: parseNumericOption(args['max-show-issues'], 'max-show-issues'),
     parsedConfig,
     rules,
     tags,
--- a/packages/knip/src/util/create-options.ts
+++ b/packages/knip/src/util/create-options.ts
@@ -193,6 +193,7 @@ export const createOptions = async (options: CreateOptions) => {
     isTreatTagHintsAsErrors: args['treat-tag-hints-as-errors'] ?? parsedConfig.treatTagHintsAsErrors ?? false,
     isUseTscFiles: options.isUseTscFiles ?? args['use-tsconfig-files'] ?? (options.isSession && !configFilePath),
     isWatch: args.watch ?? options.isWatch ?? false,
+    maxIssues: parseNumericOption(args['max-issues'], 'max-issues') ?? 0,
     maxShowIssues: parseNumericOption(args['max-show-issues'], 'max-show-issues'),
     parsedConfig,
     rules,
--- a/packages/knip/src/cli.ts
+++ b/packages/knip/src/cli.ts
@@ -2,7 +2,7 @@
 import { fix } from './IssueFixer.ts';
 import { run } from './run.ts';
 import type { IssueType, ReporterOptions } from './types/issues.ts';
-import parseArgs, { helpText } from './util/cli-arguments.ts';
+import parseArgs, { helpText, parseNumericOption } from './util/cli-arguments.ts';
 import { createOptions } from './util/create-options.ts';
 import {
   getKnownErrors,
@@ -75,7 +75,7 @@ const main = async () => {
       isShowProgress: options.isShowProgress,
       isTreatConfigHintsAsErrors: options.isTreatConfigHintsAsErrors,
       isTreatTagHintsAsErrors: options.isTreatTagHintsAsErrors,
-      maxShowIssues: args['max-show-issues'] ? Number(args['max-show-issues']) : undefined,
+      maxShowIssues: parseNumericOption(args['max-show-issues'], 'max-show-issues'),
       options: args['reporter-options'] ?? '',
       preprocessorOptions: args['preprocessor-options'] ?? '',
       selectedWorkspaces,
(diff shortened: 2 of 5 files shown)


packages/knip/src/util/create-options.ts
```diff
@@ -193,7 +193,7 @@ export const createOptions = async (options: CreateOptions) => {
     isTreatTagHintsAsErrors: args['treat-tag-hints-as-errors'] ?? parsedConfig.treatTagHintsAsErrors ?? false,
     isUseTscFiles: options.isUseTscFiles ?? args['use-tsconfig-files'] ?? (options.isSession && !configFilePath),
     isWatch: args.watch ?? options.isWatch ?? false,
-    maxShowIssues: args['max-show-issues'] ? Number(args['max-show-issues']) : undefined,
+    maxShowIssues: parseNumericOption(args['max-show-issues'], 'max-show-issues'),
```
webpro, line 197: Let's try to parse/throw both values early in the process and only once, such as in `createOptions`

webpro (commented):
Thanks for the PR! Happy to merge the concept, but I do have a remark.

packages/knip/src/util/create-options.ts
devYRPauli: You are right. The parse ran in three places. Both values now parse once, in `createOptions`.

- `createOptions` returns `maxIssues` and `maxShowIssues`.
- `cli.ts` reads `options.maxIssues` and `options.maxShowIssues`. It no longer calls the parser.

`createOptions` runs before `run()`, so an invalid value throws before the analysis starts. Before this change `knip --max-issues=abc` ran the full analysis and printed the report, then threw:

```
before:  Unused dependencies (1)
         package.json: unused-dep
         ERROR: Option --max-issues expects a non-negative integer, got: abc

after:   ERROR: Option --max-issues expects a non-negative integer, got: abc
```

The exit code stays 2 in both cases. The test file is unchanged and still passes, 7 of 7. `pnpm run ci` and `tsc` pass.

webpro (changes requested):
Thanks, that addresses the remark. Parsing once in `createOptions` means an invalid value now fails before analysis starts, and `cli.ts` just reads numbers. The source side is good.

One thing blocks the merge: the seventh test breaks CI on Node 20.

`test/util/cli-arguments.test.ts` spawns `process.execPath` on `src/cli.ts` with a cleared environment. That works on Node 24 because type stripping is on by default. The `ci-legacy-5.0.4` workflow runs `pnpm test --runtime node --smoke` on Node 20 and 22 across macOS, Ubuntu and Windows, and on Node 20 the child cannot load a `.ts` file:

```
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for packages/knip/src/cli.ts
```

The test then sees exit code 1 and unrelated stderr. Running that file on Node 20 the way CI does gives 6 passing, 1 failing.

`test/helpers/exec.ts` exists for this: it picks `src/cli.ts` on Bun and `dist/cli.js` on Node. Tests that spawn the CLI live in `test/cli/`, which the smoke run deliberately skips. `test/util/` is for unit tests without fixtures, and it is in the smoke set.

Splitting the file along that line fixes it. Keep the six unit tests where they are and move the spawn to `test/cli/cli-max-issues.test.ts`:

```ts
import assert from 'node:assert/strict';
import test from 'node:test';
import { exec } from '../helpers/exec.ts';
import { resolve } from '../helpers/resolve.ts';

test('knip --max-issues=abc (invalid value does not disable the exit code)', () => {
  const cwd = resolve('fixtures/compact-reporter');
  const result = exec('knip --max-issues=abc', { cwd });
  assert.equal(result.stderr, 'ERROR: Option --max-issues expects a non-negative integer, got: abc');
  assert.equal(result.status, 2);
});
```

That passes on Node 20, Node 24 and Bun.

Two smaller notes, neither blocking:

- The `value === ''` guard is there because `Number('')` is 0, but `Number('  ')` is 0 as well, so `--max-issues="  "` still lands on 0. A `/^\d+$/` test covers both and states the contract directly.
- The example in the description does not quite hold. An unset variable yields `--max-issues=`, and since `Number('')` is 0 that already failed closed. The real exposure is a variable holding a non-numeric value, which is still worth closing.

Also worth a rebase on main before the next push.

devYRPauli:
Moved the spawn test to `test/cli/cli-max-issues.test.ts` with your file, unchanged. The unit tests stay in `test/util/cli-arguments.test.ts`, and the now-unused `spawnSync` and `resolve` imports are gone.

```
node --test test/util/cli-arguments.test.ts test/cli/cli-max-issues.test.ts
pass 8, fail 0

pnpm test --runtime node --smoke
tests 971, pass 969, fail 2
```

The smoke run no longer contains the spawn test. The two failures are the moonrepo and Nx plugin tests. They fail the same way on `main` without this branch.

The guard is now `/^\d+$/`. It rejects `"  "`, which `Number('  ') === 0` accepted before. I added a unit test for that value. Reverting the regex leaves that one test failing and the other seven passing. The regex also rejects `1e3`, which returned 1000 before.

Fixed the description. An unset variable gives `--max-issues=`, and `Number('')` is 0, so that case already failed closed. The text now names a non-numeric CI variable as the real exposure.

Rebased on `main`.

webpro:
Thanks Yash!

webpro:

PR #1976, opened 2026-08-26
Add Knip to resolve .mjs, .mts, .cts, .cjs (#1968)

### LINKED ISSUE #1968: 💡 Support automatic discovery of .mjs, .cjs, .mts, and .cts config files
### Description

Knip can load .mjs, .cjs, .mts, and .cts configuration files when they are explicitly passed through --config, but it does not discover them automatically.

Supporting these extensions in the default config lookup would allow users to explicitly choose the module format of their Knip configuration without relying on or changing the project-wide "type" field in package.json.

This is useful for:
- ESM projects that need a CommonJS configuration (.cjs or .cts)
- CommonJS projects that need an ESM configuration (.mjs or .mts)
- TypeScript projects using explicit Node.js module formats
- Tooling consistency, as these extensions are already supported when using --config

For example, a project containing knip.config.mjs should be able to run:
```
knip
```
instead of:
```bash
knip --config knip.config.mjs
```
Since Knip can already load these formats through --config, extending the default discovery locations may be a relatively small, backwards-compatible improvement. The CLI help and configuration documentation should be updated accordingly.

DreamLongYT:
Added
'knip.config.mts',
'knip.config.mjs',
'knip.config.cts',
'knip.config.cjs',
to KNIP_CONFIG_LOCATIONS for it to resolve Feature https://github.com/webpro-nl/knip/issues/1968

I added more Tests and a Documentation too

Fixes #1968

### DIFF
--- a/packages/docs/src/content/docs/reference/cli.md
+++ b/packages/docs/src/content/docs/reference/cli.md
@@ -36,6 +36,10 @@ Use an alternative path for the configuration file. Default locations:
 - `knip.js`
 - `knip.config.ts`
 - `knip.config.js`
+- `knip.config.mts`
+- `knip.config.mjs`
+- `knip.config.cts`
+- `knip.config.cjs`
 - `package.json#knip`
 
 Shortcut: `-c`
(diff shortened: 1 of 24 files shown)


webpro:
Thanks for the pull request. Please see https://github.com/webpro-nl/knip/issues/1968#issuecomment-5435299786, closing this with it.

PR #1977, opened 2026-08-26
💡 Add support for `node:fs/promises` `glob`

### LINKED ISSUE #1974: 💡 Add support for `node:fs/promises` `glob`
### Suggest an idea for Knip

While Knip already supports `import.meta.glob()`, the `node:fs/promises` `glob` remains uncovered.

Example usage:

```ts
const getMigrationFileRelativePaths = async () => {
	const migrationsRoot = path.resolve(process.cwd(), migrationsDirectory);

	const relativePaths = await Array.fromAsync(
		glob(['typescript/*.ts', 'drizzle/*/migration.sql'], {
			cwd: migrationsRoot
		})
	);

	return relativePaths.map((relativePath) => normalizeToPlatformAgnosticPath(relativePath));
};
```

DreamLongYT:
Add support for resolving static glob patterns from `node:fs/promises`.

The Node.js plugin now recognizes named imports, namespace imports, and CommonJS destructuring, including static `cwd` options. Regression coverage and a dedicated fixture are included.

Fixes #1974


packages/knip/test/imports/node-fs-promises-glob.test.ts
```diff
@@ -0,0 +1,14 @@
...
+import { createOptions } from '../helpers/create-options.ts';
+import { resolve } from '../helpers/resolve.ts';
+
+const cwd = resolve('fixtures/imports/node-fs-promises-glob');
+
+test('Resolve node:fs/promises glob patterns as entry files', async () => {
+  const options = await createOptions({ cwd });
+  const { issues } = await main(options);
+
+  assert(!('migrations/one.ts' in issues.files));
```
webpro, line 31: - Please actually add such files to the fixture dir.
- Assert `counters` like other tests do.

packages/knip/src/plugins/node/visitors/fsPromisesGlob.ts
```diff
@@ -0,0 +1,88 @@
...
+      } else if (arg.type === 'ArrayExpression') {
+        patterns = [];
+        for (const element of arg.elements) {
+          if (element && isStringLiteral(element)) patterns.push(getStringValue(element)!);
+        }
+      }
+      if (!patterns?.length) return;
+
+      const cwd = getFirstPropertyValue(node.arguments[1], 'cwd');
+      ctx.addImportGlob(patterns, cwd === undefined ? undefined : { base: cwd });
```
webpro: We should only add patterns if the `cwd` prop is omitted or if value is a literal string (i.e. only if statically readable, not if dynamic). Probably good to do validate this and bail out early.

This will also rule out expressions such as `{ cwd: path.resolve(process.cwd(), ...) }` which we can add later if there is demand. Please add such ignored cases to the fixtures/assertions.

webpro (commented):
Thanks for the pull request! Good idea, but it needs a bit more resilience.

DreamLongYT:
Implemented support for statically analyzable node:fs/promises glob patterns, including named imports, namespace imports, CommonJS destructuring, and literal cwd options. All applicable CI checks pass.

packages/knip/test/imports/node-fs-promises-glob.test.ts
webpro: This isn't resolved, please add `counters` and assert values.

packages/knip/src/plugins/node/visitors/fsPromisesGlob.ts
webpro: This isn't resolved, please add more cases to the fixtures such as:

- `glob(pattern, options)` with a static `options.cwd` that's not in root dir
- `glob(pattern, options)` without a `options.cwd` that's not in root dir
- with options `{ cwd: process.cwd() }`
- etc.
- etc.

webpro (commented):
Thanks for the changes, but please give it another shot.

DreamLongYT:
I hope it should like work now, sry for before, the Bun test failed because of Time Out, but everything else should work I think

webpro:
No worries! Glad you flagged it and fixed it. Are you OK with me wrapping this one up?

DreamLongYT:
Sure, no problem

webpro:
Made some changes and tried to make it slightly more safe/defensive (`fs.glob` is not `tinyglobby` and e.g. negation `!` has different semantics) and more obvious what is (not) supported. If you are okay with this I'm happy to merge.

DreamLongYT:
@webpro just fixed some issues but I think its ready to merge now

webpro:

PR #1978, opened 2026-08-26
Add support for TypeScript content mappers

remcohaszing:
Hello from the future :wave: 

TypeScript 7.1 will add support for content mappers, allowing you to check files other than TypeScript, such as MDX, Vue, Ember, etc.

This change makes Knip understand the the `package` field in the `contentMappers` entries refers to an npm package name.

This feature is still experimental. The configuration might change before the TypeScript 7.1 releases, but I don’t expect any changes.

### DIFF
--- a/packages/knip/src/plugins/typescript/index.ts
+++ b/packages/knip/src/plugins/typescript/index.ts
@@ -30,7 +30,12 @@ const resolveConfig: ResolveConfig<TsConfigJson> = async (localConfig, options) 
       ?.filter(reference => reference.path.endsWith('.json'))
       .map(reference => toConfig('typescript', reference.path, { containingFilePath: options.configFilePath })) ?? [];
 
-  if (!(compilerOptions && localConfig)) return compact([...extend, ...references]);
+  const contentMappers =
+    localConfig.contentMappers?.map(contentMapper =>
+      toConfig('typescript', contentMapper.package, { containingFilePath: options.configFilePath })
+    ) ?? [];
+
+  if (!(compilerOptions && localConfig)) return compact([...contentMappers, ...extend, ...references]);
 
   const jsx = (compilerOptions?.jsxImportSource ? [compilerOptions.jsxImportSource] : []).map(toProductionDependency);
 
@@ -41,6 +46,7 @@ const resolveConfig: ResolveConfig<TsConfigJson> = async (localConfig, options) 
   const importHelpers = compilerOptions?.importHelpers ? ['tslib'] : [];
 
   return compact([
+    ...contentMappers,
     ...extend,
     ...references,
     ...types.map(id => toDeferResolve(id, { isTypeOnly: true, dir: options.cwd })),
--- a/packages/knip/src/plugins/typescript/index.ts
+++ b/packages/knip/src/plugins/typescript/index.ts
@@ -2,7 +2,7 @@ import type { ConfigArg } from '../../types/args.ts';
 import type { IsPluginEnabled, Plugin, ResolveConfig } from '../../types/config.ts';
 import type { TsConfigJson } from '../../types/tsconfig-json.ts';
 import { compact } from '../../util/array.ts';
-import { toConfig, toDeferResolve, toProductionDependency } from '../../util/input.ts';
+import { toConfig, toDeferResolve, toDependency, toProductionDependency } from '../../util/input.ts';
 import { join } from '../../util/path.ts';
 import { hasDependency } from '../../util/plugin.ts';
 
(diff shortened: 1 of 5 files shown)


packages/knip/src/plugins/typescript/index.ts
```diff
@@ -30,7 +30,12 @@ const resolveConfig: ResolveConfig<TsConfigJson> = async (localConfig, options)
       ?.filter(reference => reference.path.endsWith('.json'))
       .map(reference => toConfig('typescript', reference.path, { containingFilePath: options.configFilePath })) ?? [];
 
-  if (!(compilerOptions && localConfig)) return compact([...extend, ...references]);
+  const contentMappers =
+    localConfig.contentMappers?.map(contentMapper =>
+      toConfig('typescript', contentMapper.package, { containingFilePath: options.configFilePath })
```
webpro: [toConfig](https://feat-bulk-suppressions.knip.pages.dev/writing-a-plugin/inputs#toconfig) targets `config` files that Knip has a plugin for. A good example is [angular.json](https://github.com/webpro-nl/knip/blob/c0a9ecf8d1c0cdd7789c1e1bea893706c060239d/packages/knip/fixtures/plugins/angular/angular.json#L24) referencing a `tsconfig.json` file that will then be handled by this TS plugin. (I'll add something like this to the docs to clarify.)

Based on "refers to an npm package name" we need `toDependency` here.

webpro (commented):
Thanks for the PR! Before merge I'll double-check on their progress first, but this is almost good as-is, yet one remark I have.

packages/knip/src/plugins/typescript/index.ts
remcohaszing: I’m not sure what the correct way is to use `toDependency` here. It would be a dev dependency. Also I guess it would be `isTypeOnly` too?

remcohaszing:
I built [`mdx-content-mapper`](https://github.com/remcohaszing/mdx-content-mapper) as a proof of concept and to provide feedback which helped shape the API. Content mappers have already been merged into the TypeScript main branch. I don’t expect the user-facing API to change, but there are no guarantees.

I think it would be nice if users can try this when TypeScript 7.1 (Beta? RC?) releases and Knip won’t complain about them adding content mapper dependencies.

See https://github.com/microsoft/TypeScript/issues/63703 for the release schedule.

From the perspective of a content mapper implementation, it would be nice if Knip understands the `package.json#/typescript/contentMapper/exec` field, which defines a command to run. For example, see https://github.com/remcohaszing/mdx-content-mapper/blob/v0.0.1/package.json#L22-L25. This will of course only affect a tiny number of users, so it might not be worth the effort.

packages/knip/src/plugins/typescript/index.ts
webpro: - There is also [toProductionDependency](https://knip.dev/writing-a-plugin/inputs#toproductiondependency), so `toDependency` is appropriate.
- `isTypeOnly` means the reference is satisfied by types alone. That doesn't seem the intention here, unless I'm missing something? Let's omit if the literal package name should be used (and not the DTS pkg like `@types/mdx-js__content-mapper`).
- `containingFilePath` can also be omitted as `options.configFilePath` is the default.

I'm thinking `toDependency('@mdx-js/content-mapper')` is what we want here.

webpro:
> Content mappers have already been merged into the TypeScript main branch. I don't expect the user-facing API to change, but there are no guarantees.
>
> I think it would be nice if users can try this when TypeScript 7.1 (Beta? RC?) releases and Knip won't complain about them adding content mapper dependencies.

Absolutely, happy to be on the forefront here. Let's make it happen.

webpro:
> From the perspective of a content mapper implementation, it would be nice if Knip understands the package.json#/typescript/contentMapper/exec field, which defines a command to run. For example, see https://github.com/remcohaszing/mdx-content-mapper/blob/v0.0.1/package.json#L22-L25. This will of course only affect a tiny number of users, so it might not be worth the effort.

I'll pick this up in the next day(s) too, seems trivial.

packages/knip/src/plugins/typescript/index.ts
remcohaszing: Ah yes, that makes sense. I thought this might be the case because it’s needed specifically for type checking.

webpro:
FYI, `exec`uted: 74640c03e450624409692f4cc17699775108fd7e (also some docs updated at 2bd367961554ad1addb520bdfad5e31125885969).

webpro:

PR #1981, opened 2026-08-27
fix(webpack): resolve entry descriptors using import field

DreamLongYT:
## Issue

Knip’s Webpack plugin did not correctly handle Webpack entry descriptors that use the standard `import` property.

For example:

```js
module.exports = {
  entry: {
    main: {
      import: './src/main.js',
    },
  },
};
```

Webpack treats `./src/main.js` as the source entry. However, Knip only recognized entry descriptors through their `filename` property. As a result, Knip failed to register `src/main.js` as an entry point and could report it as an unused file, even though Webpack would process it.

## Fix

The fix updates the Webpack plugin’s entry parsing logic. When an entry object contains `import`, Knip now extracts that value and registers it as an entry point. Both a single string and an array of imports are supported:

```js
import: './src/main.js'
```

and:

```js
import: ['./src/main.js', './src/polyfill.js']
```

The existing `filename` handling remains unchanged as a fallback for previously supported entry descriptor formats. Non-string values are filtered out safely instead of being passed into path resolution.


DreamLongYT:
oxfmt failed bc it wanted to change the line endings from CRLF (\r\n) to LF (\n)

webpro:
I think the line endings should be `\n` — any chance you could fix up?

packages/knip/test/plugins/webpack-entry-import.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import { createOptions } from '../helpers/create-options.ts';
+import { resolve } from '../helpers/resolve.ts';
+
+const cwd = resolve('fixtures/plugins/webpack-entry-import');
+
+test('Resolve Webpack entry descriptors using the import field', async () => {
+  const options = await createOptions({ cwd });
+  const { issues, counters } = await main(options);
+
+  assert(!('src/main.js' in issues.files));
```
webpro: This is fine but `counters.files: 0` so a bit redundant.

webpro (commented):
Thanks! Happy to merge after formatting is fixed.

packages/knip/fixtures/plugins/webpack-entry-import/webpack.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  entry: {
+    main: {
+      import: './src/main.js',
```
webpro: Please use a different filename, as this is a default and will always be an entry and thus used.

DreamLongYT:
Everything is fixed i guess

webpro:

PR #1982, opened 2026-08-27
fix(vitest): resolve array entries and root-relative setup dependencies

DreamLongYT:
# fix(vitest): resolve array entries and root-relative setup dependencies

## Issue

Knip’s Vitest plugin did not correctly handle several valid Vitest configuration patterns.

For example, a library entry may be configured with an array value:

```
export default {
  build: {
    lib: {
      entry: {
        main: ['./src/main.ts'],
      },
    },
  },
};
```

Vitest accepts this configuration, but Knip attempted to pass the array directly to `path.join()`, causing a runtime error.

In addition, configuration functions that depend on `ssrBuild` were evaluated only with `ssrBuild: undefined`. As a result, dependencies referenced only by the SSR branch could be reported as unused.

Finally, `test.globalSetup` was resolved relative to the Vitest configuration directory rather than the configured `test.root`. A setup file such as the following could therefore be reported as unresolved or unused:

```
export default {
  test: {
    root: './tests',
    globalSetup: '../src/globalSetup.ts',
  },
};
```

## Fix

The Vitest plugin now normalizes `build.lib.entry` before resolving paths. Single strings, top-level arrays, and named entries containing either strings or arrays are supported safely.

Vitest build configuration functions are now evaluated with `ssrBuild: undefined`, `false`, and `true`, so Knip can discover dependencies from both regular and SSR-specific branches.

`globalSetup` entries are now resolved relative to Vitest’s effective `test.root`, matching Vitest’s path resolution behavior.

The corresponding `ViteConfig` type has also been updated to allow array-valued named library entries.

Regression coverage was added for all three cases in `vitest-regressions.test.ts`, using a dedicated fixture.


webpro (commented):
Thanks for the PR! Much appreciated. Please fix the formatting and I'll merge right away.

packages/knip/test/plugins/vitest-regression.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const cwd = resolve('fixtures/plugins/vitest-regression');
+
+test('Find entries and setup dependencies in Vitest configurations with arrays, SSR branches, and custom roots', async () => {
+  const options = await createOptions({ cwd });
+  const { issues, counters } = await main(options);
+
+  assert.equal(Object.keys(issues.files).length, 0);
+  assert.equal(Object.keys(issues.unresolved).length, 0);
+  assert.equal(Object.keys(issues.unlisted).length, 0);
```
webpro, line 16: Redundant (`counters` assert the same), next time please leave it out for less cognitive overhead.

webpro:

PR #1984, opened 2026-08-28
fix(nuxt): resolve string extends configs

DreamLongYT:
Issue
The Nuxt plugin treated extends as an iterable without normalizing it first. Nuxt accepts both a single string and an array, but a value such as extends: './configs/base' was processed character by character. Knip consequently reported characters such as c, o, n, f and / as dependencies or unresolved paths instead of resolving configs/base.ts.
Fix
The Nuxt plugin now accepts extends as string | string[] and normalizes a single string to an array before resolving each configuration. Existing array-based configurations continue to work unchanged.


webpro (commented):
Thanks for the PR! All good, but please revert unrelated changes.

packages/knip/fixtures/plugins/nuxt-config/nuxt.config.ts
```diff
@@ -1,5 +1,5 @@
 export default defineNuxtConfig({
-  extends: ['~~/some-layer', 'a-nuxt-theme'],
+  extends: './configs/base',
```
webpro: This fixture covered three cases the new value drops: the array form, `~~/` alias resolution to a layer directory, and the package fallback to `toDependency`. With `extends: './configs/base'` neither `some-layer/nuxt.config.ts` nor `a-nuxt-theme` is referenced any more, which is what the new `files: 1` and `dependencies: 1` counters absorb. Could you leave this fixture as it was and add a separate one for the string form?

For that new fixture, note `resolveConfig` globs `nuxt.config.*` inside the extends target, so a file path like `./configs/base` misses and falls through to `toDependency`. A string pointing at a layer directory such as `./some-layer` exercises the resolution path instead. The current fixture shows the difference: `configs/base.ts` sets `css: ['./base.css']` with no such file, yet `unresolved` stays 0 because the extended config is never parsed.

webpro:
The unrelated changes still seem to be there btw.

DreamLongYT:
yep i hope everythings ready now

webpro:
Thank you!

webpro:

Issue #1966, opened 2026-08-25
vi.mock(import(...)) exempts a module's exports from unused-exports detection entirely


mike-4040:
### Prerequisites

- [x] I'm using the latest version (verified on 6.32.0, 6.32.1, 6.32.2)
- [x] I've read the relevant documentation
- [x] I've searched for existing issues
- [x] I've checked the list of known issues
- [x] I've read the issue reproduction guide

### Reproduction

I don't have a hosted sandbox link — here's a minimal inline repro instead (happy to build one if that's required):

```ts
// pool.ts
export function getPool() {
  return {};
}

// dead code — never imported anywhere
export function closePool() {}
```

```ts
// consumer.ts
import { getPool } from "./pool.js";
getPool();
```

```ts
// consumer.test.ts
import { vi, test } from "vitest";

vi.mock(import("./pool.js"), () => ({
  getPool: () => ({}),
}));

test("uses pool", () => {});
```

### Description of the issue

`closePool` has zero importers anywhere in the project and should be reported under "Unused exports." It isn't.

Once any test file mocks a module via `vi.mock(import("./module.js"), factory)`, Knip's Vitest plugin appears to exempt that **entire module** from unused-exports checking — not just the specific keys the factory replaces. In the repro above, the factory only provides `getPool`, but `closePool` still isn't flagged.

Verified this isn't a general regression: a fresh, obviously-dead export added to a file with no `vi.mock` reference is flagged correctly in the same run. It's specifically triggered by the presence of a `vi.mock(import(...))` reference to that file, anywhere in the test suite — independent of what the mock factory actually returns.

This means any module partially mocked this way becomes a permanent blind spot for unused-export detection, which can hide real dead code indefinitely (that's how we found this — a `pg.Pool`-wrapping module with 3 genuinely dead exports, invisible to Knip only because one other export from the same file is mocked in a test).

Happy to provide more detail or test against a real-world file if useful.

webpro:
:rocket: _This issue has been resolved in v6.33.0. See [Release 6.33.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.33.0) for release notes._

_Using Knip in a commercial project? Please consider [becoming a sponsor](https://knip.dev/sponsors)._

mike-4040:
Thanks a lot, Lars.

PR #1985, opened 2026-08-28
fix(vitest): don't exempt all exports via vi.mock(import(...), factory)

IndianCoder3:
## What does this PR do?

When a Vitest test uses `vi.mock(import('./module'), factory)`, only the
exports named in `factory` are actually mocked — the rest keep their real
implementations. Knip was treating this dynamic import as an opaque import,
which marks **every** export of the module as used, so genuinely unused exports
were never reported.

This PR teaches Knip to only keep the exports enumerated in the mock factory
alive, while still keeping the imported file itself "used".

## Example

```ts
// pool.ts
export function getPool() {}
export function closePool() {} // ← unused

// consumer.test.ts
vi.mock(import('./pool.js'), () => ({ getPool: () => ({}) }));
import('./consumer.js');
Before: closePool was silently ignored (module treated as opaque).
After: only getPool is considered used; closePool is correctly reported as
an unused export.
Changes
- packages/knip/src/typescript/visitors/calls.ts — detect
vi.mock(import('...'), factory) and register each key from the factory as a
used import, instead of treating the module as opaque.
- Add fixture packages/knip/fixtures/plugins/vitest-vi-mock-import/ and a
regression test packages/knip/test/plugins/vitest-vi-mock-import.test.ts.
Edge cases covered
- vi.mock(import('./x')) without a factory (auto-mock) — unchanged, still opaque.
- Plain import('./x').then(...) — unchanged, still opaque.
- Arrow, block-body, and object-literal factory forms.
Notes
The README changes are separate and cosmetic only (new badge style, feature
blurb, and an auto-updating contributors grid). Happy to drop them from this PR
if you'd prefer them kept to the bug fix.

### DIFF
--- a/packages/knip/README.md
+++ b/packages/knip/README.md
@@ -10,66 +10,116 @@
 <div align="center">
 
 [![NPM Version][2]][1] [![NPM Downloads][3]][1] [![GitHub Repo stars][5]][4]
+[![License][7]][6] [![Contributors][9]][8] [![PRs welcome][11]][10]
 
 </div>
 
-Knip finds and fixes **unused dependencies, exports and files** in your
-JavaScript and TypeScript projects. Less code and dependencies lead to improved
+**Knip finds and fixes unused dependencies, exports and files in your JavaScript
+and TypeScript projects.** Less code and fewer dependencies lead to improved
 performance, less maintenance and easier refactorings.
 
-- Website: [knip.dev][6]
+## What Knip does
+
+Knip analyzes your project's module graph, configuration and source code to
+surface everything you no longer need — so you can safely delete it:
+
+- **Unused files** — files that are never referenced from anywhere.
+- **Unused and duplicate exports** — dead code that quietly ships in the
+  dependency graph.
+- **Unused and unlisted dependencies** — packages you no longer import, or
+  import but forgot to declare.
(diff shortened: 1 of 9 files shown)


IndianCoder3:
@webpro @DreamLongYT Hey There! Can You Please Check And Merge! Happy to Contribute!

DreamLongYT:
wth i aint a member???

IndianCoder3:
Ah RIP :skull:

Well, I just realized it only shows 100 contributors, let me fix, thanks for the time!

webpro:
Thanks for working on this. I am going to close this because the fix models the mock factory incorrectly. I'll work on a different fix for #1966.

The keys returned by a Vitest mock factory are replacement implementations, not uses of the corresponding real exports. Vitest strips the module-promise `import()` before execution, and missing factory exports throw unless the factory explicitly obtains them through `importOriginal`. Marking every returned key as a real import would therefore preserve false negatives. Additionally, the regression does not exercise that logic because `consumer.ts` already imports `getPool`; the test still passes if the `getMockedExports` loop is removed. The core matcher also accepts any unshadowed-looking `vi.mock` by name without verifying that `vi` came from Vitest.

PR #1988, opened 2026-08-28
fix(vite): resolve nested HTML entry points in multi-page apps

DreamLongYT:
### Issue:
The Vite plugin only inspected the root `index.html` file for module script entries. In Vite multi-page applications, additional pages can be defined in nested directories, such as `nested/index.html`, through `build.rollupOptions.input`. Module scripts referenced by those nested HTML entry points were not added to Knip’s dependency graph and could therefore be incorrectly reported as unused files.

### Fix:
Update the Vite HTML-entry resolver to discover and process all `**/index.html` files within the Vite root, rather than only `index.html` at the root level. Each discovered HTML file is scanned for module scripts, and those scripts are added as production entries.

### Regression test:
Add a Vite multi-page fixture containing an explicit `vite.config.ts` with `build.rollupOptions.input`, a nested `nested/index.html`, and its `nested/nested-entry.ts` module. The regression test verifies that `nested/nested-entry.ts` is not reported as an unused file.


packages/knip/src/plugins/vite/helpers.ts
```diff
@@ -71,8 +72,11 @@ export const getHtmlScriptEntries = async (htmlPath: string): Promise<Input[]> =
   return getModuleScriptSources(html).map(src => toProductionEntry(join(dir, src)));
 };
 
-export const getIndexHtmlEntries = (rootDir: string): Promise<Input[]> =>
-  getHtmlScriptEntries(join(rootDir, 'index.html'));
+export const getIndexHtmlEntries = async (rootDir: string): Promise<Input[]> => {
```
webpro: Blocking: this misses the entries it means to find, because Vite input paths carry any name.

```js
build: { rollupOptions: { input: { main: 'index.html', admin: 'pages/admin.html' } } }
```

`pages/admin.ts` is reported as an unused file, on `main` and on this branch alike. Reading `build.rollupOptions.input` covers that case and the one in your fixture.

packages/knip/src/plugins/vite/helpers.ts
```diff
@@ -71,8 +72,11 @@ export const getHtmlScriptEntries = async (htmlPath: string): Promise<Input[]> =
...
+  const htmlPaths = await _glob({ cwd: rootDir, patterns: ['**/index.html'], gitignore: false });
```
webpro: Blocking: `gitignore: false` means every `index.html` under the workspace counts, build output included. With a `dist/` that is not gitignored:

```
dist/index.html               <script type="module" src="/assets/index-B7xK2p.js">
dist/assets/index-B7xK2p.js   import 'left-pad'
```

`main` reports the bundle as an unused file. This branch turns it into a production entry and reports what it imports:

```
Unlisted dependencies (1)
left-pad  dist/assets/index-B7xK2p.js:1:8
```

packages/knip/fixtures/plugins/vite-mpa/vite.config.ts
```diff
@@ -0,0 +1,9 @@
+export default {
+  build: {
+    rollupOptions: {
```
webpro, line 3: This config is never read, so the fixture does not cover the behavior the description claims. Replace the contents with `export default {}`, or delete the file, and the test still passes: it passes because `nested/index.html` matches the glob, not because it is a declared input. Once the inputs come from the config, this fixture becomes a real test.

packages/knip/test/plugins/vite-mpa.test.ts
```diff
@@ -0,0 +1,13 @@
...
+import { main } from '../../src/index.ts';
+import { createOptions } from '../helpers/create-options.ts';
+import { resolve } from '../helpers/resolve.ts';
+
+const cwd = resolve('fixtures/plugins/vite-mpa');
+
+test('Find module entries from nested Vite multi-page index.html files', async () => {
+  const { issues } = await main(await createOptions({ cwd }));
+
+  assert.equal(issues.files['nested/nested-entry.ts'], undefined);
```
webpro, line 13: Please also assert `counters` against `baseCounters` the way the other plugin tests do (see `test/plugins/vitest9.test.ts`), so a changed file or dependency count cannot slip past.

webpro (changes requested):
Thanks for the PR. The multi-page gap is real and the fixture reproduces it, but I'd like the fix to come from the config rather than from a glob.

Vite declares multi-page entries in `build.rollupOptions.input` (the docs now list it as a deprecated alias of `build.rolldownOptions.input`). This branch never reads it, and globbing `**/index.html` instead is both too narrow and too wide. See the two notes on `helpers.ts`.

What I'd like instead: read `build.rollupOptions.input` in `resolveConfig` (`packages/knip/src/plugins/vitest/index.ts`), resolve each value against the Vite root, and run `.html` values through `getHtmlScriptEntries`. The `electron-vite` plugin already has that shape in `packages/knip/src/plugins/electron-vite/index.ts`, and `ViteConfig['build']` in `packages/knip/src/plugins/vitest/types.ts` needs the `rollupOptions` property. `getIndexHtmlEntries` can stay as it is.

I tried that locally: your fixture passes, a `pages/admin.html` input resolves, the `dist/` case below is untouched, and the rest of the suite stays green.

packages/knip/src/plugins/vite/helpers.ts
```diff
@@ -71,8 +72,11 @@ export const getHtmlScriptEntries = async (htmlPath: string): Promise<Input[]> =
...
+  const entries = await Promise.all(htmlPaths.map(getHtmlScriptEntries));
+  return entries.flat();
+};
```
webpro: Blocking, carried over. This glob is now redundant with the config path, and it suppresses genuine reports: any `index.html` in the tree that Vite never builds turns its module scripts into production entries. Reverting to the single root `index.html` keeps the new fixture green.

packages/knip/src/plugins/vitest/index.ts
```diff
@@ -209,6 +209,16 @@ export const resolveConfig: ResolveConfig<ViteConfigOrFn | VitestWorkspaceConfig
       }
     }
     for (const dependency of findConfigDependencies(cfg, options, vitestRoot)) inputs.add(dependency);
+    const _input = cfg.build?.rollupOptions?.input ?? [];
```
webpro: `build.rollupOptions` is an alias in Vite 8, typed `@deprecated Use rolldownOptions instead`, and Vite normalizes it with `rolldownOptions ??= rollupOptions`. A config written against the current name is missed here: with `build.rolldownOptions.input`, `pages/admin.ts` is still reported as unused.

```ts
const _input = cfg.build?.rolldownOptions?.input ?? cfg.build?.rollupOptions?.input ?? [];
```

`ViteConfig['build']` needs the matching `rolldownOptions` property.

packages/knip/src/plugins/vitest/index.ts
```diff
@@ -209,6 +209,16 @@ export const resolveConfig: ResolveConfig<ViteConfigOrFn | VitestWorkspaceConfig
...
+    const inputSpecifiers =
+      typeof _input === 'string' ? [_input] : Array.isArray(_input) ? _input : Object.values(_input).flat();
+    for (const specifier of inputSpecifiers) {
+      const resolved = toAbsolute(specifier, viteRoot);
+      if (resolved.endsWith('.html')) {
```
webpro, line 218: Non-HTML inputs are dropped, so `input: { worker: 'src/worker.ts' }` still reports `src/worker.ts` as unused, and every value in `input` is a build entry by definition. `electron-vite`'s `resolveConfig` pushes `toProductionEntry(resolved)` for each one and gates the script scan on `.html`; the same `else` branch fits here. This goes past what I asked for last time, so take it or leave it.

webpro (changes requested):
Reading the inputs from the config is the right shape, and the fixture earns its keep now: emptying `vite.config.ts` fails on `pages/admin.ts`.

One item from the last round is still open. `getIndexHtmlEntries` still globs `**/index.html`, and now that the inputs come from the config it finds nothing the config path misses. Revert it to `getHtmlScriptEntries(join(rootDir, 'index.html'))` and drop the `_glob` import: your fixture stays green, because `nested/index.html` and `pages/admin.html` both arrive through `build.rollupOptions.input`. I ran the 16 vite/vitest plugin tests on that shape and they all pass.

Keeping the glob costs real reports. Any `index.html` that is not a declared input silently promotes its module scripts to production entries:

```
examples/index.html   <script type="module" src="./demo.ts">
examples/demo.ts      export const demo = true
```

`main` reports `examples/demo.ts` as an unused file, correctly, because Vite never builds that page. This branch reports nothing. The `dist/` bundle case from the last round is unchanged too.

The reach is also wider than the config path. `src/plugins/vite/index.ts` calls `getIndexHtmlEntries(options.cwd)` from its `resolve` hook, so the glob applies wherever the Vite plugin is enabled, config file or not. Vite's own default input is the single root `index.html`.

Two smaller things in the block you added, both one line, in the notes on `index.ts`. `toAbsolute(specifier, viteRoot)` is the right base: I checked it against a Vite 8 build with `root` set, and relative inputs do resolve against `root`.

packages/knip/src/plugins/vitest/index.ts
```diff
@@ -209,6 +209,17 @@ export const resolveConfig: ResolveConfig<ViteConfigOrFn | VitestWorkspaceConfig
...
     }
     for (const dependency of findConfigDependencies(cfg, options, vitestRoot)) inputs.add(dependency);
+    const _input = cfg.build?.rolldownOptions?.input ?? cfg.build?.rollupOptions?.input ?? [];
+    const inputSpecifiers =
+      typeof _input === 'string' ? [_input] : Array.isArray(_input) ? _input : Object.values(_input).flat();
+    for (const specifier of inputSpecifiers) {
+      const resolved = toAbsolute(specifier, viteRoot);
+      inputs.add(toProductionEntry(resolved));
+      if (resolved.endsWith('.html')) {
+        for (const input of await getHtmlScriptEntries(resolved)) inputs.add(input);
```
webpro, line 219: Blocking: resolve root-absolute module script URLs from the Vite root. `getHtmlScriptEntries` strips the leading slash and joins every source with the HTML file's directory, so a nested entry containing `<script type="module" src="/src/root-relative-entry.js">` is recorded as `nested/src/root-relative-entry.js`. Vite 8.2.2 builds the real `src/root-relative-entry.js`, while Knip reports it as unused. Pass the Vite root into the HTML helper and use it as the base for leading-slash sources; keep `./` and `../` sources relative to the HTML file.

webpro (changes requested):
Nested HTML entry points still misresolve root-absolute module scripts, so this needs one more change.

DreamLongYT:
Sorry for responding late, just came back from something, will be fixing the issues soon

PR #1991, opened 2026-08-29
fix(nuxt): resolve custom shared directories through #shared

DreamLongYT:
**Issue:**
The Nuxt plugin did not resolve the `#shared` alias when Nuxt was configured with a custom shared directory through `dir.shared`. As a result, valid imports such as `#shared/utils/value` were reported as unresolved, and the corresponding files in the custom shared directory were incorrectly reported as unused.

**Fix:**
Register the configured Nuxt shared directory as the `#shared` alias and add a wildcard alias for subpath imports:

```ts
const sharedDir = toAbsolute(localConfig.dir?.shared ?? 'shared', cwd);
inputs.push(toAlias('#shared', sharedDir));
inputs.push(toAlias('#shared/*', join(sharedDir, '*'), { dir: cwd }));
```

**Regression tests:**
Add a Nuxt fixture with `dir.shared: 'shared-custom'`, a file at `shared-custom/utils/value.ts`, and a server handler importing it through `#shared/utils/value`. The regression test verifies that the import is no longer reported as unresolved and that the shared file is no longer reported as unused.


webpro:
Thank you! 👍

webpro:

PR #1995, opened 2026-08-30
docs: add unused dependencies/exports pages to sidebar

andriyor:
`typescript/unused-dependencies` and `typescript/unused-exports` were not in the left navigation. They were only reachable through inline links in prose (homepage list, `why-use-knip`), so readers who orient by the sidebar never found them.

- Nest both under `FAQ` in the Troubleshooting group (same pattern as `Writing A Plugin`)
- Drop `prev: false` / `next: false` from their frontmatter so prev/next chains within the group

`pnpm astro build` passes.

### DIFF
--- a/packages/docs/astro.config.ts
+++ b/packages/docs/astro.config.ts
@@ -86,7 +86,10 @@ export default defineConfig({
             'guides/handling-issues',
             'reference/known-issues',
             'features/auto-fix',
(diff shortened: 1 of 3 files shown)


webpro:
Thanks for the PR. They pages exist for SEO purposes only and are deliberately hidden for humans.

PR #1997, opened 2026-08-31
fix: compile files with workspace-scoped compilers

### LINKED ISSUE #1930: 🐛 Knip compiler registration process is not workspace-aware, leading to false positives in monorepos
### Prerequisites

### Reproduction url

https://github.com/caseycarroll/knip-marko-monorepo-repro

### Reproduction access

### Description of the issue

When Knip registers plugin compilers, it stores them project-wide by file extension without retaining their workspace scope.

The Marko plugin correctly builds a tag-to-dependency map for each workspace using that workspace’s package.json and module-resolution path. However, Knip retains only the first compiler registered for .marko; later workspace-specific compilers are discarded.

Because Knip processes the root workspace first, when the root enables Marko, every .marko file is compiled using the root-derived tag map. Tags provided only by child-workspace dependencies are therefore not converted into synthetic imports, causing those dependencies to be reported as unused.

A workspace-aware compiler registry and lookup could resemble: 
```
.marko
  ├── repository root → rootMarkoCompiler
  ├── brandhub-ui     → brandhubMarkoCompiler
  └── map-view        → mapViewMarkoCompiler
```

When compiling a file, Knip would select the compiler belonging to the workspace that owns that file.
This is not necessarily limited to Marko. Other plugins that construct workspace-specific compilers—such as Nuxt, whose compiler captures workspace-specific generated auto-import metadata—may be affected by the same project-wide compiler registry. Stateless compilers are not affected.

https://github.com/webpro-nl/knip/blob/f21bcbb653377682dd690fb99b0f02977ac524d2/packages/knip/src/ProjectPrincipal.ts#L82-L95

gioboa:
Closes #1930

## Problem

Plugin-registered compilers are stored project-wide by file extension, keeping only the first one registered. Some plugins build workspace-specific compilers — Marko embeds a tag→dependency map derived from that workspace's `package.json`, Nuxt embeds auto-import maps from its `.nuxt` dir. Since the root workspace is processed first, every `.marko`/`.vue` file in a monorepo was compiled with the root-derived compiler, so tags/auto-imports provided only by child-workspace dependencies were not converted into synthetic imports → false positives (unused dependencies, unused files/exports).

## Fix

Files are now compiled with the compiler registered by the workspace that owns them:

- `ProjectPrincipal` records a dir-scoped entry for every registration; lookup picks the longest matching dir prefix, falling back to the existing global per-extension map (first registrant) for files outside any registering workspace.
- Recording every registration (not just conflicting ones) keeps resolution order-independent: workspaces are processed nested-child-before-ancestor, so an ancestor registering later must not shadow the child's own files (covered by the `marko-workspace-nested` fixture).
- The global maps are untouched for extension bookkeeping (`has`/`keys`/`size` call sites unchanged), and async compilers are untouched — plugins can only register sync compilers and config-level async compilers are global. Single-workspace projects and stateless built-in/config compilers take the exact same path as before.

The second commit applies this to Nuxt layers (the case confirmed in the issue comments): `registerCompilers` now also registers when a workspace has generated `.nuxt` definition files, even if its manifest is a dependency-less stub (the layers convention), so each layer's files are compiled with its own auto-import maps. A layer without its own `.nuxt` still falls back to the root compiler. Happy to drop this commit if the confirmer prefers to PR the plugin side separately.

## Tests

Three fixtures + tests, each verified failing before the fix and passing after:

- `compilers/marko-workspace`: root + child both enable Marko; only the child depends on the taglib package (mirrors the issue reproduction)
- `compilers/marko-workspace-nested`: nested workspace registers first; ensures registration order doesn't matter
- `plugins/nuxt-layers`: layer stub workspace with its own `.nuxt`; auto-imported composable/component no longer reported, genuinely unused component still is

Verified against the issue's reproduction repo (`knip-marko-monorepo-repro`): exits clean.


packages/knip/src/ProjectPrincipal.ts
```diff
@@ -95,6 +101,12 @@ export class ProjectPrincipal {
     }
   }
 
+  getSyncCompiler(ext: string, filePath: string) {
```
webpro: `ConfigurationChief.findWorkspaceByFilePath` already answers "which workspace owns this file". It matches `availableWorkspaceDirs` deepest-first and memoizes per file path. Reuse it instead of building a second prefix index per extension, and key the scoped map by workspace name:

```ts
private scopedSyncCompilers = new Map<string, Map<string, CompilerSync>>();
```

`run.ts` can pass the lookup into the constructor:

```ts
const principal = new ProjectPrincipal(
  options,
  toSourceFilePath,
  findWorkspacePackageTarget,
  filePath => chief.findWorkspaceByFilePath(filePath)?.name
);
```

That also drops the `sort` on line 93. A name-keyed `Map` is order-independent by construction, so the `marko-workspace-nested` guarantee holds without maintaining sorted prefixes.

packages/knip/src/ProjectPrincipal.ts
```diff
@@ -75,17 +76,22 @@ export class ProjectPrincipal {
...
     this.fileManager = new SourceFileManager({
-      compilers: [this.syncCompilers, this.asyncCompilers],
+      asyncCompilers: this.asyncCompilers,
+      getSyncCompiler: (ext, filePath) => this.getSyncCompiler(ext, filePath),
     });
     this.walkAndAnalyze = timerify(this.walkAndAnalyze.bind(this), 'walkAndAnalyze');
   }
 
-  addCompilers(compilers: [SyncCompilers, AsyncCompilers]) {
+  addCompilers(compilers: [SyncCompilers, AsyncCompilers], dir: string) {
```
webpro: Take the workspace name rather than its dir (`build.ts` has it as `name`), and store a dispatcher in `syncCompilers` itself. Then every existing reader of that map keeps working:

```ts
addCompilers(workspaceName: string, compilers: Compilers) {
  for (const [ext, compiler] of compilers[0]) {
    const workspaceCompilers = this.scopedSyncCompilers.get(ext);
    if (workspaceCompilers) {
      workspaceCompilers.set(workspaceName, compiler);
    } else {
      const workspaceCompilers = new Map([[workspaceName, compiler]]);
      this.scopedSyncCompilers.set(ext, workspaceCompilers);
      this.syncCompilers.set(ext, (source, filePath) => {
        const owner = this.findWorkspaceNameByFilePath(filePath);
        return ((owner ? workspaceCompilers.get(owner) : undefined) ?? compiler)(source, filePath);
      });
      this.extensions.add(ext);
    }
  }
}
```

The first registrant stays the fallback for files outside any workspace that registered one, so nothing changes for what the scoped map does not cover.

packages/knip/src/ProjectPrincipal.ts
```diff
@@ -75,17 +76,22 @@ export class ProjectPrincipal {
...
       if (!this.syncCompilers.has(ext)) {
         this.syncCompilers.set(ext, compiler);
         this.extensions.add(ext);
       }
+      const scoped = this.scopedSyncCompilers.get(ext) ?? [];
+      scoped.push([`${dir}/`, compiler]);
+      scoped.sort((a, b) => b[0].length - a[0].length);
+      this.scopedSyncCompilers.set(ext, scoped);
     }
     for (const [ext, compiler] of compilers[1]) {
```
webpro, line 103: You are right that nothing registers a workspace-specific async compiler today: `registerCompiler` only writes to `compilers[0]`, and `asyncCompilers` comes off the root config. Please scope this loop the same way anyway. I would rather the two halves of `addCompilers` stay one shape than have the async half quietly keep the old first-wins behavior once something does register per workspace.

packages/knip/src/typescript/SourceFileManager.ts
```diff
@@ -1,28 +1,29 @@
...
 export class SourceFileManager {
   sourceTextCache = new Map<string, string>();
-  syncCompilers: SyncCompilers;
   asyncCompilers: AsyncCompilers;
+  getSyncCompiler: SourceFileManagerOptions['getSyncCompiler'];
 
-  constructor({ compilers }: SourceFileManagerOptions) {
-    this.syncCompilers = compilers[0];
-    this.asyncCompilers = compilers[1];
+  constructor({ asyncCompilers, getSyncCompiler }: SourceFileManagerOptions) {
```
webpro: With the dispatcher in `syncCompilers`, this file can go back to untouched. Compiler selection belongs to `ProjectPrincipal` rather than split across the two.

packages/knip/src/plugins/nuxt/index.ts
```diff
@@ -113,20 +113,21 @@ const addAppEntries = (inputs: Input[], srcDir: string, serverDir: string, confi
...
+const definitionFiles = [
+  '.nuxt/imports.d.ts',
+  '.nuxt/components.d.ts',
+  '.nuxt/types/nitro-routes.d.ts',
+  '.nuxt/types/nitro-imports.d.ts',
+];
+
 const registerCompilers: RegisterCompilers = async ({ cwd, hasDependency, registerCompiler }) => {
-  if (hasDependency('nuxt') || hasDependency('nuxt-nightly')) {
+  const paths = definitionFiles.map(file => join(cwd, file));
```
webpro: The Nuxt plugin is not root-only, so `registerCompilers` runs for every workspace of every project. For a workspace with no `.nuxt` at all, which is most of them, this stats four paths that cannot exist. One check on the directory covers it, and `||` short-circuits it away for workspaces that do have the dependency:

```ts
if (hasDependency('nuxt') || hasDependency('nuxt-nightly') || isDirectory(cwd, '.nuxt')) {
```

Then `paths` can move back inside the branch. Hoisting `definitionFiles` to module scope is a good change, please keep that.

packages/knip/fixtures/compilers/marko-workspace/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@fixtures/compilers-marko-workspace",
+  "workspaces": ["packages/*"],
+  "devDependencies": {
```
webpro, line 7: Both workspaces here register a compiler, so nothing covers the workspace that registers none, which is the path every non-Marko workspace in a real repo takes. Could you add a `packages/fallback` without the `marko` devDependency, plus a taglib on the root that only that workspace references? Give the root `index.marko` a plain `<div>`, so the tag in `packages/fallback` is the only thing keeping the root dependency alive and the fixture fails if that file stops being compiled.

webpro (changes requested):
The diagnosis is right and the fixtures are exactly the shape I want: `marko-workspace` mirrors the reported setup, `marko-workspace-nested` pins that registration order does not matter, and `nuxt-layers` covers the case confirmed in the issue thread. A few requests below, mostly about reusing what is already there.

Please keep the Nuxt commit. Without it the scoping fix does nothing for a layers repo, since no layer ever registers a compiler.

packages/knip/src/ProjectPrincipal.ts
gioboa: Fix this and the other comments with [811ba02](https://github.com/webpro-nl/knip/pull/1997/commits/811ba022927449a1165fd0d6db404926f26b90fc)

packages/knip/src/ProjectPrincipal.ts
```diff
@@ -67,29 +70,45 @@ export class ProjectPrincipal {
   constructor(
     options: MainOptions,
     toSourceFilePath: ToSourceFilePath,
-    findWorkspacePackageTarget?: WorkspacePackageTargetHandler
+    findWorkspacePackageTarget?: WorkspacePackageTargetHandler,
+    findWorkspaceNameByFilePath?: (filePath: string) => string | undefined
```
webpro: `run.ts` is the only caller, so leaving these optional buys nothing and `new ProjectPrincipal(options, toSourceFilePath)` still type-checks. That builds a principal where every file resolves to no owner and falls back to the first registrant, which is the behavior this PR removes. Make both required and drop the default on line 79:

```ts
findWorkspacePackageTarget: WorkspacePackageTargetHandler | undefined,
findWorkspaceNameByFilePath: (filePath: string) => string | undefined
```

packages/knip/src/ProjectPrincipal.ts
```diff
@@ -67,29 +70,45 @@ export class ProjectPrincipal {
...
     this.cache = new CacheConsultant('root', options);
     this.toSourceFilePath = toSourceFilePath;
     this.findWorkspacePackageTarget = findWorkspacePackageTarget;
+    this.findWorkspaceNameByFilePath = findWorkspaceNameByFilePath ?? (() => undefined);
     this.tsConfigFile = options.tsConfigFile ? toAbsolute(options.tsConfigFile, options.cwd) : undefined;
     this.pluginVisitorObjects.push(createBunShellVisitor(this.pluginCtx));
-    this.fileManager = new SourceFileManager({
-      compilers: [this.syncCompilers, this.asyncCompilers],
-    });
+    this.fileManager = new SourceFileManager({ compilers: [this.syncCompilers, this.asyncCompilers] });
```
webpro: Unrelated reformat. oxfmt takes it either way, so please restore the three-line form from `main` and keep the diff to the change itself.

webpro (changes requested):
Everything from the last round landed, thanks for the quick turnaround. The fixture is right too: with the root `index.marko` back to a plain `<div>`, the only thing keeping `@orchard/basket-tags` alive is the workspace that registers no compiler, so the fallback is genuinely covered. One more on the constructor, plus a nit.

packages/knip/src/ProjectPrincipal.ts
gioboa: Fixed with [c8eca06](https://github.com/webpro-nl/knip/pull/1997/commits/c8eca0605822bc9b5e9dfc987c7e63dd7504e93b)

packages/knip/src/ProjectPrincipal.ts
gioboa: Fixed with [c8eca06](https://github.com/webpro-nl/knip/pull/1997/commits/c8eca0605822bc9b5e9dfc987c7e63dd7504e93b)

webpro:
Thank you, Giorgio! What an entry ✨👌

webpro:

PR #1999, opened 2026-09-01
fix: track pnpm workspace override catalogs

### LINKED ISSUE #1998: 🐛 Doesn't detect overrides from pnpm-workspace.yaml
### Prerequisites

### Reproduction url

https://github.com/returnsvoidjanet/knip-pnpm-catalog-overrides-sscce/tree/main/repro <-- IDK Janet overengineered it and didn't give me access, ignore the rest of the repo lol, this dir seems S&SC

### Reproduction access

### Description of the issue

pnpm allows using catalogs in overrides; glancing at the src Knip seems to support pnpm.overrides from package.json but not from the `pnpm-workspace.yaml`.

gioboa:
Closes #1998

Recognize catalog references in pnpm-workspace.yaml overrides, preventing used entries from being reported as unused.

### Before

Catalog entries used by workspace overrides were reported unused.

###  After

Override references are recognized correctly.

### DIFF
--- a/packages/knip/fixtures/dependencies/catalog-pnpm/pnpm-workspace.yaml
+++ b/packages/knip/fixtures/dependencies/catalog-pnpm/pnpm-workspace.yaml
@@ -5,3 +5,13 @@ catalog:
   react: ^18.0.0
   typescript: ^5.0.0
   lodash: ^4.17.21
+  left-pad: ^1.3.0
+  ms: ^2.1.3
+  bar: ^1.0.0
+  kleur: ^4.1.5
+
+overrides:
+  left-pad@^1: 'catalog:'
+  debug>ms: 'catalog:'
+  foo>bar@>=1: 'catalog:'
+  foo@3 || >=2>kleur: 'catalog:'
--- a/packages/knip/fixtures/dependencies/catalog-references/packages/app/package.json
+++ b/packages/knip/fixtures/dependencies/catalog-references/packages/app/package.json
@@ -3,5 +3,10 @@
   "dependencies": {
     "lodash": "catalog:",
     "express": "catalog:backend"
+  },
+  "pnpm": {
+    "overrides": {
+      "debug>react": "catalog:"
+    }
   }
 }
--- a/packages/knip/src/CatalogCounselor.ts
+++ b/packages/knip/src/CatalogCounselor.ts
@@ -5,23 +5,32 @@ import { PackagePeeker } from './PackagePeeker.ts';
 import type { Fixes } from './types/exports.ts';
 import type { Issue } from './types/issues.ts';
 import type { Catalog, Catalogs, WorkspacePackage } from './types/package-json.ts';
-import { type CatalogReference, extractCatalogReferences, parseCatalog } from './util/catalog.ts';
+import {
+  type CatalogReference,
+  extractCatalogReferences,
+  getOverrideCatalogReferences,
+  type OverrideCatalogReference,
+  parseCatalog,
+} from './util/catalog.ts';
 import type { MainOptions } from './util/create-options.ts';
 import { extname } from './util/path.ts';
 import { YamlCatalogPeeker } from './YamlCatalogPeeker.ts';
 
-export type CatalogContainer = { filePath: string; catalog?: Catalog; catalogs?: Catalogs };
+export type CatalogContainer = { filePath: string; catalog?: Catalog; catalogs?: Catalogs; overrides?: Catalog };
 
 export class CatalogCounselor {
   private filePath: string;
   private entries = new Set<string>();
   private referencedEntries = new Set<string>();
   private referenceIssues: Issue[] = [];
+  private overrideReferences: OverrideCatalogReference[] = [];
   private fileContent?: string;
 
   constructor(options: MainOptions) {
     this.filePath = options.catalog.filePath;
     this.entries = parseCatalog(options.catalog);
+    this.overrideReferences = getOverrideCatalogReferences(options.catalog.overrides);
+    for (const reference of this.overrideReferences) this.addReference(reference);
   }
 
   private addReferencedCatalogEntry(entryName: string) {
@@ -39,14 +48,17 @@ export class CatalogCounselor {
     manifestStr,
   }: Pick<WorkspacePackage, 'name' | 'manifest' | 'manifestPath' | 'manifestStr'>) {
     const catalogReferences = extractCatalogReferences(manifest);
+    if (workspace === ROOT_WORKSPACE_NAME) {
+      for (const reference of getOverrideCatalogReferences(manifest.pnpm?.overrides)) catalogReferences.push(reference);
+    }
     if (catalogReferences.length === 0) return;
 
     const peeker = new PackagePeeker(manifestStr);
-    for (const { catalogName, packageName } of catalogReferences) {
+    for (const { catalogName, packageName, selector } of catalogReferences) {
       const catalogEntryName = `${catalogName}:${packageName}`;
       this.addReferencedCatalogEntry(catalogEntryName);
       if (!this.entries.has(catalogEntryName)) {
-        const pos = peeker.getCatalogReferenceLocation(packageName, catalogName);
+        const pos = peeker.getCatalogReferenceLocation(selector ?? packageName, catalogName);
         this.referenceIssues.push({
           type: 'catalogReferences',
           filePath,
@@ -64,12 +76,13 @@ export class CatalogCounselor {
     const filePath = this.filePath;
     const workspace = ROOT_WORKSPACE_NAME;
     const catalogIssues: Issue[] = [];
+    const overrideReferenceIssues: Issue[] = [];
 
-    if (this.entries.size > 0) {
+    if (this.entries.size > 0 || this.overrideReferences.length > 0) {
       this.fileContent = await readFile(filePath, 'utf-8');
       const isYaml = ['.yml', '.yaml'].includes(extname(filePath));
-      const Peeker = isYaml ? YamlCatalogPeeker : JsonCatalogPeeker;
-      const peeker = new Peeker(this.fileContent);
+      const yamlPeeker = isYaml ? new YamlCatalogPeeker(this.fileContent) : undefined;
+      const peeker = yamlPeeker ?? new JsonCatalogPeeker(this.fileContent);
 
       for (const entry of this.entries.keys()) {
         if (!this.referencedEntries.has(entry)) {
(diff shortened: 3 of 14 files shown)


packages/knip/src/CatalogCounselor.ts
```diff
@@ -22,6 +22,9 @@ export class CatalogCounselor {
   constructor(options: MainOptions) {
     this.filePath = options.catalog.filePath;
     this.entries = parseCatalog(options.catalog);
+    for (const reference of extractCatalogReferences({ pnpm: { overrides: options.catalog.overrides } })) {
```
webpro: P1 Override selector keys are still missed here.

pnpm accepts version and nested selectors as override keys and resolves `catalog:` for all of them. Confirmed with `pnpm install --lockfile-only` on 11.24.0:

```yaml
catalog:
  left-pad: 1.3.0
  ms: 2.1.3

overrides:
  left-pad@^1.0.0: 'catalog:'  # lockfile: left-pad@^1.0.0: 1.3.0
  debug>ms: 'catalog:'         # lockfile: debug>ms: 2.1.3
  debug>ms@>=2: 'catalog:'     # lockfile: debug>ms@>=2: 2.1.3
```

The reference is keyed on the raw override key, so `left-pad@^1.0.0` never matches the `left-pad` entry. All three are still reported unused on this branch, and `--fix` still removes them, reproducing the same `ERR_PNPM_CATALOG_IN_OVERRIDES` breakage this PR sets out to prevent. The gap predates the PR for `package.json#pnpm.overrides`, where it is worse: the selector key also produces a bogus "unresolved catalog reference".

Since overrides are this PR's subject, settling the key grammar here beats leaving a second syntax with the same blast radius.

Splitting on the last `>`, or on `>` generally, gets this wrong, because ranges contain `>` too. pnpm splits at the first `>` not preceded by a space, `|`, or `@`, then takes the package name off the target half ([`parse-overrides`](https://github.com/pnpm/pnpm/blob/main/pnpm11/config/parse-overrides/src/index.ts): `DELIMITER_REGEX = /[^ |@]>/`, then `alias: result.targetPkg.name`). `getPackageNameFromModuleSpecifier` already handles that second half. Together they resolve `debug>ms@>=2` to `ms` and `foo@3 || >=2>bar` to `bar`.

packages/knip/src/CatalogCounselor.ts
```diff
@@ -22,6 +22,9 @@ export class CatalogCounselor {
...
+      this.addReference(reference);
```
webpro: P2 Unresolved references in `pnpm-workspace.yaml` are silently accepted.

`addReference` skips the entry check that `addWorkspace` applies, so an override pointing at a missing entry goes unreported:

```yaml
overrides:
  no-such-pkg: 'catalog:'   # ERR_PNPM_CATALOG_IN_OVERRIDES
```

That is a hard failure in pnpm, and Knip reports exactly this for `package.json#pnpm.overrides` as a `catalogReferences` issue. So the authoritative location stays quiet while the deprecated one reports.

Reporting it needs a position lookup for the `overrides` block, and `YamlCatalogPeeker` indexes only `catalog:` and `catalogs:`, so deferring is defensible. Worth making that an explicit decision rather than a side effect.

packages/knip/src/util/catalog.ts
```diff
@@ -77,7 +77,7 @@ export const parseCatalog = (container: CatalogContainer) => {
   return entries;
 };
 
-export const extractCatalogReferences = (manifest: PackageJson): CatalogReference[] => {
+export const extractCatalogReferences = (manifest: Partial<PackageJson>): CatalogReference[] => {
```
webpro: P2 The synthetic manifest at the call site widens this signature.

`extractCatalogReferences({ pnpm: { overrides: options.catalog.overrides } })` builds an object literal only to fit the parameter type, and that is what forces `Partial<PackageJson>` here. Keeping the original signature with the literal in place fails:

```
TS2345: Argument of type '{ pnpm: { overrides: Catalog | undefined; }; }' is not assignable to parameter of type 'PackageJson'. Type ... is missing the following properties from type 'PluginConfig<PluginMap>': angular, astro, "astro-db", "astro-markdoc", and 177 more.
```

Every caller now accepts a partial manifest to accommodate one fake one. Lifting the inner loop into a shared helper and exporting a dedicated override entry point keeps `extractCatalogReferences(manifest: PackageJson)` intact and reads better at the call site:

```ts
for (const reference of getOverrideCatalogReferences(options.catalog.overrides)) {
  this.addReference(reference);
}
```

I have that built together with the selector handling above; happy to hand it over as a patch.

packages/knip/fixtures/dependencies/catalog-pnpm/pnpm-workspace.yaml
```diff
@@ -5,3 +5,7 @@ catalog:
   react: ^18.0.0
   typescript: ^5.0.0
   lodash: ^4.17.21
+  left-pad: ^1.3.0
+
+overrides:
+  left-pad: 'catalog:'
```
webpro: P3 The fixture change is asserted only through the counter.

`catalog: 1` in `test/dependencies/catalog.pnpm.test.ts` does catch the regression, but the intent is easier to read directly:

```ts
assert(!issues.catalog['pnpm-workspace.yaml']?.['default.left-pad']);
```

`overrides: { pkg: 'catalog:frontend' }` already works on this branch; a case in `catalog-named` would pin it.

webpro (commented):
Thanks @gioboa. Right fix, right place, and it holds up: verified against the reproduction from #1998 and against pnpm 11.24.0 itself. Full `pnpm qa` passes on the branch, and the fixture is a real guard, failing on `main` with `catalog: 2` against the expected `catalog: 1`.

Two things make this worth more than the description suggests.

It is not only a false positive. On `main`, `knip --fix --fix-type catalog` deletes the entry and leaves the workspace uninstallable:

```
[ERR_PNPM_CATALOG_IN_OVERRIDES] Could not resolve a catalog in the overrides: No catalog entry 'left-pad' was found for catalog 'default'.
```

And `pnpm-workspace.yaml` is no longer one location among several. Since pnpm 11.0.0 the `pnpm` field in `package.json` is not read at all ([#10086](https://github.com/pnpm/pnpm/pull/10086), [docs](https://pnpm.io/package_json)):

```
[WARN] The "pnpm" field in package.json is no longer read by pnpm. The following keys were ignored: "pnpm.overrides". See https://pnpm.io/settings for the new home of each setting.
```

The override is dropped, and the lockfile records none. So before this PR, Knip read overrides only from the location pnpm has stopped reading. Both are still worth supporting, since pnpm 10 does read `pnpm.overrides`.

Scope checks out as complete. `overrides` is the only field in `pnpm-workspace.yaml` that consumes catalog entries. `packageExtensions` passes `catalog:` straight through into the lockfile unresolved, `peerDependencyRules.allowedVersions` rejects it with `ERR_PNPM_INVALID_ALLOWED_VERSION_SELECTOR` even when the entry exists, `patchedDependencies` with `ERR_PNPM_PATCH_NON_SEMVER_RANGE`, and top-level `resolutions` is not a pnpm setting at all.

Four inline notes below. Only the first affects behavior.

One more that has no line to attach to: `packages/docs/src/content/docs/features/catalogs.md` says references are resolved from `dependencies`, `devDependencies`, `peerDependencies`, `optionalDependencies`, `resolutions` and `pnpm.overrides`. The `pnpm-workspace.yaml` `overrides` key belongs there, and given pnpm 11 that line reads backwards as it stands.

packages/knip/src/CatalogCounselor.ts
gioboa: Thanks, fixed with  [f2c0087](https://github.com/webpro-nl/knip/pull/1999/commits/f2c00876c7194fa2f7c878c13dac0452a6aca4f2)

packages/knip/src/util/catalog.ts
```diff
@@ -77,26 +80,41 @@ export const parseCatalog = (container: CatalogContainer) => {
...
-  checkDependencies(manifest.peerDependencies);
-  checkDependencies(manifest.optionalDependencies);
-  checkDependencies(manifest.resolutions);
-  checkDependencies(manifest.pnpm?.overrides);
+  collectCatalogReferences(manifest.dependencies, catalogReferences);
+  collectCatalogReferences(manifest.devDependencies, catalogReferences);
+  collectCatalogReferences(manifest.peerDependencies, catalogReferences);
+  collectCatalogReferences(manifest.optionalDependencies, catalogReferences);
+  collectCatalogReferences(manifest.resolutions, catalogReferences);
+  collectCatalogReferences(manifest.pnpm?.overrides, catalogReferences, true);
```
webpro: P2 `pnpm.overrides` is read from every workspace manifest, but pnpm honors it only in the root.

pnpm 10 writes no lockfile overrides for a nested one, and says so. pnpm 11 ignores the field everywhere.

```
packages/app | WARN  The field "pnpm.overrides" was found in packages/app/package.json.
This will not take effect. You should configure "pnpm.overrides" at the root of the
workspace instead.
```

Now that selectors resolve, a nested override marks a catalog entry as referenced and hides a real report. With catalog entry `ms` in the root `pnpm-workspace.yaml`, and `packages/app/package.json` carrying `"pnpm": {"overrides": {"debug>ms": "catalog:"}}`, `main` reports `ms` unused alongside a bogus `default.debug>ms` unresolved reference, and this branch reports neither.

The plain-key form (`"ms": "catalog:"`) already hid it on `main`, so the gap predates this PR and selector support widens it. It stays inside config no pnpm version reads, so nothing is blocked, but overrides are this PR's subject and this is a good place to close it.

Gating on the root workspace covers it, and drops the `isOverride` flag on the way, since `extractCatalogReferences` no longer owns overrides:

```ts
const catalogReferences = extractCatalogReferences(manifest);
if (workspace === ROOT_WORKSPACE_NAME) {
  for (const reference of getOverrideCatalogReferences(manifest.pnpm?.overrides)) catalogReferences.push(reference);
}
```

webpro (commented):
All five points from the last round are addressed. `pnpm qa` is green on f2c0087, and every changed assertion fails on `main` for the right reason, so the fixtures are real guards. The selector grammar matches `@pnpm/parse-overrides` exactly, and pnpm 11.24.0 resolves each nested key in the fixture.

One loose end, and no blocker. `pnpm.overrides` is read from every workspace manifest, and pnpm honors it only in the root. The gap predates this PR, but selector support makes it reachable through one more syntax, and a new fixture case pins the behavior. Nice to wrap up here while overrides are the subject. Details inline.

packages/knip/fixtures/dependencies/catalog-references/packages/app/package.json
```diff
@@ -3,5 +3,10 @@
   "dependencies": {
     "lodash": "catalog:",
     "express": "catalog:backend"
+  },
+  "pnpm": {
+    "overrides": {
+      "missing-package-json@^1": "catalog:"
```
webpro: P3 This pins behavior pnpm has in no version. `pnpm.overrides` outside the root manifest is ignored, so the case expects an unresolved reference for a field nothing reads.

Moving the block to the fixture's root `package.json` makes it a real scenario, and the `{ line: 9, col: 8 }` assertion holds unchanged there.

packages/knip/src/util/catalog.ts
gioboa: Fixed with [b92f2b9](https://github.com/webpro-nl/knip/pull/1999/commits/b92f2b9804ef8b29c7ca3e8e181c29ce869182f8)

packages/knip/fixtures/dependencies/catalog-references/packages/app/package.json
gioboa: Fixed

webpro:
Thanks for the pull request! Love to see such gaps being filled up properly 🙏

webpro:

PR #2000, opened 2026-09-02
Add Varlock plugin support

Joehoel:
Adds built-in Varlock support so packages and local files referenced by Varlock configuration are recognized instead of being reported as unused or unresolved. Diagnostics remain attributed to the env file containing the reference.

The implementation uses Varlock’s official [`@env-spec/parser`](https://www.npmjs.com/package/@env-spec/parser) for env-spec syntax. Knip does not execute Varlock or resolve secrets and dynamic expressions; it extracts only the dependency-bearing configuration described below.

### Supported Varlock surface

| Varlock feature | Knip behavior | Varlock documentation |
| --- | --- | --- |
| `@plugin(...)` | Tracks static npm package references and local file/directory references. A fixed npm version is optional because Varlock can download it; unversioned packages and version ranges must be installed. | [Plugins](https://varlock.dev/guides/plugins/) |
| `@import(...)` | Recursively follows static local file and directory imports. Respects literal `enabled=false` and `allowMissing=true`. | [Imports](https://varlock.dev/guides/import/) |
| `@disable` / `@disable=true` | Skips the file and its imports. A dynamic expression is treated conservatively as potentially enabled. | [`@disable`](https://varlock.dev/reference/root-decorators/#disable) |
| `@currentEnv=$NAME` | Selects environment-specific files from a static value in `.env.schema`, `.env`, or `.env.local`, with `process.env` taking precedence. It overrides the CLI `--env` fallback. | [`@currentEnv`](https://varlock.dev/reference/root-decorators/#currentenv) |
| `@envFlag=NAME` | Supports the deprecated static form for compatibility. | [`@envFlag`](https://varlock.dev/reference/root-decorators/#envflag-deprecated) |
| `package.json#varlock.loadPath` | Supports a file, directory, or array of paths. | [`loadPath`](https://varlock.dev/reference/cli-commands/#packagejson-configuration) |
| CLI `--path` / `-p` and `--env` | Uses explicit path and environment arguments found in package scripts. Commands after `varlock run --` continue through Knip’s normal script parser. | [CLI common options](https://varlock.dev/reference/cli-commands/#common-options) |

Dynamic plugin/import paths cannot be resolved statically and are skipped. Dynamic `enabled`, `allowMissing`, and `@disable` expressions are not executed; Knip conservatively follows the reference unless it sees a literal disabling value.

### Files loaded for a directory

| Order | File |
| --- | --- |
| 1 | `.env.schema` |
| 2 | `.env` |
| 3 | `.env.local` |
| 4 | `.env.<currentEnv>` |
| 5 | `.env.<currentEnv>.local` |

Environment-specific files are included only when an environment is selected through `@currentEnv`, its deprecated `@envFlag` equivalent, or the CLI `--env` fallback. A path to a specific file loads only that file and its recursive imports. This mirrors Varlock’s documented [directory loading and precedence](https://varlock.dev/guides/import/#import-precedence-and-merging-multiple-sources).

### Runtime requirement

> [!IMPORTANT]
> `@env-spec/parser` officially requires Node.js 22. Consequently, this PR raises the minimum supported Node.js version for `knip`, `@knip/language-server`, and `@knip/mcp` from Node 20.19 to Node 22.12. Node.js 20 reached end-of-life on March 24, 2026.
>
> This may be a dealbreaker. If dropping Node.js 20 support is not acceptable for this release, using `@env-spec/parser` directly is not viable under its current engine contract. In that case this PR must either retain a small custom env-spec parser or wait until the official parser supports Knip's Node.js range.


webpro:
Thanks for the PR! Would be great to work this plugin in. Just before posting my feedback, noticed it got converted to a draft. Just let me know if/when you want me to post it.

Joehoel:
Yeah please do, i found some other improvements i could add so thats why i converted it back. But please provide feedback!

webpro (commented):
Thanks for this. Varlock support is worth having, and the fixture confirms the gap: on `main` that same tree reports `@varlock/1password-plugin`, `quoted-plugin` and `unscoped-plugin` as unused devDependencies plus `local-plugin.js` as an unused file, and the plugin clears all four. The registration plumbing is correct, the generated files regenerate byte-identically, and lint, format, `tsc`, both knip self-checks and the smoke suite are green.

The blocking issue is the mechanism. `registerCompiler({ extension: '.schema' })` claims a generic extension that Varlock does not own, and the specific consequences are in the inline comments.

Compilers exist for source languages whose files are modules. `.env.schema` is a config file, so the house pattern applies: a plugin-scoped `resolve` returning `Input[]` with `toDependency()` / `toDeferResolve()`, the way `cspell`, `markdownlint`, `husky` and `lefthook` already do. `vite` reads a raw non-JS text file this way (`index.html` via `loadFile` in `plugins/vite/helpers.ts`) without claiming `.html` as a compiler extension. `Plugin.resolve` needs no loader and is called unconditionally, so it works today: one read per workspace, leaving the project glob and the resolver untouched, and inputs carry no position so knip reports the file rather than a wrong location.

Worth settling that shape first, since it removes most of the inline findings at a stroke.

packages/knip/src/plugins/varlock/index.ts
```diff
@@ -0,0 +1,27 @@
...
+const title = 'Varlock';
+
+const enablers = ['varlock'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const entry = ['.env.schema'];
+
+const registerCompilers: RegisterCompilers = ({ registerCompiler, hasDependency }) => {
+  if (hasDependency('varlock')) registerCompiler({ extension: '.schema', compiler });
```
webpro: P1. `.schema` is only the `extname()` tail of the dotfile `.env.schema`, so this registers `**/*.schema` for the whole workspace. Every other plugin-registered extension (`.vue`, `.svelte`, `.astro`, `.prisma`, `.marko`) belongs to its tool.

Three consequences, all reproducible:

- Unrelated `*.schema` files (Avro, JSON Schema, dbt, GraphQL all use the suffix) become project files and are reported unused. `--fix --allow-remove-files` then deletes them. Clean on `main`.
- Module resolution changes project-wide. With `src/config.schema` present and no `src/config.ts`, `import "./config"` resolves to the schema file; without varlock installed it is correctly an unresolved import. The effect leaks into sibling workspaces that never depend on varlock: with varlock in `packages/app` only, `packages/sibling`'s unresolved import is silently suppressed.
- It introduces an `@import(...)` regression. `# @import(./shared/base.env.schema)` now reports `shared/base.env.schema` as an unused file, clean on `main`.

A plugin-scoped `resolve` returning `Input[]` avoids all three.

packages/knip/src/plugins/varlock/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// https://varlock.dev/guides/plugins/
+
+const title = 'Varlock';
+
+const enablers = ['varlock'];
+
+const isEnabled: IsPluginEnabled = ({ dependencies }) => hasDependency(dependencies, enablers);
+
+const entry = ['.env.schema'];
```
webpro: P2. Three gaps in what `entry` covers:

- No `production` counterpart, but Varlock resolves the schema at runtime. `knip -p` on this PR's own fixture reports `local-plugin.js` unused, and a secret-backend plugin in `dependencies` is reported unused under `--production`.
- `.env.schema` matches `.env*`, which `create-next-app` ships verbatim in its generated `.gitignore`. With that ignore present the plugin becomes a silent no-op: the fixture then reports all three plugin packages unused, and only `--no-gitignore` surfaces the real `missing-plugin`. Since `.env.schema` is meant to be committed, this at least wants a `docs.note` on the generated page.
- Plugins are not confined to `.env.schema`. Varlock's docs say they "will be available in all `.env` files in your project" and warn only that an inactive file is skipped. Moving `@plugin(...)` to `.env.production`, a common split for prod-only secret backends, reports the package unused. Varlock also documents `{"varlock": {"loadPath": "./"}}` in `package.json` and a `--path` / `-p` flag, so a workspace whose env files live outside its cwd is not scanned at all. `plugins/prisma/index.ts` handles this shape with `config` plus `resolveConfig` and `args.resolveInputs`.

packages/knip/src/plugins/varlock/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (hasDependency('varlock')) registerCompiler({ extension: '.schema', compiler });
+};
+
+const plugin: Plugin = {
+  title,
+  enablers,
+  isEnabled,
+  entry,
+  registerCompilers,
+};
```
webpro, line 45: P1. No `Plugin.args`, so the script parser discards everything after `--`. `"start": "varlock run -- node app.js"` reports `app.js` unused, and `"dev": "varlock run -- some-cli dev"` reports `some-cli` as an unused devDependency. This is Varlock's primary documented script form.

`fromArgs: parsed => (parsed._[0] === 'run' ? (parsed['--'] ?? []) : [])` clears both and leaves the existing test green; `nx` has the same shape.

packages/knip/src/plugins/varlock/compiler.ts
```diff
@@ -0,0 +1,29 @@
+const pluginMatcher = /^\s*#\s*@plugin\(\s*(?:"([^"\r\n]*)"|'([^'\r\n]*)'|([^)'"\r\n]+?))\s*\)/gm;
```
webpro: P1. The pattern anchors `@plugin` immediately after `#\s*`, so only the first decorator on a line is seen. Varlock allows several root decorators per line, and the first schema example in its README opens with `# @defaultSensitive=false @defaultRequired=infer @currentEnv=$APP_ENV`.

- `# @defaultSensitive=false @plugin(@varlock/1password-plugin)` reports the package unused.
- `# @plugin(first-pkg) @plugin(second-pkg)` emits only `first-pkg`.

Both `\s*` also match newlines, so a lone `#` pairs with a `@plugin` on a later line. `SECRET=\n#\n@plugin(some-pkg)` credits `some-pkg` as used from an uncommented line, hiding a genuine unused-dependency finding. Long whitespace runs backtrack quadratically: 32ms at 2000 blank lines, 8.3s at 32000.

`[^\S\r\n]*` in both positions fixes the over-match and the blowup together.

packages/knip/src/plugins/varlock/compiler.ts
```diff
@@ -0,0 +1,29 @@
...
+const packageNameMatcher = /^(?:@[a-z\d][a-z\d._-]*\/)?[a-z\d][a-z\d._-]*$/;
```
webpro: P1. This re-implements grammar core already owns, and rejected descriptors are dropped with a bare `continue`, turning a used dependency into a false unused report that `--fix` then removes. Confirmed cases, each with the package installed and declared:

- Mixed-case names. `@plugin(JSONStream)` is dropped; legacy mixed-case names are npm-legal and resolve fine.
- Subpaths and `~`. `@plugin(subpath-plugin/entry)` and `@plugin(foo~bar)` are dropped.

`getPackageNameFromModuleSpecifier` (`src/util/modules.ts`) strips version and subpath and is used for exactly this in `plugins/expo/helpers.ts` and `plugins/eslint/helpers.ts`. Feeding raw descriptors through it produces no issues on any of the above.

packages/knip/src/plugins/varlock/compiler.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const compiler = (text: string) => {
+  if (!text.includes('@plugin')) return '';
+  const imports = [];
+  let match: RegExpExecArray | null;
+
+  pluginMatcher.lastIndex = 0;
+  while ((match = pluginMatcher.exec(text))) {
+    const descriptor = (match[1] ?? match[2] ?? match[3])?.trim();
+    if (!descriptor || descriptor.includes(':')) continue;
```
webpro: P2. An unanchored substring test standing in for a URL-scheme test, so it over-drops and under-drops:

- `@plugin(pkg@workspace:*)` and `@plugin(pkg@catalog:)` hit this guard before the version is stripped, so the package is never marked used.
- `@plugin(//cdn.example.com/plugin.js)` has no colon, passes the guard, matches the `startsWith('/')` branch below and is emitted as a local path, surfacing as an unresolved import.

`urlSchemeMatcher` in `plugins/tailwind/compiler.ts` is the anchored scheme test, and `isInternal` (`src/util/path.ts`) replaces the `startsWith` chain on the next line.

packages/knip/src/plugins/varlock/compiler.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const versionSeparator = descriptor.indexOf('@', 1);
+    const packageName = versionSeparator === -1 ? descriptor : descriptor.slice(0, versionSeparator);
+    if ((versionSeparator !== -1 && !descriptor.slice(versionSeparator + 1)) || !packageNameMatcher.test(packageName)) {
+      continue;
+    }
+    imports.push(`import ${JSON.stringify(packageName)};`);
+  }
+
+  return imports.join('\n');
```
webpro: P2. Joining the synthesized imports invents positions. A `@plugin(missing-plugin@2.0.0)` on line 8 of a schema is reported at `.env.schema:1:8`, and the column is always 8, an offset into `import "..."`. The fixture only looks right because its directives sit on lines 1 to 5.

Inputs returned from `resolve` carry no position, so knip reports the file rather than a wrong location.

packages/knip/fixtures/plugins/varlock/.env.schema
```diff
@@ -0,0 +1,11 @@
+# @plugin(@varlock/1password-plugin@^1.0.0)
+# @plugin(unscoped-plugin@1.0.0)
+# @plugin("quoted-plugin@~1.0.0")
+# @plugin(missing-plugin@2.0.0)
+# @plugin('./local-plugin.js')
+# @plugin(https://example.com/plugin.js)
+# @plugin(npm:registry-plugin)
```
webpro: P2. The fixture leans on forms Varlock does not document while omitting ones it does. Varlock's plugin guide lists four descriptor formats: bare package name, versioned specifier, local file path, and local package directory. Quoted, `https` and `npm:` descriptors appear in none of them, yet lines 3, 5, 6 and 7 cover exactly those, and the quoted half of the matcher plus the `match[1] ?? match[2] ?? match[3]` chain exists only to serve them. Deleting the `includes(':')` guard leaves compiler output byte-identical across the whole fixture, so lines 6 and 7 assert nothing.

Missing: version-less scoped `@plugin(@varlock/1password-plugin)`, which is the form in every Varlock example, and the documented directory form `@plugin(../shared/plugin/)`. Also worth pinning: the `.schema` scope, `--production`, the gitignore interaction, and reported positions. `test/compilers/tailwind-import-url.test.ts` is the precedent for table-testing extraction output directly.

Joehoel:
@webpro I fixed all the review comments you flagged, made some improvements myself, and marked it ready for review. 

I am not completely happy yet with some of the parsing logic, and I feel like it could be cleaned up a bit. So maybe you have some ideas on that.

DreamLongYT:
Congrats on new Plugin AND beeing the 2000th Pull Request

packages/knip/src/plugins/varlock/parse.ts
```diff
@@ -0,0 +1,63 @@
+import {
+  parseEnvSpecDotEnvFile,
+  ParsedEnvSpecFunctionCall,
+  ParsedEnvSpecKeyValuePair,
+  ParsedEnvSpecStaticValue,
+} from '@env-spec/parser';
+
+export type Directive = {
```
webpro: P1. This fails the repo's own knip self-check. `node ./packages/knip/src/cli.ts --directory .` exits 1 with `Unused exported types (1) / Directive  type  packages/knip/src/plugins/varlock/parse.ts:8:13`, and it is clean on `main`. `ci.yml` runs both `knip` and `knip --production --strict`, so both go red.

`Directive` is only referenced at line 22, so dropping `export` fixes it.

packages/knip/src/plugins/varlock/scan.ts
```diff
@@ -0,0 +1,97 @@
...
+    const { directives, disabled } = parseVarlockFile(readFileSync(realPath, 'utf8'));
+    if (disabled) continue;
+
+    for (const { name, descriptor, enabled, allowMissing } of directives) {
+      if (!descriptor || descriptor.includes('$') || isExternalUrl(descriptor)) continue;
+
+      if (name === 'plugin') {
+        if (isLocalPath(descriptor)) {
+          inputs.push(
+            toDeferResolve(descriptor.startsWith('~/') ? resolvePath(descriptor, dirname(realPath)) : descriptor, {
```
webpro: P1. Varlock inputs are never marked production, so under `--production` the files `.env.schema` loads at runtime become unused-file false positives.

`knip --production` on this PR's own `fixtures/plugins/varlock` reports `Unused files (3): app.js, local-plugin.js, local-plugin/index.js`, and `--fix --allow-remove-files` deletes them. `create-input-handler.ts:144` drops deferResolve inputs when `isProduction && !input.production`, and line 122 only raises `unlisted` in production when `input.production`.

`varlock-production.test.ts` cannot catch this because its fixture has no source files (`processed: 0, total: 0`). `sst/resolveFromAST.ts:51` shows the helper to use.

packages/knip/src/plugins/varlock/parse.ts
```diff
@@ -0,0 +1,63 @@
...
+        if (arg.key === 'allowMissing') directive.allowMissing = value;
+      }
+      directives.push(directive);
+    }
+
+    for (const item of file.configItems) {
+      const value = getStaticString(item.value);
+      if (value !== undefined) staticValues.set(item.key, value);
+    }
+  } catch {}
```
webpro: P1. This bare `catch {}` turns any parser error into a total loss of directives with no diagnostic, not even under `--debug`, which converts used dependencies into reported-unused that `--fix` then removes.

Two ordinary inputs trip it: a UTF-8 BOM on `.env.schema`, and an indented comment. Either takes this PR's own fixture from `Unlisted dependencies (1) missing-plugin` to `Unused files (2)` plus `Unused devDependencies (9)`.

Knip's own `loadJSON` strips BOMs, and `WorkspaceWorker.ts:477-484` logs and evicts the cache on config-load failure. At minimum this needs `debugLog`.

packages/knip/package.json
```diff
@@ -120,7 +121,7 @@
     "webpack": "^5.110.1"
   },
   "engines": {
-    "node": "^20.19.0 || >=22.12.0"
+    "node": ">=22.12.0"
```
webpro, line 124: P1. Dropping Node 20 is a call for the project, not something a plugin PR should carry.

The narrowing is forced once `@env-spec/parser` is a dependency: restoring `^20.19.0 || >=22.12.0` makes `installed-check --ignore-dev` fail with `Narrower "engines.node" is needed: >=22.12.0`. Worth knowing the parser's code itself runs fine on Node 20.19.3, so the constraint is the declaration rather than the runtime.

Right now it is half-applied. `packages/docs/src/content/docs/overview/getting-started.mdx:10` still reads "Knip v6 requires at least Node.js v20.19.0", the v6 blog post says the same, and `.github/workflows/ci-legacy-5.0.4.yml:42` still runs a `node: [20, 22]` matrix on three operating systems against a runtime this PR declares unsupported.

Either the drop is deliberate and lands with docs, blog post and that workflow updated, or the dependency does not come in. Let me decide that separately and tell you which way to go.

packages/knip/src/plugins/varlock/scan.ts
```diff
@@ -0,0 +1,97 @@
...
+  const sources: Map<string, string>[] = [];
+  for (const path of paths) {
+    if (!isFile(path)) continue;
+    const source = readFileSync(path, 'utf8');
+    const parsed = parseVarlockFile(source);
+    sources.push(parsed.staticValues);
+    key ??= parsed.environmentKey;
+  }
+  if (!key) return fallback;
+  let environment = process.env[key];
```
webpro: P2. Reading `process.env[key]` makes the report depend on the machine running Knip: the file names the variable, and the ambient shell supplies its value, so which `.env.<env>` files are scanned varies per developer.

This PR's own test demonstrates it. `bun test test/plugins/varlock-production.test.ts` passes; `APP_ENV=staging bun test test/plugins/varlock-production.test.ts` fails. `@currentEnv=$NODE_ENV` is the obvious real-world form and CI images export `NODE_ENV`, so CI and local runs disagree. In a scratch fixture, `APP_ENV=production knip --fix` removed `staging-plugin` from package.json.

packages/knip/src/plugins/varlock/scan.ts
```diff
@@ -0,0 +1,97 @@
...
+  if (environment === undefined) for (const values of sources) environment = values.get(key) ?? environment;
+  return environment || fallback;
+};
+
+const expandLoadPath = (path: string, environment?: string) => {
+  if (!isDirectory(path)) return [path];
+  const paths = [join(path, '.env.schema'), join(path, '.env'), join(path, '.env.local')];
+  const currentEnvironment = getEnvironment(paths, environment);
+  if (currentEnvironment) {
+    paths.push(join(path, `.env.${currentEnvironment}`), join(path, `.env.${currentEnvironment}.local`));
```
webpro: P2. Only the resolved environment is expanded, so plugins declared in any other `.env.<env>` are reported unused and `--fix` removes them.

With `@currentEnv=$APP_ENV` and a development default in `.env.schema`, a plugin declared in `.env.production` is reported `Unused dependencies`, and removing it breaks secret resolution on the next deploy. This is fully deterministic, since the value comes from the committed schema.

The fixtures avoid it by leaving `inactive-environment-plugin` and `inactive-staging-plugin` out of package.json, so no test can regress on it.

packages/knip/src/plugins/varlock/scan.ts
```diff
@@ -0,0 +1,97 @@
...
+const isExternalUrl = (path: string) => path.startsWith('//') || urlSchemeMatcher.test(path);
+
+const isLocalPath = (path: string) => path.startsWith('~/') || isInternal(path);
+
+const getEnvironment = (paths: string[], fallback?: string) => {
+  let key: string | undefined;
+  const sources: Map<string, string>[] = [];
+  for (const path of paths) {
+    if (!isFile(path)) continue;
+    const source = readFileSync(path, 'utf8');
```
webpro: P2. This `readFileSync` (and the one at line 58) runs after only an `isFile` stat, and `plugin.resolve` is the one hook `WorkspaceWorker` does not wrap in try/catch, so an unreadable env file aborts the whole run.

`chmod 000` on `.env.local` in the fixture exits 1 with a raw `EACCES: permission denied` stack trace through `getEnvironment -> expandLoadPath -> scanVarlockFiles -> resolve -> build -> cli`. A root-owned `.env.local` in a container bind mount does the same. Every other file read in Knip goes through `loadFile`/`LoaderError` or `this.readFile`; `WorkspaceWorker.ts:467` wraps `resolveConfig` but nothing wraps `resolve`.

webpro (commented):
Big step up. Every blocking item from the first round is genuinely fixed, not papered over:

- The compiler is gone. `Plugin.resolve` plus `args.resolveInputs` is the right mechanism, and it removes the `.schema` glob widening, the project-wide resolver mutation, the `--fix --allow-remove-files` hazard on unrelated files, and the invented positions in one move.
- Descriptor handling now goes through `getPackageNameFromModuleSpecifier`, `isInternal` and an anchored scheme test, so `JSONStream`, subpaths and `workspace:` descriptors resolve.
- A real parser replaced the regex, so multi-decorator lines and the newline over-match are gone.
- `varlock run -- <cmd>` is handled, `@import` is followed, `loadPath` and `--path` are supported, and the gitignore no-op is fixed.

Four blockers inline: the repo's own knip self-check now fails, `--production` reports the fixture's own runtime files as unused, a parse failure silently deletes dependencies, and the Node 20 drop needs a decision from me rather than from a plugin PR.

## On cleaning up the parsing logic

You flagged this yourself, and the shape that would help most is separating "resolve the environment" from "collect directives" so each file is parsed once. `getEnvironment` and the main loop in `scanVarlockFiles` never share a parse, so every env file is read and PEG-parsed at least twice.

Smaller things in the same area:

- `getStrings` overlaps `arrayify` in `src/util/array.ts`.
- `urlSchemeMatcher` is now the third copy of that regex, alongside `src/compilers/scss.ts` and `src/plugins/tailwind/compiler.ts`. Worth extracting to `src/util/`.
- `getEnvironment` ignores the `disabled` flag it already computes, so a `@disable`d `.env.local` still supplies the `@currentEnv` value while the main loop correctly skips the file.

One dependency-weight note for later: `@env-spec/parser` is imported at the top level through the plugins barrel, so its ~107KB ESM bundle loads on every Knip run for every user, not just Varlock users.

packages/knip/src/plugins/varlock/scan.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+  if (!key) return fallback;
+  let environment = process.env[key];
+  if (environment === undefined) for (const values of sources) environment = values.get(key) ?? environment;
+  return environment || fallback;
+};
+
+const expandLoadPath = (path: string, environment?: string) => {
+  if (!isDirectory(path)) return [path];
+  const paths = [join(path, '.env.schema'), join(path, '.env'), join(path, '.env.local')];
```
webpro: P2. `.env` and `.env.local` are probed directly, bypassing the gitignore filtering the rest of discovery applies, so Knip reads local secrets files and reports issues located in them.

With `.gitignore` containing `.env*` and a `@plugin()` in `.env.local`, the report is `Unlisted dependencies (1) secret-only-plugin  .env.local` — an issue pointing at a file no colleague or CI runner has, so results and `--fix` behaviour differ per machine. Every user who merely installs varlock gets this, since `isEnabled` is dependency-only and `resolve` needs no config file.

packages/knip/src/plugins/varlock/scan.ts
```diff
@@ -0,0 +1,97 @@
...
+          );
+        } else {
+          const packageName = getPackageNameFromModuleSpecifier(descriptor);
+          const versionSeparator = descriptor.indexOf('@', 1);
+          const version = versionSeparator === -1 ? undefined : descriptor.slice(versionSeparator + 1);
+          if (packageName) {
+            inputs.push(
+              toDependency(packageName, {
+                containingFilePath: realPath,
+                optional: exactVersionMatcher.test(version ?? ''),
```
webpro: P2. This inverts Varlock's rule for the only context Knip runs in.

Varlock's plugin guide states that in a JavaScript project with a `package.json` you must install the plugin as a dependency and load it without a version specifier; auto-download applies to the standalone binary with no `package.json`. So marking pinned descriptors optional suppresses exactly the reports that matter: `@plugin(pinned-not-installed@2.0.0)` is silently skipped while `@plugin(ranged-not-installed@^2.0.0)` is reported.

`.env.schema:6` bakes this into the fixture with `downloaded-plugin@2.0.0`.

packages/knip/src/WorkspaceWorker.ts
```diff
@@ -375,7 +376,8 @@ export class WorkspaceWorker {
 
       const inputs: Input[] = [];
       const seenConfigFiles = getSeenConfigFiles(pluginName);
-      const addInput = (input: Input, containingFilePath = input.containingFilePath) => {
+      const addInput = (input: Input, fallbackContainingFilePath = containingFilePath) => {
+        const containingFilePath = input.containingFilePath ?? fallbackContainingFilePath;
```
webpro: P2. This inverts `addInput` precedence for four call sites and the eight plugins that set `containingFilePath`, but it is inert on every existing fixture and has no test of its own.

Keeping `addInput` as it is on `main` and changing the single call site instead passes all four varlock tests:

```js
for (const id of dependencies) addInput(id, id.containingFilePath ?? containingFilePath);
```

Worth knowing either way: `create-input-handler.ts:149-150` prefers `input.dir` over `containingFilePath`, so wherever a plugin stamps `dir` on its inputs the containing path still loses. `taskfile/index.ts:119` is the case in tree today.

packages/knip/src/plugins/varlock/parse.ts
```diff
@@ -0,0 +1,63 @@
...
+  let environmentKey: string | undefined;
+
+  try {
+    const file = parseEnvSpecDotEnvFile(source);
+
+    for (const decorator of file.decoratorsArray) {
+      if (decorator.name === 'disable') disabled ||= decorator.simplifiedValue === true;
+
+      if (decorator.name === 'currentEnv' && decorator.value instanceof ParsedEnvSpecFunctionCall) {
+        const value = decorator.value.name === 'ref' ? getStaticString(decorator.value.data.args.values[0]) : undefined;
```
webpro: P3. `@currentEnv` is only understood when its value is a bare `$VAR`. The documented dynamic forms yield no environment key, so no `.env.<env>` file is scanned at all.

Checked through `parseVarlockFile`: `$APP_ENV` and `${APP_ENV}` give `APP_ENV`, while `fallback($APP_ENV, development)` and `remap($VARLOCK_ENV, production=prod)` both give `undefined`. A repo using `@currentEnv=fallback($APP_ENV, development)` with `.env.development` declaring a plugin never has that file read.

packages/knip/src/plugins/varlock/parse.ts
```diff
@@ -0,0 +1,63 @@
...
+
+      if (decorator.name !== 'plugin' && decorator.name !== 'import') continue;
+      const args = decorator.bareFnArgs?.values;
+      const descriptor = getStaticString(args?.[0]);
+      if (descriptor === undefined) continue;
+
+      const directive: Directive = { name: decorator.name, descriptor };
+      for (const arg of args?.slice(1) ?? []) {
+        if (!(arg instanceof ParsedEnvSpecKeyValuePair)) continue;
+        const value = getStaticBoolean(arg.value);
```
webpro, line 35: P3. `getStaticBoolean` collapses "argument absent" and "argument present but dynamic" into the same `undefined`, so a conditional reads as unconditionally enabled.

Checked through `parseVarlockFile`: `@import(./x.env, enabled=false)` yields `enabled: false`, but `@import(./x.env, enabled=forEnv(production))` yields a directive with no `enabled` key at all. Varlock evaluates `enabled` and skips the import before its existence check, so a conditionally-disabled missing import is a false `Unresolved imports` here — on the exact pattern `enabled=` exists for. `@disable=forEnv(...)` has the same root cause at line 31.

webpro:
> @webpro I fixed all the review comments you flagged, made some improvements myself, and marked it ready for review.
> 
> I am not completely happy yet with some of the parsing logic, and I feel like it could be cleaned up a bit. So maybe you have some ideas on that.

Thanks, that's some nice improvements. Much appreciated. I am in the phase of both not having much bandwidth and also trying out agentic reviews. I review the review carefully and I think it's working quite well (also e.g. #1999). But I do realize it might feel a bit distant, sorry about that.

Joehoel:
TBH, does feel a bit distant. But its fine. I did this whole PR with an agent too.

From your reply/review, I'm not quite sure if I should keep the `@env-spec/parser` package that requires us to drop Node version 20. Can you decide on that?

webpro:
> TBH, does feel a bit distant. But its fine. I did this whole PR with an agent too.

No worries! On the other hand it's also not entirely fair to expect maintainers to do the whole review manually knowing that most agentic PRs aren't high quality to begin with. At this point I don't mind too much, but yeah this is where it's at. Feels like just a bit of a personal human touch goes a long way already. Cheers, will have more time soon.

webpro:
Thanks for fixing it all up, that's some good progress. Yet the new dependency (and the Node 22 bump, #1853) blocks this PR: none of the >180 plugins warrants its own production dependency (lots of efforts went/go into keeping Knip mean & lean). So I'm inclined to think removing the dependency is the way forward, unless we'd be re-implementing it in full.

Joehoel:
I've replaced the dep with a custom parser

packages/knip/src/plugins/varlock/parse.ts
```diff
@@ -0,0 +1,179 @@
...
+};
+
+const pairs: Record<string, string> = { '(': ')', '[': ']', '{': '}' };
+
+const getRootComments = (source: string) => {
+  const comments: string[] = [];
+  let block: string[] = [];
+  let hasConfigItem = false;
+
+  for (const line of source.replace(/^\uFEFF/, '').split('\n')) {
```
webpro, line 15: P1: Preserve CRLF line endings when reading root comments. [parse.ts:15](https://github.com/webpro-nl/knip/blob/95d687d4285737e44d46c20c68e86b9d3f87dd03/packages/knip/src/plugins/varlock/parse.ts#L15) splits only on `\n`, leaving `\r` at the end of each line. The comment regex then fails, treats the first header line as a config item, and returns no directives. A CRLF file containing `# @plugin(apple-plugin)` reports that used dependency as unused, and `--fix --fix-type dependencies` removes it. The same content with LF, and the previous parser with CRLF, are clean. Normalize CRLF before matching and add a test with an explicit `\r\n` string; the repository normalizes fixture line endings to LF.

webpro (changes requested):
The dependency and Node-version blockers are resolved. One parser regression remains below.

packages/knip/src/plugins/varlock/parse.ts
Joehoel: Fixed!

webpro (approved):
Thanks a lot! Great new plugin we got here 🙏

webpro:

PR #2004, opened 2026-09-03
fix: don't flag a tag that suppresses member issues

### LINKED ISSUE #1957: 🐛 `tags` exemption is reported back as an "Unused tag" hint

devYRPauli:
fixes #1957

A `@knipignore` tag on an export gets an "Unused tag" hint whenever the export itself is referenced. The check stops there. The same tag also exempts that export's members, so a tag that suppresses enum member or namespace member issues is reported as unused.

`analyze.ts` emitted the hint at the top of the export branch, then ran `if (isIgnored) continue;` before the member scan. At that point the code cannot know whether the tag suppressed anything.

The member scan now lives in `getMemberIssues`, and the export-level hint fires only when that scan finds nothing to report. The scan itself is unchanged: same guards, same order, same per-member checks. It had one call site before and it has two now, so the ignored branch and the reported branch ask the same question.

Measured on the new fixture. It holds three exports, each tagged `@knipignore`, with `tags: ["-knipignore"]` in `knip.json`:

| export | members used | members the tag hides | hint on main | hint on this branch |
|---|---|---|---|---|
| `StatusCode` (enum) | `normal` | `warning`, `critical` | yes | no |
| `Level` (enum) | `low`, `high` | none | yes | yes |
| `Settings` (namespace) | `timeout` | `retries` | yes | no |

`Level` is the control. Its tag really does nothing, so the hint stays.

The second test drops the exemption with `tags: ['-unrelated']`. It then reports what the tag hides: `StatusCode.warning`, `StatusCode.critical` and `Settings.retries`.

One note on the reporter's own reproduction. On `main` today it prints the hint, and that is now correct. #1956 landed, so `StatusCode[Number.parseInt(code)]` counts every member as used. The tag there hides nothing. The fixture reaches the same defect through a case that survives that fix, where one member of the enum is used by name.

Verified:

```
node scripts/run-test.js --runtime node    1101 passed, 0 failed
pnpm build                                 exit 0
pnpm exec oxlint                           no findings
pnpm exec oxfmt --check                    all matched files use the correct format
pnpm knip && pnpm knip:production          no findings
```

With the `analyze.ts` change reverted, the first new test fails and the second still passes.

### DIFF
--- a/packages/knip/fixtures/tags-hints/tag-suppresses-members/index.ts
+++ b/packages/knip/fixtures/tags-hints/tag-suppresses-members/index.ts
@@ -1,0 +1,5 @@
+import { isNormal, levels, timeout } from './utils.js';
+
+isNormal(1000);
+levels;
+timeout;
--- a/packages/knip/fixtures/tags-hints/tag-suppresses-members/knip.json
+++ b/packages/knip/fixtures/tags-hints/tag-suppresses-members/knip.json
@@ -1,0 +1,3 @@
+{
+  "tags": ["-knipignore"]
+}
--- a/packages/knip/fixtures/tags-hints/tag-suppresses-members/package.json
+++ b/packages/knip/fixtures/tags-hints/tag-suppresses-members/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "@fixtures/tag-suppresses-members"
+}
--- a/packages/knip/fixtures/tags-hints/tag-suppresses-members/status.ts
+++ b/packages/knip/fixtures/tags-hints/tag-suppresses-members/status.ts
@@ -1,0 +1,18 @@
+/** @knipignore */
+export enum StatusCode {
+  normal = 1000,
+  warning = 3000,
+  critical = 4000,
+}
+
+/** @knipignore */
+export enum Level {
+  low = 1,
+  high = 2,
+}
+
+/** @knipignore */
+export namespace Settings {
+  export const timeout = 1000;
+  export const retries = 3;
+}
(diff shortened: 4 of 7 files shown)


webpro:
Thanks! ✨

webpro:

PR #2006, opened 2026-09-04
fix: resolve Nuxt Virtual Component Imports

### LINKED ISSUE #2005: Nuxt: components imported from '#components' are reported as unused files
### Description

In a Nuxt project, a component used **only** via the `#components` virtual module is reported as an unused file:

```ts
import { LanguageSelect } from '#components'
```

`plugins/nuxt/index.ts` ignores that specifier:

```ts
toIgnore('#components', 'unresolved'),
```

Ignoring it silently suppresses the unresolved-import warning, but the import then creates no reference, so the component file is reported under `files`.

Deleting such a file on knip's advice breaks the build:

```
[unimport] failed to find "LanguageSelect" imported from "#components"
```

In our repo 12 of 105 reported unused files were live components imported this way.

`#imports` presumably has the same issue for auto-imported composables.

### Version

knip 6.34.0

gioboa:
Closes #2005

Prevent #components imports from being reported as unused files.

### DIFF
--- a/packages/knip/fixtures/plugins/nuxt-auto-import/app.vue
+++ b/packages/knip/fixtures/plugins/nuxt-auto-import/app.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { StatusBadge } from '#components';
+
 const { count } = useCounter();
 const date = ref(formatDate(new Date()));
 const theme = { default: 'system' };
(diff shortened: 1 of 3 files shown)


webpro:
Thank you! No notes.

webpro:

PR #2007, opened 2026-09-04
Mark the Knip CLI as a production entry

ayush-that:
## Summary

- mark the Knip CLI as a production entry
- avoid redundant-entry hints for production-only patterns
- add regression coverage for production entries also referenced by package scripts

## Testing

- `node --test test/tags-hints/configuration-hints.test.ts`
- `pnpm build`
- `pnpm knip`
- `pnpm knip:production`
- `pnpm --dir ../.. run ci`

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -5,7 +5,7 @@
       "project": ["!templates"]
     },
     "packages/knip": {
-      "entry": ["test/**/*.ts"],
+      "entry": ["src/cli.ts!", "test/**/*.ts"],
       "project": ["src/**/*.ts!", "!src/util/empty.ts", "!**/_template"],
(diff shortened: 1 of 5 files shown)


ayush-that:
hi. small backstory on this pr.

muse spark 1.3 is a benchmaxxed dumb model. i was testing it and prompted it with something like, "install knip on my production codebases and make a pr." i had a bunch of repos in the cwd, and this dude somehow went and made a pr here instead of on my codebase.

in another incident, i asked it to run tokei on "this" repo. the cwd was a git repo. bro installed tokei, cloned tokei's own repo, and then ran tokei on that instead.

it genuinely feels like the model has no intuition for cwd. dumb.

PR #2008, opened 2026-09-04
fix(nuxt): resolve components imported from '#components'

### LINKED ISSUE #2005: Nuxt: components imported from '#components' are reported as unused files

fabkho:
Closes #2005.

### Problem

`#components` resolves to nothing on disk, so it is ignored as unresolved:

```ts
toIgnore('#components', 'unresolved'),
```

That suppresses the warning, but it also means a named import from it creates no reference — so a component used only this way is reported as an unused file:

```ts
import { LanguageSelect } from '#components'
```

Deleting such a file on that advice breaks the build:

```
[unimport] failed to find "LanguageSelect" imported from "#components"
```

In the repo where this surfaced, 12 of 105 reported unused files were live components imported this way.

### Fix

Map named imports from `#components` back to their files using the component map the plugin already builds from `.nuxt/components.d.ts` — the same source template-tag resolution uses. Bindings are aliased (`Foo$knip`) so the original import remains valid in the compiled output.

The `.ts` compiler path bailed out when the import map was empty, so it now also runs when only the component map is populated — `.ts` files importing from `#components` (typing a template ref, for instance) are covered too.

### Test

`nuxt-auto-import` gains `components/VirtualOnly.vue`, referenced **only** through `import type { VirtualOnly } from '#components'` in `app.vue` and never used as a tag, so it isolates this path.

- without the fix: `files: 3` — `VirtualOnly.vue` reported unused
- with the fix: `files: 2` — correctly resolved

Full suite on this machine: **992 pass / 107 fail**, against a baseline of **991 / 108** on the same checkout — one test fixed, nothing else changed. The 107 appear to be pre-existing environment failures here rather than anything related to this change.

### Note

`#imports` is likely affected the same way for auto-imported composables, but I have not touched it — happy to follow up if you would like it in the same shape.

### DIFF
--- a/packages/knip/src/plugins/_vue/auto-import.ts
+++ b/packages/knip/src/plugins/_vue/auto-import.ts
@@ -227,6 +227,31 @@ const getSyntheticImports = (maps: AutoImportMaps, identifiers: Set<string>, tem
   return syntheticImports;
 };
 
+
+const virtualComponentsMatcher = /import\s+(?:type\s+)?\{([^}]*)\}\s*from\s*['"]#components['"]/g;
+
+/**
+ * Nuxt exposes auto-imported components through the `#components` virtual module.
+ * That specifier resolves to nothing on disk, so imports from it are ignored and
+ * the component files look unused. Map the named imports back to their files.
+ */
+const getVirtualComponentImports = (source: string, componentMap: Map<string, string[]>) => {
+  if (componentMap.size === 0 || !source.includes('#components')) return [];
+  const imports: string[] = [];
+  for (const [, names] of source.matchAll(virtualComponentsMatcher)) {
+    for (const binding of names.split(',')) {
+      const name = binding.replace(/^\s*type\s+/, '').trim().split(/\s+as\s+/)[0]?.trim();
+      if (!name) continue;
+      const specifiers = componentMap.get(name);
+      if (!specifiers) continue;
+      // aliased: the original binding is still in scope in the compiled output
+      imports.push(`import { default as ${name}$knip } from '${specifiers[0]}';`);
+      for (let i = 1; i < specifiers.length; i++) imports.push(`import '${specifiers[i]}';`);
+    }
+  }
+  return imports;
+};
+
 const compileVueSfc = (source: string, path: string, maps: AutoImportMaps, root: string) => {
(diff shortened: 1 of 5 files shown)


gioboa:
#2006 is solving the same issue :eyes:

packages/knip/src/plugins/_vue/auto-import.ts
```diff
@@ -227,6 +227,31 @@ const getSyntheticImports = (maps: AutoImportMaps, identifiers: Set<string>, tem
   return syntheticImports;
 };
 
+
+const virtualComponentsMatcher = /import\s+(?:type\s+)?\{([^}]*)\}\s*from\s*['"]#components['"]/g;
```
webpro, line 231: **P0: Parse `#components` imports from the OXC AST**

The raw-source matcher does not model valid import syntax. It misses lazy component imports because `componentMap` intentionally omits generated `Lazy*` declarations, even though [Nuxt's direct-import example](https://nuxt.com/docs/4.x/directory-structure/app/components#direct-imports) imports `LazyMountainsList` from `#components`. With `import type { LazyVirtualOnly } from '#components'`, this branch still reports `components/VirtualOnly.vue` as unused. The matcher also accepts a commented-out import and incorrectly keeps that file live. Collect `ImportDeclaration` specifiers during the existing OXC visit, normalize the `Lazy` prefix to the base component name, and emit side-effect imports for the mapped specifiers so the existing oxc-resolver path retains ownership of module resolution.

webpro (changes requested):
Requesting changes because the new raw-source matcher misses a documented Nuxt import form and can also hide genuinely unused components.

webpro:
Closing this as dupe of #2006.

PR #2012, opened 2026-09-04
fix: honor explicit entry export analysis

### LINKED ISSUE #2011: includeEntryExports has no effect when a package.json script makes the file an entry (monorepo)
### Summary

With `includeEntryExports: true`, a dead export in a workspace entry file is reported — **unless** a `package.json` script in the root workspace also references that file. Adding the script silently switches the export back to exempt, and knip then advises removing the `entry` pattern that would have covered it.

The two arms below differ by **one line** in the root `package.json`.

### Versions

| | |
|---|---|
| knip | 6.32.2 |
| node | v22.22.2 |
| typescript | 6.0.3 |
| OS | Linux 6.8.0 |

### Reproduction

```bash
mkdir -p repro/packages/app/src && cd repro

cat > package.json <<'JSON'
{
  "name": "knip-entry-exports-repro",
  "private": true,
  "type": "module",
  "workspaces": ["packages/*"],
  "scripts": { "dev": "node packages/app/src/main.ts" },
  "devDependencies": { "knip": "6.32.2", "typescript": "6.0.3" }
}
JSON

cat > packages/app/package.json <<'JSON'
{ "name": "@repro/app", "private": true, "type": "module", "version": "0.0.0" }
JSON

cat > knip.jsonc <<'JSON'
{
  "includeEntryExports": true,
  "workspaces": {
    "packages/app": { "entry": ["src/main.ts"], "project": ["src/**/*.ts"] }
  }
}
JSON

printf 'export const DEAD = 1\n' > packages/app/src/main.ts

npm install
npx knip --no-progress          # arm 1
```

### Arm 1 — with the root script

```
Configuration hints (1)
src/main.ts  packages/app  knip.jsonc  Remove redundant entry pattern
```
exit 0. **`DEAD` is not reported.**

### Arm 2 — delete `"scripts"` from the root `package.json`, change nothing else

```
Unused exports (1)
DEAD  packages/app/src/main.ts:1:14
```
exit 1. **`DEAD` is reported.**

`includeEntryExports: true` and `entry: ["src/main.ts"]` are identical in both arms.

### Expected

Arm 1 reports `DEAD`, like arm 2. The file is a declared entry of `packages/app` and `includeEntryExports` is on.

### Why this is easy to miss

Two things compound:

1. The exemption is invisible — knip exits 0 and says nothing about the export.
2. The hint actively pushes you toward it: `Remove redundant entry pattern` names the very declaration that, on its own, would have made the export checked.

### Where it was found

A pnpm monorepo whose root `package.json` has `"dev": "tsx packages/tui/src/main.tsx"` and `"exec": "tsx packages/cli/src/main.ts"`. All four such files were exempt from the dead-export gate while the config asked for the opposite. Narrowed from that repo to the fixture above; I also checked and could **not** reproduce with a single package, with an auto-discovered entry, or with a workspace entry alone — the root-script reference is the trigger.

### Possibly intended?

If a script-derived entry is deliberately treated as a public surface, that would explain it — but then `includeEntryExports` cannot mean what the name says for those files, and the "redundant entry pattern" hint is misleading. Happy to be told this is by design; in that case documenting the interaction would be enough.

I did not find an existing issue for this: #1442 requests the capability that `includeEntryExports` now provides, and #564 discusses the option with `.js` files.

gioboa:
Closes #2011

Explicit workspace entries now override script-derived export exemptions and avoid incorrect redundancy hints.

### DIFF
--- a/packages/knip/src/ProjectPrincipal.ts
+++ b/packages/knip/src/ProjectPrincipal.ts
@@ -31,6 +31,7 @@ export class ProjectPrincipal {
   projectPaths = new Set<string>();
   programPaths = new Set<string>();
   skipExportsAnalysis = new Set<string>();
+  private includeExportsAnalysis = new Set<string>();
 
   pluginCtx: PluginVisitorContext = {
     filePath: '',
@@ -164,7 +165,14 @@ export class ProjectPrincipal {
     if (!isInNodeModules(filePath) && this.hasAcceptedExtension(filePath)) {
       this.entryPaths.add(filePath);
       this.projectPaths.add(filePath);
-      if (options?.skipExportsAnalysis) this.skipExportsAnalysis.add(filePath);
+      if (options) {
+        if (options.skipExportsAnalysis) {
+          if (!this.includeExportsAnalysis.has(filePath)) this.skipExportsAnalysis.add(filePath);
+        } else {
+          this.includeExportsAnalysis.add(filePath);
+          this.skipExportsAnalysis.delete(filePath);
+        }
+      }
       this.onPathAdded?.(filePath);
     }
   }
@@ -190,6 +198,8 @@ export class ProjectPrincipal {
   removeProjectPath(filePath: string) {
     this.entryPaths.delete(filePath);
     this.projectPaths.delete(filePath);
+    this.skipExportsAnalysis.delete(filePath);
+    this.includeExportsAnalysis.delete(filePath);
     this.invalidateFile(filePath);
     this.deletedFiles.add(filePath);
   }
--- a/packages/knip/src/ProjectPrincipal.ts
+++ b/packages/knip/src/ProjectPrincipal.ts
@@ -59,6 +59,7 @@ export class ProjectPrincipal {
   toSourceFilePath: ToSourceFilePath;
   private findWorkspacePackageTarget: WorkspacePackageTargetHandler | undefined;
   private findWorkspaceNameByFilePath: (filePath: string) => string | undefined;
+  private isReportExports: boolean;
 
   fileManager: SourceFileManager;
   private resolveModule: ResolveModule = () => undefined;
@@ -78,6 +79,7 @@ export class ProjectPrincipal {
     this.toSourceFilePath = toSourceFilePath;
     this.findWorkspacePackageTarget = findWorkspacePackageTarget;
     this.findWorkspaceNameByFilePath = findWorkspaceNameByFilePath;
+    this.isReportExports = options.isReportExports;
     this.tsConfigFile = options.tsConfigFile ? toAbsolute(options.tsConfigFile, options.cwd) : undefined;
     this.pluginVisitorObjects.push(createBunShellVisitor(this.pluginCtx));
     this.fileManager = new SourceFileManager({
@@ -226,7 +228,7 @@ export class ProjectPrincipal {
         const isProjectPath = this.projectPaths.has(filePath);
 
         // Cached project files: skip read+parse and pass the cached FileNode through.
-        const cachedFile = isProjectPath ? this.cache.getCachedFile(filePath) : undefined;
+        const cachedFile = isProjectPath ? this.getCachedFile(filePath) : undefined;
 
         if (cachedFile) {
           const internalPaths = analyzeFile(filePath, undefined, '', cachedFile);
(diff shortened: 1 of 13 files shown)


usetheodev:
Reporter of #2011 here — I built this branch and ran it against the fixture from that issue, plus a control. It fixes the case I filed.

### Verified

Fixture from #2011 (root `scripts.dev` naming a file inside a workspace, `includeEntryExports: true`, `entry: ["src/main.ts"]` in the workspace):

| build | result |
|---|---|
| `main` @ c6497a1 | `Configuration hints (1) src/main.ts … Remove redundant entry pattern` — exit 0, `DEAD` **not** reported |
| this branch @ 2e88de4 | `Unused exports (1) DEAD  packages/app/src/main.ts:1:14` — exit 1 |

The misleading hint is gone too, which was half of what made this hard to notice.

The control matters: `main` was rebuilt from source and still reproduces, so the fix is this PR and not something between 6.32.2 (where I originally measured) and 6.34.0.

### It does real work outside the fixture

Run against the repository where I found this — a pnpm monorepo whose workspace `entry` globs are the test files:

```
main @ c6497a1     exit 0, clean
+ this PR          exit 1
                   Unused exports (1)
                   omittingAGateMustNotCompile  packages/agent/src/hooks/fail-safe-defaults.test.ts:43:17
```

A true positive: a compile-time assertion helper, exported only so eslint would not flag an unused local. Its own docblock says *"the dead-export scan lists it"* — written by someone who assumed it was covered. It had never been reported, because test files are that workspace's declared entries. Exempting it locally with `tags: ["-knipignore"]` + `/** @knipignore */` silences it cleanly, verified on this branch.

So the change surfaces genuinely invisible exports, not just fixture ones.

### One case still exempt — asking rather than asserting

When the file is an entry **only** through the root script, with no explicit `entry` for it in the workspace:

```jsonc
// knip.jsonc
"workspaces": { "packages/app": { "entry": ["src/**/*.test.ts"], "project": ["src/**/*.ts"] } }
// root package.json
"scripts": { "dev": "node packages/app/src/main.ts" }
```

`DEAD` in `packages/app/src/main.ts` is still not reported on this branch (exit 0).

That may be exactly the intended semantics — an explicit declaration is a signal, and without one a script-derived entry is a program entry point whose exports are arguably its surface. #2011 asked the same question and I have no stake in the answer; I only want the boundary written down somewhere, since the option is named `includeEntryExports` and that case is an entry whose exports are not included.

If it is intended, the doc line is enough and the remaining fix for consumers is "declare the entry" — which this PR now makes work.

### Reproduce

```bash
git fetch origin pull/2012/head:pr2012 && git switch pr2012
cd packages/knip && npx tsc
node packages/knip/bin/knip.js --no-progress   # from the fixture above
```

Thanks for turning this around so fast.

packages/knip/src/ProjectPrincipal.ts
```diff
@@ -164,7 +165,14 @@ export class ProjectPrincipal {
     if (!isInNodeModules(filePath) && this.hasAcceptedExtension(filePath)) {
       this.entryPaths.add(filePath);
       this.projectPaths.add(filePath);
-      if (options?.skipExportsAnalysis) this.skipExportsAnalysis.add(filePath);
+      if (options) {
+        if (options.skipExportsAnalysis) {
+          if (!this.includeExportsAnalysis.has(filePath)) this.skipExportsAnalysis.add(filePath);
+        } else {
+          this.includeExportsAnalysis.add(filePath);
+          this.skipExportsAnalysis.delete(filePath);
```
webpro, line 175: [P1] Revalidate cached file analysis when entry policy changes

The new precedence is rebuilt from configuration, but unchanged source files are returned from `CacheConsultant` before `skipExportsAnalysis` is consulted. With `--cache`, seed a run while `src/custom.js` is script-only, then add it as an explicit workspace entry without touching the source: the next run exits 0 with no findings, while the same uncached run reports the unused exports. Removing the explicit entry after seeding the cache produces the inverse stale false positives. Store the export-collection mode with the cached file or invalidate and reanalyze when the effective mode changes.

webpro (changes requested):
Thanks for this. The entry precedence looks right on uncached runs, but configuration changes can leave cached results stale. I'd like to see that fixed before merging.

packages/knip/src/ProjectPrincipal.ts
gioboa: Thanks for the review. fixed by [fc71d53](https://github.com/webpro-nl/knip/pull/2012/commits/fc71d538982201c73a6ca80a9dfc2f3b1999326d)

webpro:
Thanks!

Pushed a nit (test failed against the old implementation).

gioboa (commented):
Thanks @webpro

webpro:

PR #2017, opened 2026-09-05
fix: ignore raw Vite import globs

### LINKED ISSUE #2016: 🐛 Vite plugin ignores `import.meta.glob` options, so a `?raw` source scan silently disables unused-export analysis
### Prerequisites

- [x] I'm using the latest version (6.34.0)

### Reproduction url

https://github.com/kellyhair/knip-import-meta-glob-raw-repro

### Reproduction access

### Description of the issue

The Vite plugin's `import.meta.glob` visitor reads the glob **pattern** but never looks at the **options object**. A glob that reads files as source text with `query: '?raw'` is therefore recorded as a genuine module import of every matched file.

The consequence that matters is not the file-reachability side — it's that **unused-export analysis is silently switched off** for everything the glob matches.

#### Reproduction

`src/scan.ts` in the linked repo:

```ts
const sources: Record<string, string> = import.meta.glob('./**/*.ts', {
  query: '?raw',
  eager: true,
  import: 'default',
});
```

This yields **strings**. No module is instantiated and no export is consumed.

```sh
npm install && npx knip
```

**Actual:** no output — 0 unused files, 0 unused exports.

**Expected:** `src/orphan.ts` (imported by nothing) and `neverCalledExport` in `src/used.ts` (an unused export in a module that *is* imported normally) both reported.

Adding `"vite": false` to `knip.json`, changing nothing else, produces both:

```
Unused files (1)
src/orphan.ts
Unused exports (1)
neverCalledExport  function  src/used.ts:2:17
```

#### Why the unused-export half looks clearly wrong

I can see an argument that a globbed file is "referenced" and so shouldn't be reported as an unused *file* — `fsPromisesGlob.ts` treats `fs.glob` the same way, which reads like a deliberate choice.

But `neverCalledExport` lives in `src/used.ts`, a module that is imported normally by the entry file. Reading that file's *text* cannot consume its exports. Treating a raw text read as an importer of every named export means one such glob disables unused-export detection across everything it matches — in our case a spec globbing `'../../**/*.{ts,tsx}'` from two directories deep, which resolves to the whole `src` tree.

#### How it was found

We added knip as a blocking CI gate and it reported **0 unused files, 0 unused exports** on first run. That was a disarmed gate, not a clean tree — a planted orphan proved it, and the real baseline was 124 findings. Worth noting for anyone adopting knip in CI: plant a known orphan and confirm it's caught before recording a baseline.

#### Cause

`packages/knip/src/plugins/vite/visitors/importMetaGlob.ts` — `node.arguments[1]` is never examined:

```ts
const arg = node.arguments[0];
let patterns: string[] | undefined;
if (isStringLiteral(arg)) { patterns = [getStringValue(arg)!]; }
else if (arg.type === 'ArrayExpression') { /* ... */ }

if (!patterns?.length) return;

ctx.addImportGlob(patterns);   // options never consulted
```

`query`, `as`, `eager` and `import` are all invisible to it.

#### Also affected

The legacy Vite spelling `{ as: 'raw' }` behaves identically.

A glob **without** a raw query — `import.meta.glob('./**/*.ts', { eager: true })` — correctly suppresses the orphan. Those genuinely are module imports, so that behavior is right and shouldn't change.

#### Possible direction

Detect a raw read (`query` containing `raw`, or legacy `as: 'raw'`) and either skip `addImportGlob` for it, or record it in a way that marks files referenced without crediting their exports as used. The second preserves the `fs.glob` semantics while fixing the export half.

I'd rather not guess which you'd prefer — happy to open a PR with a fixture and failing test against whichever shape you want.

#### Workaround

`"vite": false`, which costs nothing when entries are listed explicitly.

gioboa:
Closes #2016

Skip raw Vite glob reads during module analysis so they don’t suppress unused file and export findings.


packages/knip/src/plugins/vite/visitors/importMetaGlob.ts
```diff
@@ -26,6 +29,12 @@ export function createImportMetaGlobVisitor(ctx: PluginVisitorContext): PluginVi
 
       if (!patterns?.length) return;
 
+      const options = node.arguments[1];
+      if (options?.type === 'ObjectExpression') {
+        const query = getStringValue(findProperty(options, 'query'));
+        if (getStringValue(findProperty(options, 'as')) === 'raw' || (query && RAW_QUERY_RE.test(query))) return;
```
webpro: [P2] Preserve file references from raw globs

Returning here discards the file reference along with export usage. A TypeScript file consumed only by `import.meta.glob('./raw/*.ts', { query: '?raw', eager: true, import: 'default' })` is now reported unused, even though Vite embeds its contents in the bundle. The equivalent direct raw import preserves the reference. Record raw matches as referenced files without marking their exports used; the fixture should not expect a consumed raw file to be unused.

packages/knip/src/plugins/vite/visitors/importMetaGlob.ts
```diff
@@ -26,6 +29,12 @@ export function createImportMetaGlobVisitor(ctx: PluginVisitorContext): PluginVi
 
       if (!patterns?.length) return;
 
+      const options = node.arguments[1];
+      if (options?.type === 'ObjectExpression') {
+        const query = getStringValue(findProperty(options, 'query'));
```
webpro: [P2] Handle Vite's other supported raw-query forms

Vite accepts `query: 'raw'` and `query: { raw: '' }`, normalizing both to `?raw`. This guard misses both, so either spelling still marks every matched file as an entry and suppresses unused-export findings. Normalize these supported forms before classifying the query and cover them in the regression test. See [Vite's query normalization](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/importMetaGlob.ts#L191-L232).

webpro (changes requested):
Thanks for working on this fix and adding the regression test. Two cases still need attention: preserving references to raw files and recognizing Vite's other supported raw-query forms. Details inline.

packages/knip/src/plugins/vite/visitors/importMetaGlob.ts
gioboa: Fixed with [07b05f6](https://github.com/webpro-nl/knip/pull/2017/commits/07b05f6242007b2d78236e80baf34a6c2805b63b)

packages/knip/src/plugins/vite/visitors/importMetaGlob.ts
gioboa: fixed with [07b05f6](https://github.com/webpro-nl/knip/pull/2017/commits/07b05f6242007b2d78236e80baf34a6c2805b63b)

webpro:
Thanks again! 😃

webpro:

PR #2018, opened 2026-09-06
fix: throw when a sync compiler returns a promise

### LINKED ISSUE #1906: Compiler returning a Promise without `async` is treated as sync, its Promise ends up as the file source text
### Prerequisites

- [x] I'm using the latest version (reproduced on `main`, v6.29.0)

### Reproduction

No StackBlitz — the reproduction is a fixture in this repo, which seemed more direct. Two fixtures, identical except for one `async` keyword.

`fixtures/compilers/promise-sync/`:

```jsonc
// package.json
{ "name": "@fixtures/compilers-promise-sync", "dependencies": { "dep": "*" } }
```
```ts
// knip.ts
export default { compilers: { foo: (text: string) => Promise.resolve(text) } };
```
```ts
// index.ts
import './styles.foo';
```
```
// styles.foo
import _$0 from 'dep';
```

`fixtures/compilers/promise-async/` is the same with `foo: async (text: string) => text`.

| compiler | result |
| --- | --- |
| `async (text) => text` | `dep` correctly seen as used — 0 issues |
| `(text) => Promise.resolve(text)` | `dep` reported as an unused dependency |

### Description of the issue

A compiler that returns a Promise without being declared `async` is classified as sync, and its Promise object ends up cached as the file's source text.

`isAsyncCompiler` only checks the function constructor:

```ts
// src/compilers/index.ts:17
// TODO This does not detect functions returning a promise (just the async keyword)
const isAsyncCompiler = (fn?: CompilerSync | CompilerAsync) =>
  fn ? fn.constructor.name === 'AsyncFunction' : false;
```

so `partitionCompilers` routes it into `syncCompilers`, and `SourceFileManager.readFile` calls it without awaiting:

```ts
// src/typescript/SourceFileManager.ts:37-39
const compiled = compiler ? compiler(contents, filePath) : contents;
this.sourceTextCache.set(filePath, compiled);
```

TypeScript then receives a `Promise` object as file contents. Every import in that file disappears and the dependencies it referenced are reported unused — no error, no warning.

What makes this easy to walk into is that the public type allows both forms:

```ts
// src/compilers/types.ts:4
export type CompilerAsync = (source: string, path: string) => Promise<string>;
```

TypeScript doesn't distinguish `async (t) => t` from `(t) => Promise.resolve(t)`, so a correctly typed compiler silently breaks depending on how it's written. Delegating compilers like `(text) => transpiler.transform(text)` land on the broken side.

I see three directions and didn't want to guess:

1. Detect a thenable on first call and reclassify the extension as async — correct in all cases, but `readFile` is synchronous by design, so this is the invasive one.
2. Throw a clear error from `readFile` when the compiler returns a non-string: *"compiler for `.foo` returned a Promise; declare it `async` or use `asyncCompilers`"*. Doesn't fix it, but turns a silent false positive into something actionable. My preference.
3. Document the limitation on the compilers page.

`asyncCompilers` is of course a valid workaround, but `compilers` advertises auto-detection and fails quietly here.

Happy to send a PR with the fixture and a regression test once you've picked a direction.

MaxFreedomPollard:
Closes #1906, option 2 from the issue.

A compiler that returns a promise without the `async` keyword is classified as sync, because `isAsyncCompiler` in `src/compilers/index.ts` only checks `fn.constructor.name === 'AsyncFunction'`. `SourceFileManager.readFile` then puts the returned Promise object into `sourceTextCache` as the file's source text, so TypeScript sees no imports in that file and everything it referenced is reported unused. Nothing is logged, and `CompilerAsync` accepts both shapes, so a delegating compiler like `text => transpiler.transform(text)` lands on the broken side while looking correctly typed.

`readFile` now throws a `ConfigurationError` when a sync compiler returns a non-string:

```
ERROR: Compiler for .foo did not return a string (…/styles.foo), a compiler returning a promise must use the `async` keyword or be set in `asyncCompilers`
```

The check is inside the branch that already runs only when the extension has a compiler, so files without one are untouched.

`fixtures/compilers/promise-sync` is the reproduction from the issue: a `.foo` compiler written as `(text: string) => Promise.resolve(text)`, and a `.foo` file importing `dep`. On main, `knip` in that fixture exits 0 and reports `dep` as an unused dependency; with the change it reports the error above. Making the compiler `async` keeps the fixture clean either way, which is the whole trap.

Also added a line to the compilers page saying the `async` keyword is what Knip goes by.

Verification: `node --test test/compilers/promise-sync.test.ts` fails on main with "Missing expected rejection" and passes with the fix; `node --test 'test/compilers/*.test.ts'` 19/19; `pnpm run --dir packages/knip test` 1100/1100; `pnpm run ci`, `pnpm build`, `node ./src/cli.ts --directory ../..` and the same with `--production --strict` all clean on Node 22.

### DIFF
--- a/packages/docs/src/content/docs/features/compilers.md
+++ b/packages/docs/src/content/docs/features/compilers.md
@@ -80,7 +80,9 @@ The compiler function interface is straightforward. Text in, text out:
 (source: string, filename: string) => string;
 ```
 
-This may also be an `async` function.
+This may also be an `async` function. Knip uses the `async` keyword to tell sync
+and async compilers apart, so a regular function that returns a promise throws
+an error.
 
 ### Examples
 
--- a/packages/knip/fixtures/compilers/promise-sync/index.ts
+++ b/packages/knip/fixtures/compilers/promise-sync/index.ts
@@ -1,0 +1,1 @@
+import './styles.foo';
--- a/packages/knip/fixtures/compilers/promise-sync/knip.ts
+++ b/packages/knip/fixtures/compilers/promise-sync/knip.ts
@@ -1,0 +1,5 @@
+export default {
+  compilers: {
+    foo: (text: string) => Promise.resolve(text),
+  },
+};
--- a/packages/knip/fixtures/compilers/promise-sync/package.json
+++ b/packages/knip/fixtures/compilers/promise-sync/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "@fixtures/compilers-promise-sync",
+  "dependencies": {
+    "dep": "*"
+  }
+}
--- a/packages/knip/fixtures/compilers/promise-sync/styles.foo
+++ b/packages/knip/fixtures/compilers/promise-sync/styles.foo
@@ -1,0 +1,1 @@
+import 'dep';
--- a/packages/knip/src/typescript/SourceFileManager.ts
+++ b/packages/knip/src/typescript/SourceFileManager.ts
@@ -2,6 +2,7 @@ import { readFileSync } from 'node:fs';
 import type { AsyncCompilers, SyncCompilers } from '../compilers/types.ts';
 import { FOREIGN_FILE_EXTENSIONS } from '../constants.ts';
 import { debugLog } from '../util/debug.ts';
+import { ConfigurationError } from '../util/errors.ts';
 import { extname, isInternal } from '../util/path.ts';
 
 interface SourceFileManagerOptions {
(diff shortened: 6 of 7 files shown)


webpro:
Closing this after a good think about the ideal design. #1906 did not contain the option to make source loading support asynchronous results end to end with a simpler compiler contract based on returned values. And I didn't think it through better myself at the time. We could go back & forth in this PR but it's just too far off at this point and this PR is from a coding agent anyway, so here's hoping closing this right away is cool with you. Thanks!

MaxFreedomPollard:
For sure, sounds good.

PR #2025, opened 2026-09-09
Explain ambiguous star exports in traces

gioboa:
Tracing an ambiguous star export now lists its competing definitions instead of reporting that no export was found. Resolution follows binding identity through aliases, nested re-exports, namespace re-exports, and cycles, while explicit exports take precedence and stars exclude default exports.

The graph retains local binding names so declarations that copy imported values remain distinct from binding-forwarding re-exports. Origin discovery runs only when tracing requests it; ordinary unused-export analysis remains conservative.

Verified with package QA (1,166 passing tests, 7 skipped), repository checks, 15 focused Node tests, and a graph test that bounds traversal through repeated diamonds.

### DIFF
--- a/packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
+++ b/packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
@@ -90,23 +90,13 @@ export const createExportOriginResolver = (graph: ModuleGraph) => {
         for (const [sourcePath, imports] of node.imports.internal) {
           if (getPassThroughReExportSources(imports, identifier)) {
             addResolution(
-              resolve(
-                sourcePath,
-                identifier,
-                isTypeOnly || isTypeOnlyEdge(node, sourcePath, identifier),
-                seen
-              )
+              resolve(sourcePath, identifier, isTypeOnly || isTypeOnlyEdge(node, sourcePath, identifier), seen)
             );
           }
           forEachAliasReExport(imports, (sourceId, alias) => {
             if (alias !== identifier) return;
             addResolution(
-              resolve(
-                sourcePath,
-                sourceId,
-                isTypeOnly || isTypeOnlyEdge(node, sourcePath, sourceId),
-                seen
-              )
+              resolve(sourcePath, sourceId, isTypeOnly || isTypeOnlyEdge(node, sourcePath, sourceId), seen)
             );
           });
         }
@@ -126,12 +116,7 @@ export const createExportOriginResolver = (graph: ModuleGraph) => {
       for (const [sourcePath, imports] of node.imports.internal) {
         if (!getStarReExportSources(imports)) continue;
         addResolution(
-          resolve(
-            sourcePath,
-            identifier,
-            isTypeOnly || isTypeOnlyEdge(node, sourcePath, IMPORT_STAR),
-            seen
-          )
+          resolve(sourcePath, identifier, isTypeOnly || isTypeOnlyEdge(node, sourcePath, IMPORT_STAR), seen)
         );
       }
     }
--- a/packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
+++ b/packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
@@ -1,0 +1,145 @@
+import { IMPORT_STAR, SYMBOL_TYPE } from '../../constants.ts';
+import type { FileNode, ModuleGraph } from '../../types/module-graph.ts';
+import {
+  forEachAliasReExport,
+  getNamespaceReExportSources,
+  getPassThroughReExportSources,
+  getStarReExportSources,
+} from '../visitors.ts';
+
+export type ExportNamespace = 'type' | 'value';
+
+export interface ExportOrigin {
+  filePath: string;
+  identifier: string;
+  namespaces: ExportNamespace[];
+}
+
+export interface ExportOriginResolution {
+  origins: ExportOrigin[];
+  hasExplicitExport: boolean;
+}
+
+const getNamespaces = (type: string, isTypeOnly: boolean): ExportNamespace[] => {
+  if (isTypeOnly || type === SYMBOL_TYPE.TYPE || type === SYMBOL_TYPE.INTERFACE) return ['type'];
+  if (type === SYMBOL_TYPE.CLASS || type === SYMBOL_TYPE.ENUM || type === SYMBOL_TYPE.NAMESPACE)
+    return ['type', 'value'];
+  return ['value'];
+};
+
+const isTypeOnlyEdge = (node: FileNode, sourcePath: string, identifier: string) => {
+  let isMatch = false;
+  let isTypeOnly = true;
+  for (const item of node.imports.imports) {
+    if (item.filePath !== sourcePath || item.identifier !== identifier) continue;
+    isMatch = true;
+    isTypeOnly &&= item.isTypeOnly;
+  }
+  return isMatch && isTypeOnly;
+};
+
+export const createExportOriginResolver = (graph: ModuleGraph) => {
+  const cache = new Map<string, ExportOriginResolution>();
+
+  const resolve = (
+    filePath: string,
+    identifier: string,
+    isTypeOnly = false,
+    seen = new Set<string>()
+  ): ExportOriginResolution => {
+    const key = `${filePath}:${identifier}:${isTypeOnly}`;
+    const isRoot = seen.size === 0;
+    const cached = isRoot ? cache.get(key) : undefined;
+    if (cached) return cached;
+    if (seen.has(key)) return { origins: [], hasExplicitExport: false };
+    seen.add(key);
+
+    const node = graph.get(filePath);
+    if (!node) return { origins: [], hasExplicitExport: false };
+
+    const origins = new Map<string, ExportOrigin>();
+    const addOrigin = (origin: ExportOrigin, typeOnly = isTypeOnly) => {
+      const namespaces: ExportNamespace[] = typeOnly ? ['type'] : origin.namespaces;
+      const originKey = `${origin.filePath}:${origin.identifier}`;
+      const existing = origins.get(originKey);
+      if (!existing) {
+        origins.set(originKey, { ...origin, namespaces: [...namespaces] });
+        return;
+      }
+      for (const namespace of namespaces)
+        if (!existing.namespaces.includes(namespace)) existing.namespaces.push(namespace);
+    };
+    const addResolution = (resolution: ExportOriginResolution, typeOnly = isTypeOnly) => {
+      for (const origin of resolution.origins) addOrigin(origin, typeOnly);
+    };
+    const exp = node.exports.get(identifier);
+    let hasExplicitExport = !!exp;
+
+    for (const [sourcePath, imports] of node.imports.internal) {
+      if (!getNamespaceReExportSources(imports, identifier)) continue;
+      hasExplicitExport = true;
+      addOrigin({
+        filePath: sourcePath,
+        identifier: IMPORT_STAR,
+        namespaces: isTypeOnlyEdge(node, sourcePath, identifier) ? ['type'] : ['type', 'value'],
+      });
+    }
+
+    if (exp) {
+      if (exp.isBindingReExport) {
+        for (const [sourcePath, imports] of node.imports.internal) {
+          if (getPassThroughReExportSources(imports, identifier)) {
+            addResolution(
+              resolve(
+                sourcePath,
+                identifier,
+                isTypeOnly || isTypeOnlyEdge(node, sourcePath, identifier),
+                seen
+              )
+            );
+          }
+          forEachAliasReExport(imports, (sourceId, alias) => {
+            if (alias !== identifier) return;
+            addResolution(
+              resolve(
+                sourcePath,
+                sourceId,
+                isTypeOnly || isTypeOnlyEdge(node, sourcePath, sourceId),
+                seen
+              )
+            );
+          });
+        }
+      }
+      if (origins.size === 0 && !exp.isBindingReExport) {
+        addOrigin({ filePath, identifier: exp.binding, namespaces: getNamespaces(exp.type, isTypeOnly) });
+      }
+    }
+
+    if (hasExplicitExport) {
+      const resolution = { origins: [...origins.values()], hasExplicitExport };
+      if (isRoot) cache.set(key, resolution);
+      return resolution;
+    }
+
+    if (identifier !== 'default') {
+      for (const [sourcePath, imports] of node.imports.internal) {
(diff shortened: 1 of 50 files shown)


webpro (changes requested):
Explaining why a barrel export cannot resolve is useful, and `getAmbiguousStarExport` is a good name for this query. Before merging, I'd like to tighten the resolution semantics and extract shared origin lookup in the graph explorer.

The editor's `getContention` answers a broader question: it finds branching and multiple definitions throughout a re-export network. That includes valid shadowing, where an explicit export selects a definition. The ambiguity query should stay separate, with both operations able to use the same underlying origin resolution. Keep `getAmbiguousStarExport` small and reuse the existing re-export visitors for traversal.

There are four correctness cases to address:

1. `export *` never forwards `default`. A barrel with two default-exporting sources currently reports an ambiguous default export. It should report no such export.
2. Identity needs to be the defining module and local binding. Two exported aliases of one local binding can converge through the same barrel name without ambiguity. Distinct bindings in one file must remain distinct. Namespace re-exports of the same module should also converge.
3. `isReExport` also represents conservative usage relationships. For example, `import { fruit } from './origin'; export const same = fruit;` creates a new local binding. Two modules doing this have competing `same` exports even though they copy the same imported value. Keep binding identity separate from the existing usage flags.
4. The type/value partition misses conflicts: TypeScript reports TS2308 when a barrel stars an interface from one source and a same-named value from another. Report the competing origins without asserting that the name is absent from every namespace.

Please add focused regressions for these cases, alongside explicit-export precedence and cycle/diamond handling. Preserve the existing conservative unused-export behavior.

The current on-demand trace call is the right boundary. Editor contention is calculated for the open file; CLI tracing can involve many files. Shared origin resolution should avoid repeating convergent paths within a query, and ordinary CLI analysis must not acquire contention scans.

Keep the editor's `conflict` classification and full contention details in traces as separate changes. They should use the same origin facts to distinguish branching, shadowing with a winner, and ambiguity, with results reused across displayed exports.

gioboa (commented):
[0f84afb](https://github.com/webpro-nl/knip/pull/2025/commits/0f84afb4f75959c92ea5b0808f243790d215e682) is addressing the comments.

packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+      for (const namespace of namespaces)
+        if (!existing.namespaces.includes(namespace)) existing.namespaces.push(namespace);
+    };
+    const addResolution = (resolution: ExportOriginResolution, typeOnly = isTypeOnly) => {
+      for (const origin of resolution.origins) addOrigin(origin, typeOnly);
+    };
+    const exp = node.exports.get(identifier);
+
+    if (exp) {
```
webpro, line 52: P2: `export * as fruit` lives in `reExportNs`, without a `node.exports` entry, so this guard skips it. Adding two starred `fruit` definitions then falsely reports ambiguity despite the explicit namespace export; distinct namespace origins are also missed. Resolve namespace edges independently and preserve their explicit precedence.

packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
```diff
@@ -0,0 +1,145 @@
...
+            addOrigin({
+              filePath: sourcePath,
+              identifier: IMPORT_STAR,
+              namespaces: isTypeOnlyEdge(node, sourcePath, identifier) ? ['type'] : ['type', 'value'],
+            });
+          }
+        }
+      }
+      if (origins.size === 0 && !exp.isBindingReExport) {
+        addOrigin({ filePath, identifier: exp.binding, namespaces: getNamespaces(exp.type, isTypeOnly) });
```
webpro: P2: `handleExportDefault` still records binding `default` for `export default function apple() {}; export { apple };`. Re-exporting both names as `fruit` through separate barrels falsely reports two origins, although Node resolves one binding. Record the local name for named default declarations; expression defaults such as `export default apple` must remain distinct.

packages/knip/src/graph-explorer/operations/get-contention.ts
```diff
@@ -45,6 +48,11 @@ const getContentionForIdentifier = (
     }
   }
 
+  for (const filePath of network.files) {
```
webpro: P2: This queries the initial name after aliases change it. If `middle.ts` re-exports `origin.source as renamed` and `origin.ts` also re-exports an unrelated `renamed`, the editor falsely adds that unrelated definition as a conflict. Repeated origin walks also make an 81-file star chain grow from 324 to 3,645 graph reads. Please defer this loop to the editor change, preserving file/name pairs and sharing traversal there.

webpro (changes requested):
The shared resolver and lazy trace query address the ownership concern. Three issues remain below.

packages/knip/src/graph-explorer/operations/get-contention.ts
gioboa: Fixed by  [fee2206](https://github.com/webpro-nl/knip/pull/2025/commits/fee22062e4e7fbb2de13f70a935e433a47807128)

packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
gioboa: Fixed by  [fee2206](https://github.com/webpro-nl/knip/pull/2025/commits/fee22062e4e7fbb2de13f70a935e433a47807128)

packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
gioboa: Thanks, Fixed by  [fee2206](https://github.com/webpro-nl/knip/pull/2025/commits/fee22062e4e7fbb2de13f70a935e433a47807128)

packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
```diff
@@ -0,0 +1,130 @@
...
+        if (!existing.namespaces.includes(namespace)) existing.namespaces.push(namespace);
+    };
+    const addResolution = (resolution: ExportOriginResolution, typeOnly = isTypeOnly) => {
+      for (const origin of resolution.origins) addOrigin(origin, typeOnly);
+    };
+    const exp = node.exports.get(identifier);
+    let hasExplicitExport = !!exp;
+
+    for (const [sourcePath, imports] of node.imports.internal) {
+      if (!getNamespaceReExportSources(imports, identifier)) continue;
```
webpro, line 47: P2: Two modules containing `import * as source from './origin'; export const fruit = source` now collapse into `origin:*`, hiding their ambiguity. `reExportNs` includes these copied values even though `isBindingReExport` is false. This worked at the previous reviewed head. Preserve the local binding for copies while retaining explicit namespace precedence.

packages/knip/src/typescript/visitors/exports.ts
```diff
@@ -272,20 +284,24 @@ export function handleExportDefault(node: ExportDefaultDeclaration, s: WalkState
   let type: SymbolType = SYMBOL_TYPE.UNKNOWN;
   let pos = decl.start;
   let members: ExportMember[] = [];
+  let binding = 'default';
 
   if (decl.type === 'FunctionDeclaration') {
     type = SYMBOL_TYPE.FUNCTION;
     pos = decl.id?.start ?? decl.start;
+    binding = decl.id?.name ?? binding;
```
webpro, line 292: P2: Named default functions in `.d.ts` files still get binding `default`: Oxc represents `export default function apple(): void` as `TSDeclareFunction`. Exporting `apple` too and forwarding both names as `fruit` through separate starred barrels falsely reports two origins. The fixture passes TypeScript. Handle `TSDeclareFunction` alongside `FunctionDeclaration` when recording the local binding.

packages/knip/src/reporters/trace.ts
```diff
@@ -75,6 +75,24 @@ export default ({ graph, explorer, options, workspaceFilePathFilter, issues }: T
     const toRel = (path: string) => toRelative(path, options.cwd);
 
     if (nodes.length === 0) {
+      const traceFile = options.traceFile;
+      const traceExport = options.traceExport;
+      const collision = traceFile && traceExport ? explorer.getAmbiguousStarExport(traceFile, traceExport) : undefined;
+      if (collision) {
+        collision.origins.sort((a, b) => compareStrings(a.filePath, b.filePath));
+        const namespaces = collision.namespaces.join(' and ');
+        const namespaceLabel =
```
webpro: P2: Local export specifiers can produce incorrect namespace labels. `import { Source } from './source'; export type { Source as fruit }` retains a class's value namespace, while locally exported interface aliases are labeled value-only. Please omit this qualifier with the compact format above; that avoids extending namespace inference for this PR.

webpro (changes requested):
Explicit namespace precedence now works, and the editor contention loop is gone. The remaining binding and namespace-labeling issues are inline.

Please use this compact terminal format and omit the type/value qualifier:

```text
barrel.ts:foo [ambiguous]
├── fruits.ts:foo
└── vegetables.ts:foo
```

packages/knip/src/reporters/trace.ts
gioboa: [997cb95](https://github.com/webpro-nl/knip/pull/2025/commits/997cb9529c72343c9cf52bb71e9040cbce8b86b4) is fixing this. Thank you.

packages/knip/src/typescript/visitors/exports.ts
gioboa: [997cb95](https://github.com/webpro-nl/knip/pull/2025/commits/997cb9529c72343c9cf52bb71e9040cbce8b86b4) is fixing this. Thank you.

packages/knip/src/graph-explorer/operations/resolve-export-origins.ts
gioboa: [997cb95](https://github.com/webpro-nl/knip/pull/2025/commits/997cb9529c72343c9cf52bb71e9040cbce8b86b4) is fixing this. Thank you.

packages/knip/src/typescript/visitors/exports.ts
```diff
@@ -250,7 +260,9 @@ export function handleExportNamed(node: ExportNamedDeclaration, s: WalkState) {
         [],
         fix as Fix,
         isReExport,
-        s.getJSDocTags(node.start)
+        s.getJSDocTags(node.start),
+        localName,
+        isReExport
```
webpro: P2: Keep dynamic-import result variables as local bindings. [exports.ts:265](https://github.com/webpro-nl/knip/blob/29c9767f1dcf6d4c3728b3e2b01e3f660937a052/packages/knip/src/typescript/visitors/exports.ts#L265) sets `isBindingReExport` for every matching `localImportMap` entry, including variables initialized with `await import()`. Two modules containing `const source = await import('./origin.ts'); export { source as fruit };` incorrectly converge at `origin.ts:*` when starred from a barrel. The trace reports no export; Node reports conflicting star exports and TypeScript reports TS2308. Use the existing `isDynamicImport` marker to preserve local binding identity without changing conservative usage metadata.

packages/knip/src/reporters/trace.ts
```diff
@@ -75,6 +75,19 @@ export default ({ graph, explorer, options, workspaceFilePathFilter, issues }: T
     const toRel = (path: string) => toRelative(path, options.cwd);
 
     if (nodes.length === 0) {
+      const traceFile = options.traceFile;
```
webpro: P2: Check exact-export ambiguity before searching member names. [trace.ts:78](https://github.com/webpro-nl/knip/blob/29c9767f1dcf6d4c3728b3e2b01e3f660937a052/packages/knip/src/reporters/trace.ts#L78) checks ambiguity only after the member fallback. If a barrel stars two `fruit` definitions and exports `enum Basket { fruit = 1 }`, `--trace-export fruit` traces `Basket` instead of reporting the ambiguous top-level export. The member match fills `nodes`, so the collision query never runs. Give the exact ambiguous export precedence; `Basket.fruit` can still use the member fallback.

packages/knip/src/reporters/trace.ts
```diff
@@ -75,6 +75,19 @@ export default ({ graph, explorer, options, workspaceFilePathFilter, issues }: T
...
+      const traceExport = options.traceExport;
+      const collision = traceFile && traceExport ? explorer.getAmbiguousStarExport(traceFile, traceExport) : undefined;
```
webpro: P2: Check the namespace root for dotted trace queries. [trace.ts:80](https://github.com/webpro-nl/knip/blob/29c9767f1dcf6d4c3728b3e2b01e3f660937a052/packages/knip/src/reporters/trace.ts#L80) passes the original dotted query to the collision resolver, although the earlier trace fallback also tries its namespace root. With competing `Produce` namespace exports, tracing `Produce` correctly lists both origins, but `Produce.value` reports no export. Apply the collision check to the namespace root too, retaining the resolved name for the compact output.

webpro (changes requested):
Thanks for the improvements! Let's pull this one over the finish line. All three previous findings are addressed, and the trace now uses the selected compact format. Three P2 cases remain on `29c9767f1dcf6d4c3728b3e2b01e3f660937a052`:

gioboa (commented):
> Let's pull this one over the finish line. 

Thanks, fixed with [4b8007b](https://github.com/webpro-nl/knip/pull/2025/commits/4b8007b847bf0b2d441731ad8124b4eb64558c64)

webpro (approved):
Thanks for the quick fixes. All three findings from the previous round are resolved on `4b8007b847bf0b2d441731ad8124b4eb64558c64`, and the reordered trace reporter introduces no regressions. Ready to merge.

Dynamic-import result variables keep their local binding identity: starring two modules containing `const source = await import('./origin'); export { source as dynamicCopy }` lists both `source` bindings. The ambiguity check runs before the exports tree fallbacks, so an enum member named like an ambiguous top-level export no longer hides the collision, while `Basket.fruit` and `Basket` trace as before. Dotted queries resolve through their namespace root, so `Produce.vegetable` reports the competing `Produce` namespace exports.

Since the collision query now runs for every `--trace-export` with `--trace-file`, I compared the built CLI against main for 720 trace queries over the PR fixture plus synthetic barrels. Explicit exports, single-origin stars, dotted queries over converged namespaces, queries without `--trace-file`, and missing files produce identical output; the only differences are the intended `[ambiguous]` results. Origin resolution remains reachable only from the trace reporter.

Verification: package QA, repository checks, GitHub checks, and the reproductions from all earlier rounds pass; an independent review reported no findings.

webpro:
Thanks a bunch for sticking to it, Giorgio! 🧡

webpro:

PR #2026, opened 2026-09-09
Add Railway plugin

jonahsnider:
Adds a Railway plugin for their new IaC system.

https://docs.railway.com/infrastructure-as-code


packages/knip/src/plugins/railway/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const title = 'Railway';
+
+const enablers = 'This plugin is enabled when `.railway/railway.ts` is found.';
+
+const isEnabled: IsPluginEnabled = ({ cwd }) => isFile(cwd, '.railway/railway.ts');
+
+const entry = ['.railway/railway.ts'];
+
+const plugin: Plugin = {
```
webpro, line 14: P2: Discover CLI configuration and runner paths. A package script containing `railway config plan --file infra/provision.ts` leaves that file and its `railway` dependency reported unused. A custom `--runner tools/iac-runner.js` is also reported unused. Please handle these flags through the existing plugin argument API. [Railway CLI reference](https://docs.railway.com/cli/config).

packages/knip/src/plugins/railway/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  title,
+  enablers,
+  isEnabled,
+  entry,
```
webpro, line 18: P2: Extract references from service commands. `start: 'node src/serve.ts'` leaves the startup file reported unused, and `preDeploy: 'npx drizzle-kit migrate'` leaves `drizzle-kit` reported unused. Please handle `build`, `start`, `preDeploy`, and `deploy.preDeployCommand` through the existing script parser, respecting the service's source and root directory when resolving local references. [Railway IaC reference](https://docs.railway.com/infrastructure-as-code/reference).

webpro (changes requested):
Thanks for adding Railway support. The default entry works, but two documented sources of entries and dependencies are still missed:

If you'd like, I can take a pass at these additions.

jonahsnider:
Thanks for the review! Feedback should be addressed now.

packages/knip/src/plugins/railway/index.ts
```diff
@@ -0,0 +1,90 @@
...
+const title = 'Railway';
+
+const enablers = 'This plugin is enabled when `.railway/railway.ts` is found.';
+
+const isEnabled: IsPluginEnabled = ({ cwd }) => isFile(cwd, '.railway/railway.ts');
+
+const config = ['.railway/railway.ts'];
+
+const args: Args = {
+  resolve: ['file', 'runner'],
```
webpro: [P2] Custom `--file` skips command analysis. The file becomes an ordinary entry, so its startup file and migration dependency remain reported unused. Route `file` through `args.config`.

packages/knip/src/plugins/railway/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  const inputs: Input[] = [];
+  const visitor = new Visitor({
+    CallExpression(node) {
+      if (node.callee.type !== 'Identifier' || !serviceNames.has(node.callee.name)) return;
+
+      const serviceConfig = node.arguments[1];
+      if (serviceConfig?.type !== 'ObjectExpression') return;
+
+      const source = findProperty(serviceConfig, 'source');
+      const sourceConfig = source?.type === 'CallExpression' ? source.arguments[1] : source;
```
webpro, line 82: [P2] Shared source constants lose their root directory. With `source: repo`, the resolver ignores the constant's `rootDirectory` and resolves commands from the repository root. This pattern is emitted by [Railway's config-pull renderer](https://github.com/railwayapp/cli/blob/f60f3a77b980c47f1136909fbd9a443e29a2b95f/src/commands/config/mod.rs#L937-L1005). Resolve static source bindings before selecting the service root.

packages/knip/src/plugins/railway/index.ts
```diff
@@ -0,0 +1,90 @@
...
+      const commands = [
+        ...getCommands(build),
+        ...getCommands(findProperty(build, 'buildCommand')),
+        ...getCommands(findProperty(serviceConfig, 'start')),
+        ...getCommands(findProperty(serviceConfig, 'preDeploy')),
+        ...getCommands(findProperty(findProperty(serviceConfig, 'deploy'), 'preDeployCommand')),
+      ];
+
+      for (const input of options.getInputsFromScripts(commands, { knownBinsOnly: true, cwd, manifest })) {
+        inputs.push({ ...input, dir: cwd });
```
webpro, line 101: [P2] Relative executables resolve against the wrong directory. With `root: 'services/worker'` and `start: './scripts/serve.js'`, Knip marks the root-level script used and reports the actual service script unused. Normalize entry paths against the service root.

packages/knip/src/plugins/railway/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  const inputs: Input[] = [];
+  const visitor = new Visitor({
+    CallExpression(node) {
+      if (node.callee.type !== 'Identifier' || !serviceNames.has(node.callee.name)) return;
+
+      const serviceConfig = node.arguments[1];
+      if (serviceConfig?.type !== 'ObjectExpression') return;
+
+      const source = findProperty(serviceConfig, 'source');
```
webpro: [P2] External service commands hide unused local code. Commands for prebuilt images or other repositories incorrectly mark local files and dependencies used. Establish source ownership before extracting local references.

webpro (changes requested):
Thanks for the improvements. Almost there! Four cases still produce incorrect unused-code reports; details inline.

jonahsnider:
Thanks, just pushed changes to address everything:
- custom files are parsed as configs
- handle source bindings + ownership
- resolve commands from service root

packages/knip/src/plugins/railway/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  config: ['file'],
+  resolve: ['runner'],
+};
+
+const getCommands = (node: unknown): string[] => {
+  const command = getStringValue(node);
+  return command ? [command] : [...getStringValues(node)];
+};
+
+const isExternalSourceType = (type: string | undefined) => type === 'github' || type === 'image' || type === 'template';
```
webpro, line 31: [P2] GitHub sources in this repository are now skipped

`github('owner/repo')` can refer to the repository Knip is analyzing; [Railway generates this from the local Git origin](https://github.com/railwayapp/cli/blob/0d092aacc3330b418d2fd024531e7c01d369d799/src/commands/config/mod.rs#L1714-L1783). This new exclusion reports that service's startup file and command dependencies as unused, although they were found before this commit. Please avoid treating `github` alone as proof that the service is external.

webpro (changes requested):
Thanks for all the work on this! The custom-file and executable-path fixes look good. There's one regression in the new GitHub-source check.

packages/knip/src/plugins/railway/index.ts
jonahsnider: Should be fixed now, thanks for reviewing!

jonahsnider:
Feedback addressed! Fourth time's the charm? :)

PR #2028, opened 2026-09-12
Fix eslintrc `parserOptions.parser` handling in ESLint plugin

bytedoe:
The ESLint plugin resolved the parser of an eslintrc config with `config.parser ?? config.parserOptions?.parser`. Since `parserOptions.parser` only exists for wrapper parsers (vue/svelte/astro-eslint-parser), which always set `parser` as well, the fallback never fired and e.g. `@typescript-eslint/parser` was reported as unused.

vue-eslint-parser also documents an object form, `parserOptions: { parser: { js: 'espree', ts: '@typescript-eslint/parser' } }`. In an `overrides` entry (where `parser` is not set) that object was passed to `toDeferResolve` as-is and crashed Knip:

    TypeError: input.specifier.startsWith is not a function

### DIFF
--- a/packages/knip/src/plugins/eslint/helpers.ts
+++ b/packages/knip/src/plugins/eslint/helpers.ts
@@ -49,20 +49,24 @@ const getInputsDeprecated = (
   // const rules = getDependenciesFromRules(config.rules); // TODO enable in next major? Unexpected/breaking in certain cases w/ eslint v8
   const rules = getDependenciesFromRules({});
   const overrides = config.overrides ? [config.overrides].flat().flatMap(d => getInputsDeprecated(d, options)) : [];
-  const deferred = compact([...extendsSpecifiers, ...plugins, ...parsers, ...settings, ...rules]).map(id =>
-    toDeferResolve(id)
-  );
-  return [...extendConfigs, ...deferred, ...babelDependencies, ...overrides];
+  const deferred = compact([...extendsSpecifiers, ...plugins, ...settings, ...rules]).map(id => toDeferResolve(id));
+  return [...extendConfigs, ...deferred, ...parsers, ...babelDependencies, ...overrides];
 };
 
+const isParserObject = (value: Record<string, unknown>) =>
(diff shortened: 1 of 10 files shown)


packages/knip/src/plugins/eslint/helpers.ts
```diff
@@ -41,20 +41,30 @@ const getInputsDeprecated = (
...
+  const deferred = compact([...extendsSpecifiers, ...plugins, ...parsers, ...settings, ...rules]).map(id =>
     toDeferResolve(id)
   );
   return [...extendConfigs, ...deferred, ...babelDependencies, ...overrides];
 };
 
+const getParsers = ({ parser, parserOptions }: BaseConfig) => {
+  const parsers: string[] = [];
+  for (const value of [parser, parserOptions?.parser]) {
+    for (const name of value && typeof value === 'object' ? Object.values(value) : [value]) {
```
webpro: [P2] Parser objects produce bogus imports

With Vue and `parserOptions.parser: require('@typescript-eslint/parser')`, `Object.values()` extracts the parser's version. Knip reports unresolved import `8.70.0`. Recognize parser objects before enumerating language maps.

packages/knip/src/plugins/eslint/helpers.ts
```diff
@@ -41,20 +41,30 @@ const getInputsDeprecated = (
...
+      if (typeof name === 'string' && name !== 'espree') parsers.push(name);
```
webpro: [P2] Explicit Espree dependencies become unused

Filtering `espree` discards a dependency explicitly selected by `"parser": "espree"`. [ESLint resolves the installed package before its fallback](https://github.com/eslint/eslintrc/blob/main/lib/config-array-factory.js#L979-L1008). Preserve that reference, using optional handling where appropriate.

webpro (changes requested):
Thanks for the pull request!

Two valid parser configurations now produce false positives; details inline.

webpro:
Thank you! 🧡

webpro:

PR #2029, opened 2026-09-12
fix(compilers): require word boundary around import keyword

thanadolps:
The `<script>` import matcher matched `import` inside ordinary words, so `'this is important'` read as an import and produced a malformed fragment. The style import appended after it never registered, and the stylesheet was reported as unused.

Minimal repro:

```svelte
<script lang="ts">
  import Inner from './Inner.svelte'
  let tags = ['important', 'x']
</script>
<style lang="scss">
  @use './styles.scss';
</style>
```

Parsed imports and parse errors for that file:

- before: `["./Inner.svelte"]`, 1 error
- after: `["./Inner.svelte", "./styles.scss", "./_styles.scss"]`, 0 errors

`importMatcher` now requires a boundary on both sides, a lookbehind that rejects `.`, word characters, `$` and `#`, plus a trailing `\b`. That drops `important`, `reimport`, and `loader.import('./x')` while leaving real imports alone. The leading lookbehind is the same guard `dynamicImportsWithinTemplate` already uses in `compilers/svelte.ts`.

The matcher is shared, so this isn't Svelte-only. tsrx (`collectImports`) and Astro MDX (`plugins/astro/compiler-mdx.ts`) go through the same code, which is why the fix sits in `compilers/compilers.ts` and not a Svelte compiler override.


webpro:
Thanks for the PR! As long as we can keep it somewhat understandable, a regex or two is fine but this PR stretches that a bit. This looks like the point you should start using an actual compiler: https://knip.dev/features/compilers#svelte

WDYT? Any chance you could give the documented approach a shot?

thanadolps:
Fair, I cut the PR down to a boundary change and adjust PR description to more realistic example.

I'd keep the boundary check. It hits the common shape, imports at the top and a string like `'important'`. 
The override has its place, but imo a stray word like `important` shouldn't need it.

webpro:
Thanks for narrowing the change. The boundary check is a useful, focused fix for words such as `important`, identifiers such as `reimport`, and member access such as `loader.import(...)`.

The boundary-only matcher still has the usual limitations of this approach. It can match a standalone `import` inside a quoted string, template literal, regular-expression literal, or comment when that text resembles an import statement, for example:

```ts
const message = "import './not-a-module'";
// import './also-not-a-module'
```

Those cases can still produce a false dependency or unresolved import. The same shared matcher is used by the Svelte, tsrx, and Astro/MDX extractors, so the limitation applies there as well. Avoiding all of these cases would push the built-in extractor toward a lexer or a real parser, which is outside its intended scope; projects that need that fidelity can use the documented framework compiler override.

I think the smaller boundary fix is a reasonable improvement, so merging this as-is. Also I'm going to add `.agents/COMPILERS.md` to make it more explicit. Thanks!

webpro:

PR #2033, opened 2026-09-12
fix: fix trailing comma on dependency removal

### LINKED ISSUE #2030: Removing the last dependency with the editor quick fix leaves invalid JSON
Accepting Remove dependency for the final item in a dependency section with multiple entries makes package.json unparsable by JSON.parse and package managers. Both dependency sections offered by the language server are affected. This does not concern the CLI's separate dependency fixer.

At `packages/language-server/src/code-actions.js:40`:

createRemoveDependencyEdit computes a whole-line deletion: 'const range = { start: { line: lineIndex, character: 0 }, end: { line: lineIndex + 1, character: 0 } };'. It never removes the comma after the preceding dependency. server.js:431-438 offers this edit for dependencies and devDependencies. PackagePeeker.ts:41-45 supplies the correct one-based line for a normal multiline manifest, so this path is reachable without fabricated positions. Root cause: a JSON property removal is implemented as a line removal without accounting for the object's separators.

## Proposed fix

Replace whole-line deletion in createRemoveDependencyEdit with property-aware removal that accounts for an adjacent comma and preserves unrelated text. Do not reformat the whole manifest. Add packages/language-server/test/code-actions.test.js using the existing node:test and assert conventions and TextDocument.applyEdits. Assert JSON.parse succeeds and only the selected key is removed for first, middle, last, and sole entries in dependencies and devDependencies, including CRLF. The last-entry case fails on the frozen code and passes after the separator handling is corrected.

## How to see it

```
Static trace, not executed. Use package.json text '{
  "dependencies": {
    "react": "^19.0.0",
    "lodash": "^4.17.21"
  }
}
', with react referenced and lodash unused. Open the manifest and accept Remove dependency for lodash. PackagePeeker returns line 4; createRemoveDependencyEdit deletes from line 3 character 0 to line 4 character 0 in zero-based LSP coordinates. The react line retains its comma immediately before the closing brace. JSON.parse of the resulting document throws. Expected: valid JSON containing react and no lodash, with all unrelated fields preserved.
```

If this looks right I can push `fix/dependency-quickfix-json-separators` on `anxkhn/knip` instead of opening a pull request first.

gioboa:
Closes #2030

Removing the last entry in dependencies/devDependencies via the editor quick fix left the previous line's trailing comma, producing invalid JSON. The edit now also strips that comma when the removed line has none. Adds tests for first/middle/last/sole removal with LF and CRLF.

### DIFF
--- a/packages/language-server/src/code-actions.js
+++ b/packages/language-server/src/code-actions.js
@@ -40,6 +40,15 @@ export const createRemoveDependencyEdit = (document, uri, issue) => {
     const range = { start: { line: lineIndex, character: 0 }, end: { line: lineIndex + 1, character: 0 } };
     const edits = [TextEdit.del(range)];
 
+    if (lineIndex > 0 && !document.getText(range).trimEnd().endsWith(',')) {
+      const prevRange = { start: { line: lineIndex - 1, character: 0 }, end: { line: lineIndex, character: 0 } };
+      const prevLine = document.getText(prevRange).trimEnd();
+      if (prevLine.endsWith(',')) {
+        const start = { line: lineIndex - 1, character: prevLine.length - 1 };
+        edits.push(TextEdit.del({ start, end: { line: lineIndex - 1, character: prevLine.length } }));
+      }
+    }
+
     return { changes: { [uri]: edits } };
   } catch (_error) {
     return null;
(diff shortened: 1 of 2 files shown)


webpro:
Thanks a bunch! 👌

webpro:

PR #2034, opened 2026-09-12
fix: recover from corrupt cache file

### LINKED ISSUE #2031: A cache read failure replaces the empty cache with undefined and crashes analysis
A truncated cache left by an interrupted write makes subsequent --cache runs crash while examining any existing source or configuration file, until the user deletes the cache. All CacheConsultant clients use this shared implementation.

At `packages/knip/src/util/file-entry-cache.ts:42`:

The field starts as 'cache = new Map<string, MetaData<T>>()', but the constructor unconditionally assigns 'this.cache = create(this.filePath)' when the cache file exists. createCache catches read/deserialization errors at lines 28-30 and returns undefined. getFileDescriptor then executes 'let meta = this.cache.get(filePath)' at line 62. The existing catch therefore does not recover from a cache read failure. Root cause: the constructor discards its valid empty Map even when loading returned no cache.

## Proposed fix

Preserve the initialized empty Map when create returns undefined, or make the caught load failure return an empty Map. Extend packages/knip/test/util/file-entry-cache.test.ts with a truncated-cache regression that asserts getFileDescriptor does not throw, changed is true, and a new payload survives reconcile and reload. Keep this to error recovery, without redesigning persistence or adding unrelated validation.

## How to see it

Static trace, not executed. Extend the existing file-entry-cache test setup with an existing source file and an empty cache file at join(cacheLocation, 'compiler'). Construct new FileEntryCache('compiler', cacheLocation), then call getFileDescriptor(sourcePath). Deserializing the empty cache enters createCache's catch and returns undefined; the constructor assigns undefined; statSync succeeds for the source; the first this.cache.get call throws TypeError. Expected: an empty in-memory cache, a changed descriptor, and successful persistence/reload after reconciliation. Also cover a cache read error through a focused stub if the test runtime supports it.

If this looks right I can push `fix/cache-load-error-recovery` on `anxkhn/knip` instead of opening a pull request first.

gioboa:
Closes #2031

A truncated or corrupt cache file (e.g. from an interrupted write) made createCache return undefined, overwriting the initial empty Map and crashing getFileDescriptor with a TypeError. Now falls back to the empty cache so analysis proceeds and the cache is rewritten on reconcile.

### DIFF
--- a/packages/knip/src/util/file-entry-cache.ts
+++ b/packages/knip/src/util/file-entry-cache.ts
@@ -39,7 +39,7 @@ export class FileEntryCache<T> {
 
   constructor(cacheId: string, _path: string) {
     this.filePath = path.resolve(_path, cacheId);
-    if (isFile(this.filePath)) this.cache = create(this.filePath);
+    if (isFile(this.filePath)) this.cache = create(this.filePath) ?? this.cache;
   }
 
   getFileDescriptor(filePath: string): FileDescriptor<T> {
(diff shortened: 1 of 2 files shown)


webpro:
Thanks! 🧡

webpro:

PR #2035, opened 2026-09-12
fix: fix language server bundled knip fallback

### LINKED ISSUE #2032: Standalone language server cannot fall back to its bundled Knip under Node.js
Users launching the documented standalone @knip/language-server command against a project without a locally resolvable knip/session receive no analysis or diagnostics. The server logs an error and remains without a session even though its own knip dependency is available. The VS Code bundled build may behave differently and is not necessary to trigger this defect.

At `packages/language-server/src/server.js:240`:

The fallback path evaluates 'createRequire(__filename).resolve("knip/session")' before returning the statically imported createOptions and createSession. This package declares type: module in package.json:12, and cli.js directly imports index.js, which directly imports server.js without a CommonJS wrapper. There is no declaration of __filename in server.js. When project-local resolution at lines 232-237 fails, evaluating this logging expression raises ReferenceError. start catches it at lines 270-272 before options or a session are created. Root cause: CommonJS-only __filename is used in the native ESM fallback path.

## Proposed fix

Use createRequire(import.meta.url) in the fallback expression. Add a focused native Node.js stdio startup test in packages/language-server/test/server.test.js that initializes the real CLI with a dependency-isolated temporary project, handles configuration requests, and waits for the module-graph-built notification with a bounded timeout and cleanup. Assert the fallback log occurs and the ReferenceError does not. Do not mock the fallback resolver or run through a CommonJS bundle, because either would hide the regression.

## How to see it

Static trace, not executed. Launch packages/language-server/src/cli.js using Node.js with --stdio against a temporary project outside the checkout's dependency ancestry, containing package.json and index.js but no local knip. Complete LSP initialize with that workspace URI and reply to workspace/configuration with a configuration that permits startup. On initialized, start calls #resolveKnipSession. Project-relative require.resolve fails and is caught; the fallback logging expression evaluates undefined __filename and throws. Expected: a 'Using bundled knip' log and module-graph-built notification. Current trace instead logs 'Error: ReferenceError: __filename is not defined' and leaves session undefined.

If this looks right I can push `fix/language-server-esm-fallback` on `anxkhn/knip` instead of opening a pull request first.

gioboa:
Closes #2032

The standalone @knip/language-server is ESM, but the bundled-knip fallback in server.js used the CommonJS-only __filename. 
When no project-local knip resolves, this threw ReferenceError: __filename is not defined and no session was created. Replaced with import.meta.url.
Adds a stdio startup test that launches the real CLI against a temp project without knip and asserts the fallback works.

### DIFF
--- a/packages/language-server/test/startup.test.js
+++ b/packages/language-server/test/startup.test.js
@@ -1,0 +1,70 @@
+import assert from 'node:assert/strict';
+import { spawn } from 'node:child_process';
+import { mkdtempSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import test from 'node:test';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+import { NOTIFICATION_MODULE_GRAPH_BUILT } from '../src/constants.js';
+
+const cli = fileURLToPath(new URL('../src/cli.js', import.meta.url));
+
+const encode = payload => {
+  const body = JSON.stringify({ jsonrpc: '2.0', ...payload });
+  return `Content-Length: ${Buffer.byteLength(body)}\r\n\r\n${body}`;
+};
+
+test('falls back to bundled knip when the project has no local knip', async () => {
+  const cwd = mkdtempSync(path.join(tmpdir(), 'knip-ls-'));
+  writeFileSync(path.join(cwd, 'package.json'), '{"name":"standalone","main":"index.js"}');
+  writeFileSync(path.join(cwd, 'index.js'), '');
+
+  const server = spawn(process.execPath, [cli, '--stdio'], { cwd, stdio: ['pipe', 'pipe', 'pipe'] });
+  const logs = [];
+  let buffer = '';
+
+  const done = new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


packages/language-server/test/startup.test.js
```diff
@@ -0,0 +1,70 @@
...
+        workspaceFolders: [{ uri: pathToFileURL(cwd).href, name: 'standalone' }],
+      },
+    })
+  );
+  server.stdin.write(encode({ method: 'initialized', params: {} }));
+
+  try {
+    await done;
+  } finally {
+    server.kill();
```
webpro, line 62: One non-blocking nit: the new test never removes its temporary project; add teardown for that.

webpro (commented):
Thanks for the pull request! One small test cleanup suggestion inline.

packages/language-server/test/startup.test.js
gioboa: Thanks, [8d91e95](https://github.com/webpro-nl/knip/pull/2035/commits/8d91e956a89fbd6004800bd2e96775592370e097) solve it.

gioboa (commented):
Fixed @webpro 💪 Thank you

webpro:

PR #2038, opened 2026-09-13
Apply NODE_OPTIONS inputs to package manager binaries

giaBaoJS:
`getDependenciesFromCommand` parses the `NODE_OPTIONS` prefix into `fromNodeOptions`, then spreads it into three of the branches below it: `pluginArgsMap` (line 46), `binary in Plugins` (line 56) and the final `fallbackResolve` return (line 61). The `KnownResolvers` branch immediately above them returns the resolver output on its own, so those inputs are dropped for `bun`, `bunx`, `nub`, `nubx`, `find`, `npm`, `npx`, `pnpm`, `pn`, `pnpx`, `pnx` and `yarn`.

Effect: a script like

```json
"dev": "NODE_OPTIONS='--import ./instrumentation.mjs' pnpm exec next dev"
```

loses `./instrumentation.mjs`, while `NODE_OPTIONS='--import ./instrumentation.mjs' next dev` (fallback branch) and `NODE_OPTIONS='--loader pkg' knex` (plugin branch) pick it up.

Reproduced with `_getInputsFromScripts` over `fixtures/binaries`, current `main`:

* `NODE_OPTIONS='--require pkg-a' program` gives `pkg-a`
* `NODE_OPTIONS='--require pkg-a' knex` gives `pkg-a`
* the same prefix on `npm run build`, `pnpm run build`, `yarn run build`, `bun run build`, `npx tsc` and `pnpm exec next dev` gives nothing

The fix spreads `fromNodeOptions` into that branch too.

The two added assertions sit next to the existing `NODE_OPTIONS` ones in `getInputsFromScripts (cross-env/env vars)`, which already cover the plugin and fallback branches. `npm run script` pins that the inputs are still appended when the resolver itself returns nothing; `pnpm exec next dev` pins the order and the `--import` form. Both fail on `main`.

`pnpm test --runtime node` in `packages/knip`: 1160 pass, 0 fail, before and after.


giaBaoJS:
One thing I left out of the diff on purpose, so it does not look like an oversight.

`spawningBinaries` (line 49 after this change) drops `fromNodeOptions` the same way:

```js
if (spawningBinaries.includes(binary)) {
  return [toBinary(binary), ...fromWords(words, options)];
}
```

`NODE_OPTIONS='--require pkg-a' cross-env program` loses `pkg-a` there too. I did not include it because cross-env's documented usage puts `NODE_OPTIONS` in its arguments rather than in the shell prefix, and that form already works and is pinned by the existing assertion at line 126. So the shell-prefix case looked rare enough that folding it in would widen the diff without a fixture that clearly justifies it.

Happy to add it to this PR if you would rather have both branches fixed at once.

webpro:
Thanks for the pull request! Yes feel free to add it, much appreciated.

giaBaoJS:
Added in 92f21a3. `spawningBinaries` now spreads `fromNodeOptions` too, so `NODE_OPTIONS='--require pkg-a' cross-env program` keeps `pkg-a`. `retry-cli` was affected the same way.

Three assertions next to the existing ones pin: multiple `--require` values from the prefix (not just the first), that it applies to `retry-cli` and not only `cross-env`, and the nested case `NODE_OPTIONS=... cross-env NODE_OPTIONS=... program` where the inner argument form and the outer prefix are both collected. Reverting the one-line source change turns them red; the existing `cross-env NODE_OPTIONS="--loader pkg" knex` assertion is unaffected, since the outer node has no prefix there.

webpro:
Thanks again!

webpro:

PR #2039, opened 2026-09-14
Add textlint plugin

anandghegde:
Adds a plugin for [textlint](https://textlint.org). It turns on when `textlint` is a dependency.

Right now, projects using textlint get every `textlint-rule-*`, `textlint-rule-preset-*`, `textlint-filter-rule-*` and `textlint-plugin-*` package reported as an unused devDependency, because those packages are only referenced by short names in the config.

## What it does

- Config files: `.textlintrc`, `.textlintrc.{json,yaml,yml,js,cjs}` and the `textlint` key in `package.json`. This is the same list textlint's `@textlint/config-loader` loads via `rc-config-loader`.
- `rules`: each key maps to `textlint-rule-*`. Preset keys (`preset-x`, `@scope/preset-x`, `textlint-rule-preset-x`) map to `textlint-rule-preset-*`. Rules set to `false` are skipped, like textlint does.
- `filters` maps to `textlint-filter-rule-*`.
- `plugins` (array or object) maps to `textlint-plugin-*`.
- Script arguments: `--rule`, `--preset` and `--plugin` resolve the same way. `--config`/`-c` sends that file through `resolveConfig`.

Name resolution follows textlint's `createFullPackageName` and `TextLintModuleResolver`. textlint tries the prefixed name first (for example `@scope/x` → `@scope/textlint-rule-x`) and then the name as written. The plugin picks whichever of those is declared in `package.json`, similar to the Remark plugin. If none is declared, it reports the prefixed name as unlisted. A name that already has the prefix (`textlint-rule-foo`, `@scope/textlint-plugin-foo`) is reported unchanged.

Example:

```json
{
  "rules": {
    "no-todo": true,
    "@textlint-rule/no-unmatched-pair": true,
    "preset-ja-technical-writing": true
  },
  "filters": { "comments": true },
  "plugins": { "@textlint/markdown": true }
}
```

This resolves to `textlint-rule-no-todo`, `@textlint-rule/textlint-rule-no-unmatched-pair`, `textlint-rule-preset-ja-technical-writing`, `textlint-filter-rule-comments` and `@textlint/textlint-plugin-markdown`.

## Tests

- `test/plugins/textlint.test.ts` covers `.textlintrc.json` plus a script with `--config`, `--rule`, `--preset` and `--plugin`. It includes unscoped, scoped, already-prefixed, preset and disabled rules, and expects two unlisted rules.
- `test/plugins/textlint-package-json.test.ts` covers the `package.json` `textlint` key with array-style `plugins`.

Commands run:

- `bun test test/plugins/textlint.test.ts test/plugins/textlint-package-json.test.ts`: pass
- `pnpm test --runtime node --smoke` (packages/knip): 1055/1055 pass
- `pnpm build`, `pnpm knip`, `pnpm knip:production` (packages/knip): clean
- `pnpm run ci` (root: oxlint, oxfmt --check, installed-check, line endings): clean

Claude Code helped write this change. I reviewed it.

https://claude.ai/code/session_01QBWFWqjbUQGS1c9LsM2bxX


packages/knip/src/plugins/textlint/helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+    manifest.devDependencies?.[name] ??
+    manifest.optionalDependencies?.[name] ??
+    manifest.peerDependencies?.[name]
+  );
+
+const toPackageDependency = (prefix: string, name: string, candidates: string[], manifest: Manifest) =>
+  toDependency(
+    candidates.find(candidate => isDeclared(manifest, candidate)) ??
+      (hasPrefix(prefix, name) ? name : createFullPackageName(prefix, name))
+  );
```
webpro, line 38: [P2] Local rule paths become bogus dependencies

A rule configured with `require.resolve('./rules/check-document.cjs')` works in textlint, but Knip prefixes the resolved path and reports an unlisted `textlint-rule-`. Preserve local paths as file inputs.

packages/knip/src/plugins/textlint/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  for (const key of plugins) inputs.push(toPluginDependency(key, manifest));
+  return inputs;
+};
+
+const args: Args = {
+  config: true,
+  string: ['rule', 'preset', 'plugin'],
+  resolveInputs: (parsed, { manifest }) => {
+    const inputs: Input[] = [];
+    for (const rule of [parsed.rule ?? []].flat()) inputs.push(toRuleDependency(String(rule), manifest));
```
webpro: [P2] Comma-separated CLI values are not split

`--rule no-todo,max-comma` loads two rules in textlint. Knip reports both dependencies unused and invents an unlisted `textlint-rule-no-todo,max-comma`. Handle array values for rules, presets, and plugins.

webpro (commented):
Thanks for adding textlint support! Two dependency-resolution issues are detailed inline.

packages/knip/src/plugins/textlint/helpers.ts
anandghegde: Fixed in 13689737. Rules, filters and plugins given as a relative or absolute path are now added as entry files instead of being turned into a `textlint-rule-` dependency. The fixture now has a `.textlintrc.cjs` using `require.resolve('./rules/check-document.cjs')`: the rule file is picked up and nothing extra is reported as unlisted.

packages/knip/src/plugins/textlint/index.ts
anandghegde: Fixed in 13689737. `--rule`, `--preset` and `--plugin` values are now split on commas (and repeated flags still work). The fixture script now uses `--rule no-dead-link,no-empty-section --preset jtf-style,ja-spacing --plugin review,latex2e`. Before this change it reported all six as unused along with three invented unlisted names. Now it reports neither.

packages/knip/src/plugins/textlint/helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+const toPresetDependency = (key: string, manifest: Manifest) => {
+  const name = key.replace(/^preset-/, '').replace(/^@([^/]+)\/preset-(.*)$/, '@$1/$2');
+  const candidates = [`${PRESET_PREFIX}${name}`, name, createFullPackageName(PRESET_PREFIX, name), key];
+  return toPackageDependency(PRESET_PREFIX, name, candidates, manifest);
+};
+
+const toModuleDependency = (prefix: string, key: string, manifest: Manifest) =>
+  toPackageDependency(prefix, key, [createFullPackageName(prefix, key), key], manifest);
+
+const toLocalEntry = (key: string, cwd: string) => toEntry(toAbsolute(key, cwd));
```
webpro, line 56: [P2] Extensionless local modules are reported unused

A `.textlintrc.cjs` rule keyed by `path.join(__dirname, 'rules/check')` can load `rules/check.js` or `rules/check/index.js` in textlint. Here, `toEntry` treats the path as a glob, so Knip reports the loaded rule file as unused. Please use [toDeferResolveEntry](https://knip.dev/writing-a-plugin/inputs#todeferresolveentry) for these local module paths, and cover extensionless files and directory modules.

webpro (changes requested):
Thanks for addressing both earlier findings. One local-module resolution issue remains.

packages/knip/src/plugins/textlint/helpers.ts
anandghegde: Fixed in 6680de68. `toLocalEntry` now uses `toDeferResolveEntry`, so the path goes through module resolution instead of being treated as a glob:

```ts
const toLocalEntry = (key: string, cwd: string) => toDeferResolveEntry(toAbsolute(key, cwd));
```

That covers rules, filters and plugins, since all three route through this helper.

The fixture's `.textlintrc.cjs` now keys rules by all three shapes textlint accepts:

```js
rules: {
  [require.resolve('./rules/check-document.cjs')]: true,
  [path.join(__dirname, 'rules/check-heading')]: true,   // rules/check-heading.js
  [path.join(__dirname, 'rules/check-link')]: true,      // rules/check-link/index.js
}
```

The test asserts `Object.keys(issues.files).length === 0` and the counters go to 4 processed. On the previous commit the two new rule files are reported unused (`files: 2`), so it fails there and passes now.

`pnpm run ci` is clean.

webpro (approved):
Thank you! 🧡

webpro:

Issue #1317, opened 2025-10-23
🐛 Knip fails to recognize relative paths from Vitest project configurations


TrevorBurnham:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://codesandbox.io/p/devbox/h4h4v7

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

Vitest correctly recognizes relative paths to `setupFiles` from a root Vitest config file. However, if you use Vitest's [Test Projects](https://vitest.dev/guide/projects) feature, relative paths from the project-level config files are not resolved correctly:

#### Root Vitest config (vitest.config.mts)

```ts
export default defineConfig({
  test: {
    projects: ["app/vitest.config.mts"],
  },
});
```

#### Project config (app/vitest.config.mts)

```ts
export default defineConfig({
  test: {
    setupFiles: ["./__test__/vitest.setup.ts"],
  },
});
```

#### Knip output

```sh
Unused files (1)
app/__test__/vitest.setup.ts  
Unresolved imports (1)
./__test__/vitest.setup.ts  app/vitest.config.mts
```

webpro:
:rocket: _This issue has been resolved in v5.72.0. See [Release 5.72.0](https://github.com/webpro-nl/knip/releases/tag/5.72.0) for release notes._

_Using Knip in a commercial project? Please consider [becoming a sponsor](https://knip.dev/sponsors)._

PR #1511, opened 2026-02-05
fix: fix vitest setupFiles resolution

tmair:
Vitest resolves its `setupFiles` relative to the root directory. 

Reference: https://vitest.dev/config/setupfiles.html#setupfiles

PR #2040, opened 2026-09-14
fix: Correctly resolve Vitest setupFiles from nested configs

CruseCtrl:
## The problem

If no `test.root` is specified in a Vitest config, then Vitest resolves the `setupFiles` option in a `vitest.config.ts` file relative to the location of the config file. Knip was resolving these paths relative to the workspace root, which is incorrect when the config file is nested inside another directory.

If a `test.root` is provided then Vitest resolves it relative to the cwd, which is the workspace root. Knip was already doing the correct thing in this case.

## The fix

When building the `vitestRoot`, if no `test.root` is provided then we now use the `configFileDir`, not the `cwd`.

## Additional info

This was raised as an issue in #1317 and had already been fixed, but seems to have regressed since then.


packages/knip/src/plugins/vitest/index.ts
```diff
@@ -174,7 +174,7 @@ export const resolveConfig: ResolveConfig<ViteConfigOrFn | VitestWorkspaceConfig
       for (const entry of await getIndexHtmlEntries(viteRoot, publicDir)) inputs.add(entry);
     }
 
-    const vitestRoot = toAbsolute(cfg.test?.root ?? '.', options.cwd);
+    const vitestRoot = cfg.test?.root ? toAbsolute(cfg.test.root, options.cwd) : toAbsolute('.', options.configFileDir);
```
webpro: [P2] Preserve the standalone config root when fixing project configs

A child config loaded through `test.projects` defaults to its config directory. Running `vitest --config config/vitest.config.mjs` from the repository root still defaults to the current working directory. With `setupFiles: ['./bootstrap.mjs']` and `include: ['tests/**/*.test.mjs']`, this change reports the live setup and test files as unused and the setup import as unresolved; the base reports neither issue. This reproduces with Vitest 4.1.10 and 5.0.0.

The same resolver is used by the Vite plugin, and `build.lib.entry` is joined to `vitestRoot`. With `vite build --config config/vite.config.mjs` and `build.lib.entry: 'src/library.js'`, Vite 8.3.0 builds successfully, while this head newly reports `src/library.js` unused. Retain the standalone root semantics and apply the config-directory default specifically to child project configs. Keep Vite library entries on the Vite root.

webpro (commented):
Thanks for adding a regression test for nested Vitest projects! It reproduces the reported bug, but the fallback change also affects standalone Vitest and Vite configurations. Details are inline.

The history explains why this needs coverage for both ways of loading a nested config:

- [The #1317 fix](https://github.com/webpro-nl/knip/commit/ff8c33d20e1cfbd9eaa7d7f963e5ddbf99678770), released in 5.72.0, switched setup and global-setup resolution to the config directory for every config. That fixed child projects but mishandled explicit roots.
- [#1511](https://github.com/webpro-nl/knip/commit/273b940f7e32ff7156c3a24875f5d9265ff2559a), released in 5.83.1, corrected explicit `test.root` handling. Its cwd fallback reintroduced the child-project setup bug.
- [#1982](https://github.com/webpro-nl/knip/commit/0ecf48538c339437e6f9dfe9a6013d307783cbd3), released in 6.33.0, moved `globalSetup` to that same root, extending the regression to global setup.

The original [vitest9 fixture](https://github.com/webpro-nl/knip/blob/ff8c33d20e1cfbd9eaa7d7f963e5ddbf99678770/packages/knip/fixtures/plugins/vitest9/package.json) declared `workspaces: ['packages/*']`, and the nested config lived beside `packages/client/package.json`. Its config directory and Knip workspace cwd were identical, so the fixture could not catch the reversal. The new fixture avoids that blind spot.

For regression coverage, please:

- Keep the new child-project fixture and cover `globalSetup` alongside `setupFiles`.
- Add a standalone nested `--config` fixture whose setup files and test entries resolve from cwd.
- Add a nested Vite config with a non-default `build.lib.entry` resolved from the Vite root.
- Preserve the explicit `test.root` coverage, with the config directory, workspace root, and test root at different locations.

packages/knip/src/plugins/vitest/index.ts
CruseCtrl: Both fixed, thanks for the detailed repro steps — they were two separate bugs in my original change:

1. `vitestRoot` was defaulting to `configFileDir` unconditionally whenever `test.root` was unset, but that default only applies to a project actually discovered through a `test.projects` glob — per Vitest's own `resolveProjects.ts`, that's the only place it special-cases root defaulting. A standalone config (however it's loaded, `-c`/`--config` included) still defaults to `process.cwd()`. I've threaded a new `isResolvedConfigFile` flag through `PluginOptions`, set only when a config file was discovered as a dependency of another config of the same plugin (e.g. via `test.projects`), so the plugin can tell the two cases apart.

2. `build.lib.entry` was accidentally joined against `vitestRoot` instead of `viteRoot` — it's a Vite-level field, not a Vitest one, so it now always resolves against the Vite root regardless of how `vitestRoot` is computed.

Added `vitest13` (your first repro) and `vitest14` (your second) as fixtures — both fail on the previous version of this branch and pass now, and the full existing `vitest*` suite is unaffected.

One related gap surfaced while digging into this, which I'm leaving out of this PR since it's pre-existing rather than a regression: a **directory** glob in `test.projects` (e.g. `test/*`, as opposed to `test/*/vitest.config.ts`) still isn't resolved to the config files inside it — Vitest itself falls back to looking for a default config file in the matched directory in that case, which this plugin doesn't do.

webpro:
Thanks a bunch, Harry.

webpro:

PR #2042, opened 2026-09-15
fix(node): add valueless Node CLI flags to boolean options

shoutoutuoadi325:
The Jest ESM guide recommends running Jest like this:

```
"test": "node --experimental-vm-modules node_modules/.bin/jest"
```

Knip reported `jest` as an unused devDependency for such a script. `--experimental-vm-modules` is a valueless flag, but it was missing from the Node plugin's boolean options, so the parser consumed `node_modules/.bin/jest` as its value and the binary was never attributed to the dependency.

This adds a few more documented valueless Node CLI flags to the list: `--enable-source-maps`, `--experimental-detect-module`, `--experimental-json-modules`, `--experimental-vm-modules`, `--expose-gc` and `--openssl-legacy-provider`.


webpro:
Thank you! 🙏

webpro:

PR #2043, opened 2026-09-15
fix(typedoc): accept string form of plugin and theme options

giaBaoJS:
## What is wrong

TypeDoc accepts a single string for both `plugin` and `theme`:

```json
// typedoc.json
{
  "plugin": "typedoc-plugin-mdn-links",
  "theme": "typedoc-github-theme"
}
```

On that config Knip reports 18 unresolved single-character specifiers
(`t`, `y`, `p`, `e`, `d`, `o`, `c`, `-`, `l`, `u`, `g`, `i`, `n`, `m`, `k`, `s`,
`h`, `b`) in `typedoc.json`, and still reports both packages as unused
devDependencies.

## Why

`packages/knip/src/plugins/typedoc/index.ts:29-31` on main:

```ts
const plugins = cfg?.plugin ?? [];
const themes = cfg?.theme ?? [];
const inputs: Input[] = [...plugins, ...themes].map(id => toDeferResolve(id));
```

Both values are spread, so a string is spread character by character.
`types.ts` declares them as `string[]`, but upstream `theme` is
[`ParameterType.String`](https://github.com/TypeStrong/typedoc/blob/master/src/lib/utils/options/sources/typedoc.ts)
and `plugin` is `ParameterType.PluginArray`, which goes through
`toStringArray()` in `options/declaration.ts` and wraps a bare string. So both
keys are legitimately `string | string[]`.

## The fix

Normalize both with `arrayify()` and widen the two types to
`string | string[]`. The array form keeps working unchanged.

## Tests

New `test/plugins/typedoc2.test.ts` with `fixtures/plugins/typedoc2`, covering
the string form in `typedoc.json` and the array form in `tsconfig.json`
(`typedocOptions`). Before the fix it fails with the 18 character specifiers
above; after it, `unresolved` is empty and only `typedoc` itself is reported
as an unused devDependency.

## Scope

Limited to `plugin` and `theme`. `customCss` and `customJs` are
`ParameterType.Path` upstream (single string only) and are already read as
strings, so they are unchanged. The CLI `args.resolve` list for `--plugin` and
`--theme` is not affected.


webpro:
Short 'n sweet! Thank you 🧡

webpro:

PR #2046, opened 2026-09-17
Fix --format name resolution in the ESLint plugin

bytedoe:
The ESLint plugin turned every `--format` value that wasn't built-in or a `./` path into `eslint-formatter-<value>`. ESLint's `loadFormatter` does more: scoped names go through `normalizePackageName` (`@microsoft/sarif` → `@microsoft/eslint-formatter-sarif`) and already-prefixed names are left alone. So `-f @microsoft/sarif` produced an unlisted `eslint-formatter-` plus an unused `@microsoft/eslint-formatter-sarif`, and `--format eslint-formatter-codeframe` produced `eslint-formatter-eslint-formatter-codeframe`.

The built-in set also depends on the ESLint major: `checkstyle`, `compact`, `jslint-xml`, `junit`, `tap`, `unix` and `visualstudio` were part of core until v8 and are separate `eslint-formatter-*` packages since v9, so on ESLint 8 they were reported as unlisted.


packages/knip/src/plugins/eslint/helpers.ts
```diff
@@ -137,12 +138,26 @@ export const getInputsFromSettings = (settings?: Settings) =>
...
+]);
+
+export const resolveFormatters = (formatters: string | string[], manifest: Manifest) => {
   const inputs: Set<Input> = new Set();
+  const isBeforeV9 = (manifest.getMajor('eslint') ?? 9) < 9;
   for (const formatter of [formatters].flat()) {
-    if (builtinFormatters.has(formatter)) continue;
-    else if (isInternal(formatter)) inputs.add(toDeferResolve(formatter));
-    else inputs.add(toDependency(`eslint-formatter-${formatter}`));
+    if (builtinFormatters.has(formatter) || (isBeforeV9 && builtinFormattersUntilV8.has(formatter))) continue;
```
webpro: [P2] Preserve installed formatter dependencies on ESLint 8

With `--format compact`, ESLint loads an installed `eslint-formatter-compact` before its built-in fallback. The unconditional skip loses that dependency reference. Please preserve the package reference when present while allowing the built-in fallback when absent. [Upstream behavior](https://github.com/eslint/eslint/blob/v8.57.1/lib/cli-engine/cli-engine.js#L1035-L1044).

packages/knip/src/plugins/eslint/helpers.ts
```diff
@@ -137,12 +138,26 @@ export const getInputsFromSettings = (settings?: Settings) =>
...
+    if (!formatter.startsWith('@') && formatter.includes('/')) {
+      inputs.add(toDeferResolve(isInternal(formatter) ? formatter : `./${formatter}`));
+    } else {
+      inputs.add(toDependency(resolveSpecifier('eslint-formatter', formatter)));
```
webpro: [P2] Use formatter-specific package qualification

`--format eslint-plugin-summary` should reference `eslint-formatter-eslint-plugin-summary`. The shared resolver instead records `eslint-plugin-summary`, causing unused and unlisted dependency reports. Formatter qualification needs to be prefix-specific. [Upstream rules](https://github.com/eslint/eslint/blob/v9.39.1/lib/shared/naming.js#L32-L55).

packages/knip/src/plugins/eslint/helpers.ts
```diff
@@ -137,12 +138,26 @@ export const getInputsFromSettings = (settings?: Settings) =>
...
+
+export const resolveFormatters = (formatters: string | string[], manifest: Manifest) => {
   const inputs: Set<Input> = new Set();
+  const isBeforeV9 = (manifest.getMajor('eslint') ?? 9) < 9;
   for (const formatter of [formatters].flat()) {
-    if (builtinFormatters.has(formatter)) continue;
-    else if (isInternal(formatter)) inputs.add(toDeferResolve(formatter));
-    else inputs.add(toDependency(`eslint-formatter-${formatter}`));
+    if (builtinFormatters.has(formatter) || (isBeforeV9 && builtinFormattersUntilV8.has(formatter))) continue;
+    if (!formatter.startsWith('@') && formatter.includes('/')) {
```
webpro, line 163: [P2] Normalize backslashes before classifying formatter paths

`.\formatters\custom.cjs` produces an unlisted `eslint-formatter-.\formatters\custom.cjs`. Normalize backslashes before classifying paths, as [ESLint does](https://github.com/eslint/eslint/blob/v9.39.1/lib/eslint/eslint.js#L1202-L1213).

webpro (changes requested):
Thanks for improving formatter resolution. The scoped-name and ESLint 8 cases are fixed, but three other cases still produce incorrect dependency reports.

webpro (approved):
Thanks again! Your contributions are really much appreciated, @bytedoe 🙏

Going to merge this one, I think we can ignore the temp (network?) issues in CI.

webpro:
:rocket: _This pull request is included in v6.37.0. See [Release 6.37.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.37.0) for release notes._

PR #2048, opened 2026-09-17
fix(graphql-codegen): mark near-operation-file outputs as entries, not the documents directory

### LINKED ISSUE #2047: GraphQL Codegen plugin marks the whole documents directory as entries for the near-operation-file preset
### Prerequisites

- [x] I'm using the latest version (6.36.0)
- [x] I've read the relevant documentation
- [x] I've searched for existing issues
- [x] I've checked the list of known issues
- [x] I've read the issue reproduction guide

### Reproduction url

https://github.com/RobHannay/knip-repro-near-operation-file

### Reproduction access

### Description of the issue

With the `near-operation-file` preset, the `generates` key is the base directory of the documents, not a location of generated files:

```ts
generates: {
  './src/': {
    documents: 'src/**/*.ts',
    preset: 'near-operation-file',
    presetConfig: { baseTypesPath: 'types.ts', folder: '__generated__' },
    plugins: ['typescript-operations'],
  },
},
```

The GraphQL Codegen plugin turns every `generates` key into a production entry and appends `**` to a key ending in a slash, so this config makes `src/**` an entry pattern. Every source file is then an entry, and Knip can never report an unused file under `src/`. Exports are still checked, since exports of entry files are only reported with `isIncludeEntryExports`, so the gap is easy to miss.

In the reproduction, `npx knip --include files` reports nothing, although `src/unused.ts` is imported by nothing. In our monorepo the same shape hid 100 unused files across three apps until the plugin was disabled for that workspace.

Expected: only the generated files are entries. For this preset that is `<key>/**/<folder>/<fileName or *><extension>`, using the preset's defaults `folder: ''` and `extension: '.generated.ts'` when `presetConfig` does not set them.

I have a fix with a fixture and test ready and will open a PR referencing this issue.

RobHannay:
Fixes #2047.

The `near-operation-file` preset writes one generated file next to each document, so its `generates` key is the documents' base directory. The plugin appended `**` to that key, which made every file under the directory a production entry and hid every unused file there (reproduction in the issue).

The plugin now recognises the preset by name (`near-operation-file`, `near-operation-file-preset`, `@graphql-codegen/near-operation-file-preset`) and marks `<key>/**/<folder>/<fileName ?? *><extension>` as entries instead, reading `folder`, `extension` and `fileName` from `presetConfig` with the preset's own defaults (`''` and `.generated.ts`). Other presets and plain outputs are unchanged. A preset passed as an object cannot be identified and keeps the previous behaviour.

New fixture and test: `fixtures/plugins/graphql-codegen-near-operation-file`, where `src/unused.ts` is now reported as an unused file while the generated file under `__generated__` stays an entry. The existing graphql-codegen tests pass.


packages/knip/src/plugins/graphql-codegen/index.ts
```diff
@@ -49,6 +56,16 @@ const getPluginPackageName = (name: string) => {
   return `@graphql-codegen/${name}`;
 };
 
+// https://the-guild.dev/graphql/codegen/docs/presets/near-operation-file
+const getOutputPattern = (output: string, outputConfig: ConfiguredOutput | ConfiguredPlugin[]) => {
+  if (isConfigurationOutput(outputConfig) && isNearOperationFilePreset(outputConfig.preset)) {
+    const presetConfig: NearOperationFilePresetConfig = outputConfig.presetConfig ?? {};
+    const { folder = '', extension = '.generated.ts', fileName = '*' } = presetConfig;
+    return join(output, '**', folder, `${fileName}${extension}`);
```
webpro: [P2] Parent-relative folders lose recursive matching

With output `src/` and `folder: '../__generated__'`, `join()` removes `**`, producing `src/__generated__/*.generated.ts`. Generated files under `src/features/__generated__/` are consequently reported as unused. Preserve recursive matching for parent-relative folders. [Upstream behavior](https://github.com/dotansimha/graphql-code-generator-community/blob/main/packages/presets/near-operation-file/src/utils.ts#L20-L33).

packages/knip/src/plugins/graphql-codegen/index.ts
```diff
@@ -49,6 +56,16 @@ const getPluginPackageName = (name: string) => {
   return `@graphql-codegen/${name}`;
 };
 
+// https://the-guild.dev/graphql/codegen/docs/presets/near-operation-file
+const getOutputPattern = (output: string, outputConfig: ConfiguredOutput | ConfiguredPlugin[]) => {
+  if (isConfigurationOutput(outputConfig) && isNearOperationFilePreset(outputConfig.preset)) {
+    const presetConfig: NearOperationFilePresetConfig = outputConfig.presetConfig ?? {};
+    const { folder = '', extension = '.generated.ts', fileName = '*' } = presetConfig;
```
webpro: [P2] filePerOperation overrides fileName

With `fileName: 'types'` and `filePerOperation: true`, query `GetUser` generates `GetUser.generated.ts`, but the pattern only matches `types.generated.ts`. Use a wildcard filename when `filePerOperation` is enabled. [Upstream precedence](https://github.com/dotansimha/graphql-code-generator-community/blob/main/packages/presets/near-operation-file/src/index.ts#L255-L272).

webpro (changes requested):
Thanks for fixing the unused-file detection here. Two supported configurations still cause generated files to be reported as unused; details are inline.

packages/knip/src/plugins/graphql-codegen/index.ts
RobHannay: Fixed in 20a3cfea5: the `..` segments are dropped and the recursive match kept, so `folder: '../__generated__'` with a document at `features/nested/deep/Query.ts` now matches `features/nested/__generated__/Query.generated.ts`. Covered in the fixture and test.

packages/knip/src/plugins/graphql-codegen/index.ts
RobHannay: Fixed in 20a3cfea5: `filePerOperation` switches the file name to `*`. The fixture now has `fileName: 'types'` with `filePerOperation: true` and a generated `operations/GetUser.generated.ts`, and the test asserts it is not reported.

packages/knip/src/plugins/graphql-codegen/index.ts
```diff
@@ -49,6 +56,19 @@ const getPluginPackageName = (name: string) => {
...
 };
 
+// https://the-guild.dev/graphql/codegen/docs/presets/near-operation-file
+const getOutputPattern = (output: string, outputConfig: ConfiguredOutput | ConfiguredPlugin[]) => {
+  if (isConfigurationOutput(outputConfig) && isNearOperationFilePreset(outputConfig.preset)) {
+    const presetConfig: NearOperationFilePresetConfig = outputConfig.presetConfig ?? {};
+    const { folder = '', extension = '.generated.ts', fileName = '*', filePerOperation = false } = presetConfig;
+    // The preset resolves `folder` against each document's own directory, so a parent-relative folder still
+    // lands below the output directory: keep the recursive match and drop the `..` segments
+    const subfolder = folder.split('/').filter(segment => segment !== '' && segment !== '.' && segment !== '..');
```
webpro: [P2] Normalize the folder before removing parent segments

`folder: 'cache/../__generated__'` becomes `cache/__generated__` instead of `__generated__`, falsely reporting generated files as unused. Normalize the folder before removing remaining leading `..` segments.

webpro (changes requested):
Thanks for fixing both earlier issues. There is one remaining regression in folder normalization.

packages/knip/src/plugins/graphql-codegen/index.ts
RobHannay: Fixed in 1e061c962: the folder goes through `normalize()` first, so `cache/../__generated__` becomes `__generated__` and only the parent segments left at the front are dropped. Fixture case added (`normalized/deep/Op.ts` with that folder) and asserted in the test.

webpro (approved):
Thank you for fixing all three issues and adding regression coverage. The generated-file patterns now handle the cases raised in review, and unused source files are reported as expected.

webpro:
:rocket: _This pull request is included in v6.37.0. See [Release 6.37.0](https://github.com/webpro-nl/knip/releases/tag/knip%406.37.0) for release notes._

PR #2051, opened 2026-09-18
Resolve file option in Mocha configuration files

giaBaoJS:
Mocha loads everything listed under `file` before the test files, but the plugin
only reads `spec` and `require` from the configuration, so those files end up in
"Unused files".

With `mocha` as a devDependency and this `.mocharc.json`:

```json
{ "spec": ["specs/*.test.js"], "file": ["hooks/db.setup.js"] }
```

`hooks/db.setup.js` is reported as unused.

`resolveConfig` in `packages/knip/src/plugins/mocha/index.ts:22-30` builds inputs
from `spec` and `require` only, while the argument parser a little further down
already handles `--file` through `toEntry(parsed.file)`. The two paths disagree.
This adds `file` to `resolveConfig` with the same `[value].flat()` idiom the
neighbouring options use, so the string and the array form both work, and with
`toEntry` because `--file` takes a file, directory or glob just like the argument
path.

The new fixture covers the array-with-glob form in `.mocharc.json` and the plain
string form in the `package.json` `mocha` property.

One thing I left alone: `ignore`/`exclude` is still only handled on the argument
path. Add `"ignore": ["specs/slow.test.js"]` to the config above and
`specs/slow.test.js` stays an entry instead of being reported. That one turns
into a negated pattern and has its own wrinkle (mocha exempts `--file` from
`--ignore`), so it felt like a separate change. `extension` only matters for
directory arguments, which `resolveConfig` never expands.


packages/knip/src/plugins/mocha/index.ts
```diff
@@ -21,10 +21,12 @@ const defaultExtensions = ['js', 'cjs', 'mjs'];
 
 const resolveConfig: ResolveConfig<MochaConfig> = localConfig => {
   const entryPatterns = localConfig.spec ? [localConfig.spec].flat() : entry;
+  const files = localConfig.file ? [localConfig.file].flat() : [];
   const require = localConfig.require ? [localConfig.require].flat() : [];
 
   const inputs: Input[] = [];
   inputs.push(...entryPatterns.map(id => toEntry(id)));
+  inputs.push(...files.map(id => toEntry(id)));
```
webpro: Resolve `file` values as module paths.

Mocha [resolves each `file` value from the process cwd with Node's module resolver](https://github.com/mochajs/mocha/blob/70db1c70cfe8b6a35d3e1f6c099a614abb9ff137/lib/cli/collect-files.cjs#L60-L91). For example, `file: "hooks/setup"` loads `hooks/setup.js`, and `file: "hooks"` can load `hooks/index.js`. Both cases successfully initialize the test environment in Mocha 12.0.2, but this PR still reports those files as unused because `toEntry()` treats the values as glob patterns.

Could you resolve these as cwd-relative module paths, including extensionless and directory-module resolution, and cover those cases in the fixture? Please also replace `hooks/*.setup.js` with supported `file` values: Mocha does not expand that glob, and running the new fixture through Mocha exits 1 because it cannot resolve the literal path.

webpro (changes requested):
Thanks for adding `file` support and covering both configuration sources! One path-resolution issue needs addressing before merging.

packages/knip/src/plugins/mocha/index.ts
giaBaoJS: Fixed. `file` values now go through `toDeferResolve` rather than `toEntry`, which matches the `require` line below it and the `setupFiles` handling in the jest and vitest plugins.

I checked this against Mocha 12.0.2 rather than the docs. `collect-files.cjs` calls `require.resolve(path.resolve(file))`, so `hooks/db.setup` loads `hooks/db.setup.js` and `teardown` loads `teardown/index.js`. The fixture now uses both of those shapes in `.mocharc.json` and keeps the explicit `hooks/globals.js` in `package.json`; running `mocha` inside it exits 0 with all three files executing, where the old `hooks/*.setup.js` exited 1.

Outside this PR's scope: `args.resolveInputs` routes `parsed.file` through the same `toEntry` path as `parsed._` and `parsed.spec`, so `mocha --file hooks/db.setup` has the same false positive. I can fold that in here or leave it for a separate PR.

webpro: Thanks, but please use `toDeferResolveEntry` (it's for files and not dependencies).

giaBaoJS: Done, switched to `toDeferResolveEntry`. Full suite on Node stays at 1325 passing, and the fixture still exits 0 under Mocha 12.0.2.

webpro:
Thanks, Bao! 🧡

webpro:

PR #2052, opened 2026-09-18
Add sitelo plugin

Paul-Browne:
Adds a plugin for [sitelo](https://github.com/Paul-Browne/sitelo), a file-based static site generator on Vite. It turns on when `sitelo` is a dependency.

sitelo discovers pages from the filesystem: every `*.ht.js` (also `.html.js`, `.ht.ts`, `.ht.tsx`, …) under `src/` is a page, and every module under `src/islands/` is a server island rendered on request. Nothing imports any of them, so today Knip reports the whole site as unused files — on sitelo's own docs site that's 593 of them.

## What it does

- Marks pages and islands as production entries. Defaults: `src/**/*.{ht,html}.{js,jsx,ts,tsx}` and `src/islands/*.{js,mjs,cjs,ts}`.
- Config file: `sitelo.config.{js,mjs}`. Read via `resolveFromAST` (no execution) from the default-export object: `root`, `pagesDir`, `pageExtensions`, `include` and `exclude`, resolved the same way sitelo's own page discovery does (`include` replaces the extension-derived globs; `exclude` becomes negated patterns). Only the top-level object is read, so nested options like `lighthouse.include` are left alone.

## Fixtures

- `plugins/sitelo` — custom `pagesDir`, a custom page extension and an `exclude`; asserts pages, islands and their imports are not reported while an excluded draft and an orphan module are.
- `plugins/sitelo-defaults` — no config file, exercises the static defaults.

With the plugin disabled the same fixtures report 7 and 4 unused files respectively.

## Notes

- Verified against sitelo's example sites: `basic` and `blog` come back clean.
- Client scripts referenced from page templates as HTML strings (`<script src="/js/app.js">`) aren't followed; users list those as `entry` themselves. A visitor over page template literals could cover this in a follow-up, but I've kept this PR to entry discovery.
- I'm the author of sitelo.


webpro:
The SSG does not seem to have a lot of traction yet, so I'd rather not maintain a plugin for it yet.

Paul-Browne:
ok, just for future reference what would constitute tractions?

webpro:
To be honest I don't have a clear picture of that. In general, a plugin like this can mostly be replaced with custom (production) entry points in Knip config. I guess a sitelo user should enter the scene wanting to reduce that ridiculous Knip config.

Issue #2053, opened 2026-09-18
🐛  oxlint plugin: ignoring extends field


matthewnitschke-wk:
### Prerequisites

- [x] I'm using the latest version
- [x] I've read the relevant [documentation](https://knip.dev)
- [x] I've searched for [existing issues](https://github.com/webpro-nl/knip/issues?q=is%3Aissue)
- [x] I've checked the [list of known issues](https://knip.dev/reference/known-issues)
- [x] I've read the [issue reproduction guide](https://knip.dev/guides/issue-reproduction)

### Reproduction url

https://github.com/matthewnitschke-wk/knip-oxlint-extends-issue

### Reproduction access

- [x] I've made sure the reproduction is publicly accessible

### Description of the issue

It appears as though the oxlint plugin is ignoring the `extends` key that can be used within an [oxlint configuration](https://oxc.rs/docs/guide/usage/linter/config.html#extend-shared-configs)

This is resulting in knip incorrectly reporting that dev dependencies are unused when in fact they are indeed used

https://github.com/matthewnitschke-wk/knip-oxlint-extends-issue contains a full repro, but in its simplest format, the following criteria can be used to replicate the issue:

```ts
// oxlint.config.ts
import {defineConfig} from 'oxlint';
import otherConfig from './other.config';

export default defineConfig({
  extends: [otherConfig]
})

// other.config.ts
export default defineConfig({
  jsPlugins: ['eslint-plugin-formatjs']
})
```

The resulting `knip` run will list `eslint-plugin-formatjs` as an unused dependency, when it is indeed used through the shared config

```
$ pnpm knip
Unused devDependencies (1)
eslint-plugin-formatjs  package.json:11:6
```

matthewnitschke-wk:
I've put up a fix for this issue here: https://github.com/webpro-nl/knip/pull/2054

webpro:
Closing this one as #2054 has been merged and released. Thanks again!

PR #2054, opened 2026-09-18
Support oxlint extends

matthewnitschke-wk:
Adds [extends](https://oxc.rs/docs/guide/usage/linter/config.html#extend-shared-configs) support to the oxlint plugin

This is the fix for the issue I originally reported here: #2053

It can be tested by using my test repository here: https://github.com/matthewnitschke-wk/knip-oxlint-extends-issue


packages/knip/src/plugins/oxlint/index.ts
```diff
@@ -36,15 +38,37 @@ const resolveJsPlugins = (jsPlugins: OxlintConfig['jsPlugins']): Input[] => {
 
 const isLoadConfig: IsLoadConfig = ({ configFileName }) => !isViteConfig(configFileName);
 
-const resolveConfig: ResolveConfig<OxlintConfig> = config => {
-  const inputs = resolveJsPlugins(config.jsPlugins);
+const resolveExtendedConfig = async (config: OxlintConfig, configFilePath: string): Promise<Input[]> => {
+  const inputs: Input[] = [];
+  for (const entry of config.extends ?? []) {
+    if (typeof entry === 'string') {
```
matthewnitschke-wk, line 44: extends can either be a string to a local config path:

```ts
extends: ["./some/config.json"]
```

or a full object:

```ts
let someObject = { ... }
extends: [someObject]
```

`resolveExtendedConfig` supports both

packages/knip/src/plugins/oxlint/index.ts
```diff
@@ -36,15 +38,37 @@ const resolveJsPlugins = (jsPlugins: OxlintConfig['jsPlugins']): Input[] => {
...
+      if (isInternal(entry)) {
```
webpro: Treat string extends entries as config-file paths.

Oxlint's [JSON extends entries are paths relative to the declaring config](https://oxc.rs/docs/guide/usage/linter/nested-config#extending-configuration-files). For a valid `extends: ["base.json"]`, `isInternal()` returns false, so this branch reports an unlisted dependency named `base.json` and leaves the shared config's plugin dependency unused. Oxlint 1.83.0 loads that same config and executes its plugin. Please resolve string entries as config-relative paths and pass them through the existing config-input handling. The package-name fixture also needs replacing: Oxlint rejects `extends: ["oxlint-config-shared"]`, which makes root `pnpm run ci` fail before linting. Package configs are supported through imported objects in TypeScript configs.

packages/knip/src/plugins/oxlint/index.ts
```diff
@@ -36,15 +38,37 @@ const resolveJsPlugins = (jsPlugins: OxlintConfig['jsPlugins']): Input[] => {
...
+          if (extendedConfig) {
+            for (const input of await resolveExtendedConfig(extendedConfig, resolvedPath)) inputs.push(input);
+          }
+        }
+      }
+    } else {
+      for (const input of await resolveExtendedConfig(entry, configFilePath)) inputs.push(input);
+    }
+  }
+  for (const input of resolveJsPlugins(config.jsPlugins)) inputs.push(input);
```
webpro: Preserve the declaring directory for inherited local plugins.

With `.oxlintrc.json` extending `./configs/base.json`, a `jsPlugins` specifier of `./rules-plugin.cjs` in that shared config refers to `configs/rules-plugin.cjs`. The helper returns `toEntry('./rules-plugin.cjs')`, which Knip expands from the workspace root. An actual Oxlint run executes this inherited plugin, while the PR reports `configs/rules-plugin.cjs` unused. Please resolve relative plugin entries against their declaring JSON config before returning them, including entries in overrides.

webpro (changes requested):
Thanks for adding extends support! The imported-config case fixes the reported unused dependency. Two JSON path-resolution issues still need addressing.

packages/knip/src/plugins/oxlint/index.ts
matthewnitschke-wk: Ah, good callout, I misunderstood the spec

Fixed in [`ca19b8e`](https://github.com/webpro-nl/knip/pull/2054/commits/ca19b8e9f5109614c746810480a1c1f98bc7de7c)

packages/knip/src/plugins/oxlint/index.ts
matthewnitschke-wk: Fixed in [`7441a06`](https://github.com/webpro-nl/knip/pull/2054/commits/7441a063c9d86e906492cc9f28082f6c716e75c9)

webpro (approved):
Thank you, Matthew! 🧡

matthewnitschke-wk:
@webpro seems like `slonik` is [importing its config](https://github.com/gajus/slonik/blob/main/oxlint.config.ts#L1) from [oxlint-config-canonical](https://github.com/gajus/oxlint-config-canonical), which _is_ [declaring its jsPlugin dependency correctly](https://github.com/gajus/oxlint-config-canonical/blob/main/package.json#L66)

I think this might have surfaced a real bug with the jsPlugin relative path resolution that was just updated. I'm looking into it

webpro:
Forgot about `isFilterTransitiveDependencies` 😅

webpro:

PR #2057, opened 2026-09-21
feat: support tsconfig moduleSuffixes resolution

### LINKED ISSUE #2056: 💡 Typescript moduleSuffixes support
### Suggest an idea for Knip

I'm currently using a custom moduleSuffixes in my tsconfig, but it appears that knip fails to parse that, showing `Unresolved import: ./FILE knip(unresolved)`.
Could support for it be added to the knip json, letting knip lookup more suffixes?
Thanks

gioboa:
Closes #2056

Read `compilerOptions.moduleSuffixes` from tsconfig and try each suffix (plus index variants) in order, since oxc-resolver lacks support.


webpro:
As you're saying, this is a missing feature in oxc-resolver. Not sure yet if we should implement it in Knip. Mainly because the initial version in this PR isn't complete (e.g. should be registered per workspace) and there's some module resolution rules not taken care of yet (I learned from Metro source code and implementing the (incomplete/incorrect) [metro plugin](https://github.com/webpro-nl/knip/blob/main/packages/knip/src/plugins/metro/index.ts) this isn't trivial).

How about first asking for support upstream?

gioboa (commented):
Agreed, upstream first. Tracked in oxc-project/oxc-resolver#1393 (opened by @ofek-frame). Closing this; happy to revisit if upstream stalls.

PR #2059, opened 2026-09-22
fix(compilers): don't treat `import.meta` as an import

vdavid:
**Problem:** `import.meta` is the one `import` with no specifier, so `importMatcher` matched it and ran on to the next quoted string in the script. The fragment that produces is malformed, which costs the file every dynamic import in it, and a module imported only that way gets reported as an unused file.

**My solution in this PR:** The matcher now rejects a `.` after the keyword. No valid import statement or call has one there, so `import.meta` is all this drops. Proved by two tests.

Feel free to drop the comment if it's too much, or I can do it, just let me know.


webpro:
Thanks David! 🧡

webpro:

